PR #5, opened 2026-03-01
fix(executor): add stream-level byte cap to prevent OOM

rjkaes:
## Summary

- **Add stream-level byte cap to `PolyglotExecutor#spawn`:** Track cumulative stdout+stderr bytes during streaming and SIGKILL the process once the combined total exceeds `hardCapBytes` (default 100MB). Previously, `#maxOutputBytes` was only applied post-mortem via `#smartTruncate` — a command like `yes` or `cat /dev/urandom | base64` could accumulate gigabytes in memory before the timeout fired, OOM-killing the MCP server.
- **Diagnostic message on cap:** Appends `[output capped at NMB — process killed]` to stderr so callers can distinguish capping from truncation.
- **Configurable via constructor:** The `hardCapBytes` option allows callers to tune the limit.

## Test plan

- [x] All 57 existing executor tests pass with no regressions
- [x] New `tests/stream-cap.test.ts` (7 tests) covers:
  - Process killed when stdout exceeds hard cap
  - Process killed when stderr exceeds hard cap
  - Cap applies to combined stdout+stderr total
  - Normal operation unaffected when output is under cap
  - Cap message reports correct MB value
  - Timeout still fires independently when cap is not reached
  - Default 100MB cap sanity check

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -28,15 +28,18 @@ interface ExecuteFileOptions extends ExecuteOptions {
 
 export class PolyglotExecutor {
   #maxOutputBytes: number;
+  #hardCapBytes: number;
   #projectRoot: string;
   #runtimes: RuntimeMap;
 
   constructor(opts?: {
     maxOutputBytes?: number;
+    hardCapBytes?: number;
     projectRoot?: string;
     runtimes?: RuntimeMap;
   }) {
     this.#maxOutputBytes = opts?.maxOutputBytes ?? 102_400;
+    this.#hardCapBytes = opts?.hardCapBytes ?? 100 * 1024 * 1024; // 100MB
     this.#projectRoot = opts?.projectRoot ?? process.cwd();
     this.#runtimes = opts?.runtimes ?? detectRuntimes();
   }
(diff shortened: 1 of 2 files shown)


mksglu:
## QA Review -- Stream-Level Byte Cap

Thanks for this contribution, @rjkaes! This is a well-identified and well-executed fix for a real vulnerability.

### Bug Verification

**The OOM bug is confirmed real.** The previous implementation in `#spawn` collected all stdout/stderr chunks unconditionally into arrays (`stdoutChunks.push(chunk)`) with no size limit. The `#smartTruncate` method only runs *after* the process exits. A command like `yes` or `cat /dev/urandom | base64` can easily produce multiple gigabytes of output within the default 30-second timeout, causing the Node.js MCP server process to OOM before the timeout fires.

The existing comment in the code actually acknowledged this gap:
> "OOM is bounded by timeout (default 30s) and maxOutputBytes (used only for truncation threshold, not stream limiting)."

The timeout alone is insufficient -- a fast-producing process can fill memory faster than the timeout can kill it.

### Test Coverage Added

I expanded the test file from 7 tests to **16 tests** (pushed to `qa/pr5-stream-byte-cap` branch, commit bb973ef). New test cases include:

| Test | What it validates |
|------|------------------|
| stdout cap | Process killed when stdout exceeds hard cap |
| stderr cap | Process killed when stderr exceeds hard cap |
| combined cap | Cap triggers on total stdout+stderr bytes |
| normal (small) | Small output below cap works correctly |
| normal (moderate) | Moderate output below cap preserves all content |
| memory bounding | Collected stdout bytes stay bounded near cap |
| format (MB value) | Cap message reports correct MB value for 2MB cap |
| format (em dash) | Cap message uses em dash and bracket format |
| timeout (slow) | Timeout still fires when output is slow and under cap |
| timeout (fast) | Cap fires before timeout; `timedOut` stays `false` |
| default cap | Executor works with default `hardCapBytes` |
| truncation interaction | `hardCap` and `maxOutputBytes` work together correctly |
| shell cross-language | Cap works with shell scripts |
| python cross-language | Cap works with python scripts |
| interleaved | Rapid alternating stdout/stderr triggers cap |
| executeFile | `executeFile` also inherits the cap |

### Technical Assessment

The implementation is clean and correct:
- **Stream-level tracking**: `totalBytes` counter across both stdout and stderr prevents memory accumulation
- **Single kill**: `capExceeded` flag ensures `proc.kill("SIGKILL")` is only called once
- **Chunks after cap are dropped**: No additional chunks are pushed after cap is exceeded
- **Diagnostic message**: Clear `[output capped at NMB -- process killed]` appended to stderr
- **Configurable**: `hardCapBytes` constructor option with sensible 100MB default
- **No regressions**: All 55 existing executor tests pass

All 16 QA tests pass. TypeScript type checking is clean.

mksglu:
Merged into `next` branch. Your stream-level byte cap fix has been validated with 16 QA tests — the OOM vulnerability was confirmed real. Thanks @rjkaes!

This will be included in the next release.

mksglu:
@rjkaes Sorry about closing this instead of merging — that was a mistake on my end. Your changes have been merged into the `next` branch and will be included in v0.8.0. Apologies for the confusion!

ca1773130n:
Test note content

ca1773130n:
Test note content

PR #10, opened 2026-03-01
fix: rebundle server.bundle.mjs from cli.ts to expose subcommands (fixes #9)

### LINKED ISSUE #9: v0.7.2/0.7.3: `build/` directory missing from npm package — `doctor` CLI command fails
## Bug Description

The `build/` directory is not included in the published npm package (v0.7.2 installed via Claude Code plugin cache, `package.json` reports v0.7.3). This causes the `doctor` CLI subcommand (and likely other CLI commands) to fail with `MODULE_NOT_FOUND`.

## Steps to Reproduce

1. Install `context-mode` as a Claude Code plugin (v0.7.2)
2. Run the `/context-mode:doctor` skill, which executes:
   ```
   node "<PLUGIN_ROOT>/build/cli.js" doctor
   ```
3. Observe the error:
   ```
   Error: Cannot find module '.../context-mode/0.7.2/build/cli.js'
       at Module._resolveFilename (node:internal/modules/cjs/loader:1455:15)
       ...
     code: 'MODULE_NOT_FOUND'
   ```

## Root Cause

The `package.json` declares `"files": ["build", ...]` and `"bin": {"context-mode": "./build/server.js"}`, but the `build/` directory does not exist in the installed package. Only `src/` contains the TypeScript source files (uncompiled):

```
src/
  cli.ts
  executor.ts
  runtime.ts
  server.ts
  store.ts
```

There is **no** `build/` directory at all:

```
$ ls build/
ls: build/: No such file or directory
```

It appears the `prepublishOnly` script (`npm run build` → `tsc`) either did not run or its output was not included in the published tarball.

## What Works

The MCP server itself works because `start.sh` falls back to `server.bundle.mjs`:

```sh
# start.sh (lines 6-8)
if [ -f server.bundle.mjs ]; then
  [ -d node_modules/better-sqlite3 ] || npm install better-sqlite3 --no-package-lock --no-save --silent 2>/dev/null &
  exec node server.bundle.mjs
fi
```

## What Doesn't Work

- `context-mode doctor` — fails (no `build/cli.js`)
- `context-mode setup` — likely fails (no `build/cli.js`)
- The `bin` entry (`"context-mode": "./build/server.js"`) — broken for global installs
- The `doctor` skill which references `build/cli.js`

## Suggested Fix

Either:

1. **Include compiled output in the npm package**: Ensure `prepublishOnly` actually compiles TypeScript before `npm publish`, and verify `build/` is in the tarball.
2. **Bundle the CLI too**: Similar to how `server.bundle.mjs` is bundled via esbuild, create a `cli.bundle.mjs` and update references.
3. **Update the doctor skill** to use `npx tsx src/cli.ts doctor` as a fallback (matching what `package.json` scripts already do).

## Environment

- Plugin version: 0.7.2 (package.json says 0.7.3)
- Node.js: v25.6.1
- Platform: macOS Darwin 25.3.0
- Installed via: Claude Code plugin marketplace

Wreos:
## Problem

`build/` is excluded by `.gitignore` and is not reliably included in the published npm tarball. Three things were broken on any fresh plugin install:

1. Skills calling `node build/cli.js doctor/upgrade` → `MODULE_NOT_FOUND`
2. `bin` entry `./build/server.js` → `MODULE_NOT_FOUND` on `npm install -g context-mode`
3. `dev` script launched `server.ts` directly, bypassing the CLI routing in `cli.ts`

## Fix

Rebundle the existing `server.bundle.mjs` from `src/cli.ts` instead of `src/server.ts`. Since `cli.ts` routes based on args, one bundle serves all roles:

```
node server.bundle.mjs          → no args → starts MCP server (start.sh, unchanged)
node server.bundle.mjs doctor   → doctor subcommand
node server.bundle.mjs upgrade  → upgrade subcommand
```

No new files. Bundle size is identical (416 KB) since `cli.ts` inlines `server.ts` via its dynamic import.

**Changes in `package.json`:**
- `bin`: `./build/server.js` → `./server.bundle.mjs`
- `bundle` script: entry point `src/server.ts` → `src/cli.ts` (same output filename)
- `prepublishOnly`: added `npm run bundle` so the bundle is always regenerated on publish
- `dev` script: `src/server.ts` → `src/cli.ts`

**Changes in skills:**
- `skills/doctor/SKILL.md`: `build/cli.js` → `server.bundle.mjs`
- `skills/upgrade/SKILL.md`: `build/cli.js` → `server.bundle.mjs`

## Tests

```
$ node server.bundle.mjs doctor
◆  Server test: PASS
◆  Hooks installed: PASS
◆  FTS5 / better-sqlite3: PASS
◆  Marketplace: PASS — v0.7.3

$ node server.bundle.mjs   # no args
Context Mode MCP server v0.7.3 running on stdio
Detected runtimes: JavaScript: bun (1.3.6) ⚡ ...

$ node server.bundle.mjs upgrade
┌   context-mode upgrade
◇  Pulling latest from GitHub...
```

All modes work. `start.sh` is unaffected.

Fixes #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,7 +23,7 @@
   "homepage": "https://github.com/mksglu/claude-context-mode#readme",
   "bugs": "https://github.com/mksglu/claude-context-mode/issues",
   "bin": {
-    "context-mode": "./build/server.js"
+    "context-mode": "./server.bundle.mjs"
   },
   "files": [
     "build",
@@ -40,7 +40,7 @@
     "build": "tsc",
     "bundle": "esbuild src/cli.ts --bundle --platform=node --target=node18 --format=esm --outfile=server.bundle.mjs --external:better-sqlite3 --minify",
     "prepublishOnly": "npm run build && npm run bundle",
-    "dev": "npx tsx src/server.ts",
+    "dev": "npx tsx src/cli.ts",
     "setup": "npx tsx src/cli.ts setup",
     "doctor": "npx tsx src/cli.ts doctor",
     "typecheck": "tsc --noEmit",


mksglu:
Hey @Wreos! Thanks for the PR but why you closed that?

PR #18, opened 2026-03-01
fix: use FTS5 highlight() for stemming-aware snippet extraction

rjkaes:
## Summary

- Replace indexOf-only snippet extraction with positions derived from FTS5 `highlight()` markers, so stemmed matches (e.g., query "configure" matching content "configuration") produce correct snippet windows
- Add `highlight(chunks, 1, char(2), char(3))` to both `search()` and `searchTrigram()` SQL queries, propagating a new `highlighted` field on `SearchResult`
- `extractSnippet` parses STX/ETX markers to find match positions, falling back to indexOf when `highlighted` is absent

## Motivation

`extractSnippet` used `indexOf` to locate query terms in content. When BM25 matched via porter stemming (e.g., query "configure" matching "configuration"), `indexOf` failed and the function fell back to a blind prefix truncation. FTS5 `highlight()` is the authoritative source — it uses the exact same tokenizer that produced the match — so we parse its marker positions instead.

## Why `highlight()` not `snippet()`

FTS5 `snippet()` returns a single best-match window with a fixed token count. The existing `extractSnippet` merges multiple windows for multi-term queries, showing several relevant regions. `highlight()` returns the full content with markers around ALL matched tokens, which we parse and feed into the existing windowing logic.

## Test plan

- [x] 4 unit tests for `positionsFromHighlight` (single marker, multiple markers, no markers, adjacent markers)
- [x] 6 `extractSnippet` tests (highlight preference over indexOf, multi-term windows, indexOf fallback, prefix fallback, short term filtering, short content passthrough)
- [x] 4 store integration tests (porter stemmed markers, trigram markers, end-to-end snippet via store-produced `highlighted`)
- [x] Full test suite green (227 passed, 0 failed)

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -72,31 +72,89 @@ const bunNote = hasBunRuntime()
 // ─────────────────────────────────────────────────────────
 // Helper: smart snippet extraction — returns windows around
 // matching query terms instead of dumb truncation
+//
+// When `highlighted` is provided (from FTS5 `highlight()` with
+// STX/ETX markers), match positions are derived from the markers.
+// This is the authoritative source — FTS5 uses the exact same
+// tokenizer that produced the BM25 match, so stemmed variants
+// like "configuration" matching query "configure" are found
+// correctly. Falls back to indexOf on raw terms when highlighted
+// is absent (non-FTS codepath).
 // ─────────────────────────────────────────────────────────
-function extractSnippet(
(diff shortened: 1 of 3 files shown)


mksglu:
@rjkaes We're live. https://github.com/mksglu/claude-context-mode/releases/tag/v0.9.16

PR #24, opened 2026-03-02
feat: OS-level sandbox enforcement for code execution

rjkaes:
## Summary

- Integrates `@anthropic-ai/sandbox-runtime` to enforce OS-level filesystem and network isolation for all code executed via context-mode's `execute` / `batch_execute` MCP tools
- On macOS uses `sandbox-exec` (Seatbelt), on Linux uses bubblewrap — matching Claude Code's own Bash tool sandbox guarantees
- Network traffic routed through a domain-allowlist proxy (HTTP + SOCKS5), with config sourced from `CONTEXT_MODE_ALLOWED_DOMAINS` env var, `~/.claude/settings.json`, or conservative defaults (GitHub, npm, PyPI, etc.)
- Gracefully degrades to unsandboxed execution when the runtime is unavailable (missing package, unsupported platform)
- Escape hatch via `CONTEXT_MODE_NO_SANDBOX=1`

## Motivation

The PreToolUse hook redirects Claude Code's sandboxed Bash tool calls to context-mode's MCP `execute`/`batch_execute` tools, which previously spawned raw `child_process` with zero OS-level restrictions. This meant prompt injection or AI mistakes could exfiltrate data, write outside the project directory, or make arbitrary network calls — bypassing all of Claude Code's built-in protections.

## Changes

- **`src/sandbox.ts`** (new): Config builder (`buildSandboxConfig`) and lifecycle manager (`initSandbox`) with three-tier domain resolution
- **`src/executor.ts`**: Added `wrapCommand` option to `PolyglotExecutor`; when active, commands are wrapped with sandbox restrictions before spawning. Parent proxy/cert env vars (`HTTP_PROXY`, `SSL_CERT_FILE`, etc.) are excluded when sandboxed to prevent TLS failures from conflicting with the sandbox's own proxy
- **`src/server.ts`**: Sandbox initialized at startup, cleanup on shutdown, status reported in startup log
- **`package.json`**: Added `@anthropic-ai/sandbox-runtime` dependency (external in esbuild bundle — ships precompiled platform binaries)
- **Tests**: 15 unit tests for config builder, 2 for `PolyglotExecutor` sandbox integration, 5 OS-level integration tests (filesystem write/read restrictions, network domain filtering)

## Test plan

- [ ] `npx tsx tests/sandbox.test.ts` — 15 config builder + lifecycle tests
- [ ] `npx tsx tests/executor.test.ts` — 64 executor tests (including 2 sandbox integration)
- [ ] `npx tsx tests/sandbox-integration.test.ts` — 5 OS-level integration tests (requires sandbox-runtime available)
- [ ] `npx tsc --noEmit` — typecheck clean
- [ ] `npm run bundle` — builds successfully
- [ ] Manual: in a new Claude Code instance, run `curl -s https://github.com` via context-mode execute tool — should succeed without TLS errors

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -334,16 +334,25 @@ export class PolyglotExecutor {
       "NPM_TOKEN",
       "NODE_AUTH_TOKEN",
       "npm_config_registry",
-      // General
-      "HTTP_PROXY",
-      "HTTPS_PROXY",
-      "NO_PROXY",
-      "SSL_CERT_FILE",
-      "CURL_CA_BUNDLE",
       // XDG (config paths for gh, gcloud, etc.)
       "XDG_CONFIG_HOME",
       "XDG_DATA_HOME",
     ];
+
+    // When sandbox wrapping is active, wrapWithSandbox() injects its own proxy
+    // env vars (HTTP_PROXY, HTTPS_PROXY, etc.) via the `env` command prefix.
+    // Passing through the parent's proxy/cert vars would be redundant at best
+    // and harmful at worst — e.g. SSL_CERT_FILE may point to a CA bundle that
+    // the seatbelt/bwrap profile blocks, causing TLS verification failures.
+    if (!this.#wrapCommand) {
+      passthrough.push(
+        "HTTP_PROXY",
+        "HTTPS_PROXY",
+        "NO_PROXY",
+        "SSL_CERT_FILE",
+        "CURL_CA_BUNDLE",
+      );
+    }
 
     const env: Record<string, string> = {
       PATH: process.env.PATH ?? (isWin ? "" : "/usr/local/bin:/usr/bin:/bin"),
--- a/tests/sandbox-integration.test.ts
+++ b/tests/sandbox-integration.test.ts
@@ -64,15 +64,15 @@ async function main() {
     assert.ok(r.stdout.includes("hello"));
   });
 
-  await test("allows write to /tmp", async () => {
-    const tmpFile = join(tmpdir(), `sandbox-test-${Date.now()}.txt`);
+  await test("allows write to sandbox TMPDIR (/tmp/claude)", async () => {
+    // The seatbelt profile blocks direct writes to /tmp but allows writing
+    // under /tmp/claude (the TMPDIR that wrapWithSandbox injects).
     const r = await executor.execute({
       language: "shell",
-      code: `echo "tmpwrite" > "${tmpFile}" && cat "${tmpFile}"`,
+      code: `mkdir -p /tmp/claude && echo "tmpwrite" > /tmp/claude/sandbox-test.txt && cat /tmp/claude/sandbox-test.txt`,
     });
     assert.equal(r.exitCode, 0);
     assert.ok(r.stdout.includes("tmpwrite"));
-    try { rmSync(tmpFile, { force: true }); } catch {}
   });
 
   await test("blocks write to home directory", async () => {


mksglu:
## Review: OS-level sandbox enforcement

Thank you for the thorough design work — the problem statement is real and the implementation is well-structured. However, after researching cross-client compatibility, I've decided **not to merge this PR** in its current form. Here's why.

---

### The core issue: Sandbox responsibility belongs to the client, not the server

context-mode is an MCP server. It works (or will work) with **any MCP client** — Claude Code, Codex CLI, Cursor, VS Code, Gemini CLI, Windsurf, OpenCode, and others. The [MCP specification](https://modelcontextprotocol.io/specification/draft/basic/security_best_practices) explicitly places sandboxing responsibility on the **client side**, not the server side.

Most serious MCP clients already provide their own OS-level sandboxing:

| Client | Own Sandbox | Mechanism |
|--------|------------|-----------|
| Claude Code | Yes | macOS Seatbelt + Linux bubblewrap |
| Codex CLI | Yes | macOS Seatbelt + Linux Landlock/bubblewrap |
| VS Code | Yes (Feb 2026) | stdio transport sandbox |
| Gemini CLI | Yes | macOS Seatbelt + Docker |
| Cursor | No (local) | User-managed (devcontainers) |
| Windsurf | No | None |
| OpenCode | No | None |

Adding server-side sandboxing creates **double-sandbox conflicts** with clients that already sandbox.

---

### What we'd give up

**1. Cross-platform support**
`@anthropic-ai/sandbox-runtime` does not support Windows. A significant portion of Cursor and VS Code users are on Windows.

**2. Double-sandbox breakage**
When running inside Claude Code (which already sandboxes MCP servers), our sandbox would nest inside theirs. This causes real problems:
- Gemini CLI has [a known issue](https://github.com/google-gemini/gemini-cli/issues/3261) where nested sandbox restricts path access (exit 126)
- Network proxy conflicts — Claude Code injects its own proxy, `sandbox-runtime` runs its own SOCKS5 proxy
- Filesystem permission intersection failures (client allows `/tmp`, server denies it, or vice versa)

**3. `fetch_and_index` breakage**
Our most valuable tool fetches **arbitrary URLs** — MDN, Supabase docs, GitHub READMEs, random blogs. The sandbox's network allowlist would either:
- Block most URLs by default (breaking the tool), or
- Allow all domains (defeating the purpose of network sandboxing)

**4. Unstable dependency**
- `v0.0.39` — pre-1.0, labeled "Beta Research Preview"
- Active security issues: [#160](https://github.com/anthropic-experimental/sandbox-runtime/issues/160) (process env leak), [#149](https://github.com/anthropic-experimental/sandbox-runtime/issues/149) (denyRead bypass via tmpfs)
- Only 2 primary maintainers, API may change at any time
- Ships under `anthropic-experimental` org — no stability guarantee

**5. Vendor lock-in perception**
The package name `@anthropic-ai/sandbox-runtime` + hardcoded `~/.claude/settings.json` discovery signals Anthropic vendor lock-in. Communities around Cursor, Codex, and other clients may hesitate to adopt an MCP plugin that depends on an Anthropic-specific package.

**6. Dependency weight**
Adds 6 production dependencies including a SOCKS5 proxy server, lodash-es, commander, and shell-quote — significant supply chain surface for a feature that most clients already provide natively.

---

### What we should do instead

1. **Trust the client's sandbox** — this is the MCP-spec-aligned approach
2. **Implement lightweight server-side hardening** without external dependencies:
   - Environment variable sanitization (strip `AWS_SECRET`, `GITHUB_TOKEN`, etc. from subprocess env) — similar to what Gemini CLI does
   - Temp directory cleanup after execution
   - `spawn()` with minimal env instead of inheriting everything
3. **Document security expectations** — clearly state that MCP clients should sandbox context-mode, and list recommended configurations per client
4. **If we ever need our own sandbox**, build a client-agnostic solution (lightweight namespace isolation or Docker-based) without Anthropic-specific dependencies

---

### Summary

The security concern is valid — unsandboxed code execution is a real risk. But the right fix is to ensure clients sandbox us properly, not to implement server-side sandboxing that conflicts with client sandboxes, breaks cross-platform support, and creates vendor lock-in.

rjkaes:
Superseded by #34

PR #25, opened 2026-03-02
fix: use AND instead of OR for multi-word search queries

### LINKED ISSUE #23: `search` returns low-relevance results for multi-word queries (OR instead of AND semantics)
### Problem

`sanitizeQuery()` in `store.ts:98` joins multi-word search terms with `OR`:

```typescript
return words.map((w) => `"${w}"`).join(" OR ");
```

A query like `"useEffect cleanup function"` becomes:

```
"useEffect" OR "cleanup" OR "function"
```

This matches any chunk containing **any** of those words. A chunk that mentions "function" once (in an unrelated context) ranks alongside a chunk that discusses useEffect cleanup functions specifically. BM25 scoring partially compensates — chunks with more matching terms score higher — but single-word matches still appear in the result set and can displace more relevant results when the limit is small (default: 3).

`sanitizeTrigramQuery()` at line 106 has the same issue.

### Expected behavior

Multi-word queries should default to AND semantics — matching only chunks that contain **all** query terms. In FTS5, space-separated quoted terms are implicit AND:

```typescript
return words.map((w) => `"${w}"`).join(" ");
// produces: "useEffect" "cleanup" "function"
// FTS5 interprets as: useEffect AND cleanup AND function
```

### Reproduction

1. Index two documents:

```
# index source="React Hooks Guide"
## useEffect cleanup
Return a cleanup function from useEffect to avoid memory leaks.

## Component lifecycle
Mount and unmount phases control rendering.
```

```
# index source="JavaScript Basics"
## What is a function
A function is a reusable block of code.
```

2. Search for `"useEffect cleanup function"` with `limit: 2`.

3. **Actual result:** Both chunks match — the "What is a function" chunk matches on the single word "function" and may appear in the top 2, displacing the more relevant "Component lifecycle" chunk from the same guide.

4. **Expected result:** Only chunks containing all three terms ("useEffect", "cleanup", "function") match. The "useEffect cleanup" chunk is the only result. The unrelated "What is a function" chunk does not appear.

The root cause is visible in the generated FTS5 query:

```sql
-- Actual (OR):
SELECT * FROM chunks WHERE chunks MATCH '"useEffect" OR "cleanup" OR "function"';
-- Matches: useEffect cleanup chunk, What is a function chunk

-- Expected (AND):
SELECT * FROM chunks WHERE chunks MATCH '"useEffect" "cleanup" "function"';
-- Matches: only chunks containing ALL three terms
```

### Suggested fix

Change the default join to AND (space), and fall back to OR when AND returns no results. This fits naturally into the existing `searchWithFallback` cascade:

```
Layer 1a: Porter + AND  (most precise)
Layer 1b: Porter + OR   (fallback when AND finds nothing)
Layer 2a: Trigram + AND
Layer 2b: Trigram + OR
Layer 3:  Fuzzy correction + re-search
```

Concretely, `sanitizeQuery` could accept a join mode parameter, or the caller (`search` / `searchTrigram`) could try AND first and retry with OR.

### Open questions

1. Should the AND/OR fallback live inside `search()` and `searchTrigram()` themselves, or should `searchWithFallback()` handle it by calling `search()` twice with different query strings?
2. Should `sanitizeTrigramQuery` get the same AND-then-OR treatment, or is OR acceptable for the trigram layer since it's already a fallback?

### Affected code

- `sanitizeQuery()` — `store.ts:98`
- `sanitizeTrigramQuery()` — `store.ts:106`
- `search()` — `store.ts:364` (caller)
- `searchTrigram()` — `store.ts:408` (caller)
- `searchWithFallback()` — `store.ts:488` (orchestrator)

Owen718:
## Summary
Fixes #23 — Multi-word search queries now use AND semantics instead of OR, returning more relevant results when multiple keywords are provided.

## Problem
Previously, `sanitizeQuery()` joined search terms with `OR`, causing queries like `"useEffect cleanup function"` to match any chunk containing **any** of those words. This broadened results instead of narrowing them, contrary to user expectations.

## Changes
- Modified `src/store.ts:98`: Changed `" OR "` → `" AND "`
- Added test case: `multi-word query requires all terms (AND semantics)`

## Before vs After
| Query | Old Behavior | New Behavior |
|-------|-------------|--------------|
| `useEffect cleanup function` | Matches "useEffect" OR "cleanup" OR "function" | Matches "useEffect" AND "cleanup" AND "function" |

## Testing
- New test verifies documents missing query terms are excluded
- Single-word queries remain unaffected (no join applied)
- Empty queries still return empty results

## Checklist
- [x] Code change minimal and focused
- [x] Test added for new behavior
- [x] No breaking changes to existing functionality
- [x] Fixes #23

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -95,7 +95,7 @@ function sanitizeQuery(query: string): string {
     );
 
   if (words.length === 0) return '""';
-  return words.map((w) => `"${w}"`).join(" OR ");
+  return words.map((w) => `"${w}"`).join(" AND ");
 }
 
 function sanitizeTrigramQuery(query: string): string {
--- a/tests/store.test.ts
+++ b/tests/store.test.ts
@@ -511,6 +511,29 @@ async function main() {
     store.close();
   });
 
+  await test("multi-word query requires all terms (AND semantics)", () => {
+    const store = createStore();
+    store.index({
+      content: "# Full Match\n\nuseEffect cleanup function example",
+      source: "doc-with-all-words",
+    });
+    store.index({
+      content: "# Partial Match\n\nfunction example only",
+      source: "doc-with-partial-words",
+    });
+
+    const results = store.search("useEffect cleanup function", 10);
+    assert.ok(
+      results.some((r) => r.source === "doc-with-all-words"),
+      "Should include document that contains all query terms",
+    );
+    assert.ok(
+      !results.some((r) => r.source === "doc-with-partial-words"),
+      "Should not include document missing query terms",
+    );
+    store.close();
+  });
+
   await test("empty query returns empty results", () => {
     const store = createStore();
     store.index({


mksglu:
cc: @rjkaes

mksglu:
Hey @Owen718! We're not ready for PR on this topic. We need to discuss the current issue. We need to test the output quality. We need to add new tests and make sure CI isn't failing. Well done. Thank you.

Owen718:
Thanks for the feedback! Understood — It need careful validation for output quality. Happy to help when the team is ready to move forward.

mksglu:
> Thanks for the feedback! Understood — It need careful validation for output quality. Happy to help when the team is ready to move forward.

Thank you so much. For your helping and support. We need that, really.

PR #29, opened 2026-03-02
fix: pass SSH_AUTH_SOCK and SSH_AGENT_PID through to sandbox

joeinnes:
## Problem

When the sandbox runs a shell command that touches an SSH remote (e.g. `git fetch`, `jj st` on a colocated repo, or `npx` resolving a git dependency), it cannot reach the SSH agent because `SSH_AUTH_SOCK` is stripped from the subprocess environment.

SSH falls back to prompting for the key passphrase directly on the TTY. This hijacks the PTY from Claude Code's session, corrupting its terminal state and effectively crashing it.

## Fix

Add `SSH_AUTH_SOCK` and `SSH_AGENT_PID` to the `buildSafeEnv` passthrough list in `executor.ts`, alongside the existing auth-related vars (`GH_TOKEN`, `AWS_ACCESS_KEY_ID`, etc.).

These vars are only forwarded if already present in the parent environment — users who do not use SSH are unaffected.

## Tests

Three new tests in `executor.test.ts`:
- `SSH_AUTH_SOCK` is passed through when set in the parent environment
- `SSH_AGENT_PID` is passed through when set in the parent environment
- `SSH_AUTH_SOCK` is absent from the subprocess when not set in the parent (no spurious injection)

All 63 tests pass.


mksglu:
## Review: SSH_AUTH_SOCK passthrough

The fix itself is correct — `SSH_AUTH_SOCK` and `SSH_AGENT_PID` absolutely need to reach subprocesses for git/jj SSH operations. The tests are solid. No objections there.

However, this PR highlights a **design problem** with `#buildSafeEnv()`: the current allowlist approach doesn't scale. Every new env var someone needs requires a PR. We've already accumulated 27 manually listed vars, and this pattern will repeat for `GPG_AGENT_INFO`, `CARGO_HOME`, `RUSTUP_HOME`, `VOLTA_HOME`, `FNM_DIR`, `ASDF_DATA_DIR`, `MISE_DATA_DIR`, `NVM_DIR`, `PYENV_ROOT`, `RBENV_ROOT`, and countless others.

### Two better approaches to discuss

**Option A: Denylist (blocklist) — pass everything, block known-dangerous vars**

```typescript
#buildSafeEnv(tmpDir: string): Record<string, string> {
  const deny = new Set([
    // context-mode internals — never leak to user code
    /^CONTEXT_MODE_/,
    /^MCP_/,
    // Prevent recursive subprocess issues
    /^NODE_CHANNEL_/,
  ]);

  const env: Record<string, string> = {};
  for (const [key, val] of Object.entries(process.env)) {
    if (val && !deny.some(rx => rx.test(key))) {
      env[key] = val;
    }
  }

  // Override specifics for sandbox behavior
  env.TMPDIR = tmpDir;
  env.PYTHONDONTWRITEBYTECODE = "1";
  env.PYTHONUNBUFFERED = "1";
  env.NO_COLOR = "1";
  return env;
}
```

**Pros:** Zero maintenance. New tools just work — no PRs needed. Matches how most shells/terminals already behave.

**Cons:** Wider surface area. If a user has something exotic in their env, subprocesses see it. (But they'd see it in normal terminal usage too.)

**Option B: Pattern-based allowlist — group by prefix instead of individual vars**

```typescript
const allowedPrefixes = [
  "SSH_", "GH_", "GITHUB_", "AWS_", "GOOGLE_", "CLOUD",
  "DOCKER_", "KUBE", "NPM_", "NODE_", "XDG_",
  "HTTP_PROXY", "HTTPS_PROXY", "NO_PROXY", "SSL_", "CURL_",
  "GO", "RUST", "CARGO_", "VOLTA_", "FNM_", "ASDF_",
  "PYENV_", "RBENV_", "NVM_", "MISE_", "BUN_",
];
```

**Pros:** Still controlled, but covers entire tool families at once. Adding `SSH_AUTH_SOCK` would have been covered by `SSH_*` from day one.

**Cons:** Still requires maintenance when new tool families emerge. Prefix collisions possible (e.g., `NODE_` catches `NODE_AUTH_TOKEN` but also `NODE_CHANNEL_FD` which we don't want).

---

### My recommendation

**Option A (denylist)** is the right call. Our subprocesses are user-initiated code execution — the user *expects* their full environment to be available, just like running a command in their terminal. The allowlist was defensive but created a worse UX problem: tools randomly failing because some env var didn't make it through.

The denylist only needs to block our own internals (`CONTEXT_MODE_*`, `MCP_*`) and Node IPC channels. Everything else should pass through.

WDYT?. @joeinnes

joeinnes:
Sounds reasonable, only objection I can think of is security. Potentially could be a deny list with patterns to block env vars like `ANTHROPIC_API_KEY` or whatever, but on the other hand, Claude theoretically already has access to those, so there's no greater risk of those leaking into this plugin I don't think. It does increase the attack surface for someone trying to exfiltrate secrets though.

rjkaes:
@mksglu A whitelist is better security, but as noted, it becomes a maintenance nightmare as new tools are added.  Given that, I think the "denylist" is easier to maintain.

mksglu:
> @mksglu A whitelist is better security, but as noted, it becomes a maintenance nightmare as new tools are added. Given that, I think the "denylist" is easier to maintain.

Yes. We need that. Until that, we can merge that PR to `next`. LGTM. cc: @joeinnes

PR #37, opened 2026-03-02
perf(store): cache prepared statements to avoid re-compiling SQL

rjkaes:
## Summary

- Cache all 19 `this.#db.prepare()` calls as 15 class fields initialized once at construction, eliminating SQL re-compilation on every method invocation
- The search hot path (`searchWithFallback`) previously called `prepare()` up to 6 times per query across its porter/trigram/fuzzy cascade — now just field lookups
- Consolidate `getStats()` from three separate queries into a single subquery statement

## Test plan

- [x] `tsc --noEmit` passes clean
- [x] All 232 tests pass across 12 test suites (store, fuzzy-search, search-wiring, search-fallback-integration, executor, etc.)
- [x] `npm run build` succeeds

### DIFF
--- a/hooks/pretooluse.mjs
+++ b/hooks/pretooluse.mjs
@@ -96,9 +96,10 @@ try {
   }
 } catch { /* best effort — don't block hook */ }
 
-let raw = "";
-process.stdin.setEncoding("utf-8");
-for await (const chunk of process.stdin) raw += chunk;
+// Synchronous fd-0 read avoids `for await` on process.stdin which can hang on
+// macOS when stdin is piped via spawnSync (Node async iterator EOF detection
+// is unreliable on some platforms).
+const raw = readFileSync(0, "utf-8");
 
 const input = JSON.parse(raw);
 const tool = input.tool_name ?? "";


mksglu:
Thanks for this, @rjkaes — the prepared statement caching is solid work. Real perf win on the search hot path (`searchWithFallback` was calling `prepare()` up to 6x per query).

**Commit 1 (store.ts):** Merged as-is into `next`. Clean, correct, staff-engineer quality.

**Commit 2 (stdin fix):** We replaced `readFileSync(0)` with an event-based pattern (`on('data')/on('end')`) because `readFileSync(0)` has known cross-platform issues:

- **Windows:** throws `EOF` or `EISDIR` when stdin is a pipe ([nodejs/node#35997](https://github.com/nodejs/node/issues/35997), [nodejs/node#19831](https://github.com/nodejs/node/issues/19831))
- **Linux:** throws `EAGAIN` if `process.stdin` was accessed before the read ([nodejs/node#42826](https://github.com/nodejs/node/issues/42826))

The event-based flowing mode uses a completely different code path from the async iterator (EventEmitter push mode vs `readable` pull mode) and is reliable on all three CI targets (macOS, Ubuntu, Windows). We also applied the same fix to `sessionstart.mjs` which had the identical `for await` pattern.

Our CI matrix runs on `ubuntu-latest`, `macos-latest`, and `windows-latest` — **every PR should consider all three platforms.** When in doubt, check the Node.js issue tracker for platform-specific fd/stdin behavior.

Cherry-picked with full attribution. All 313 tests pass. Pushed to `next` at `5e9a7ea`.

PR #41, opened 2026-03-03
fix: Windows UTF-8 encoding for execute_file + doctor assertion on exit

xxnbyy:
## What

Three fixes — two Windows-specific, one cross-platform CI fix:

1. `execute_file` fails with `UnicodeDecodeError` when reading files containing non-ASCII characters (Chinese, Japanese, Korean, emoji, etc.)
2. `context-mode doctor` prints a libuv assertion error on exit: `Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76`
3. Hook integration test `Bash + curl` times out (5 s) on macOS CI — self-heal block runs `cpSync` on every invocation because the marker file is never written

## Why

**Fix 1 — UTF-8 encoding:**
On Windows, Python/Ruby/Perl/R default to the system locale encoding (e.g. GBK on Chinese Windows) instead of UTF-8. When `execute_file` processes any file with non-ASCII content, the generated wrapper code fails to read it.

Reproducer on Chinese Windows:
```
Error processing file (exit 1):
UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 575: illegal multibyte sequence
```

**Fix 2 — doctor assertion:**
`doctor()` calls `process.exit(code)` while `@clack/prompts` still holds open stdin/stdout handles. On Windows, libuv detects the handle is still in closing state and fires an assertion, printing noise to stderr even when all checks pass.

**Fix 3 — macOS CI timeout:**
In CI (and any non-Claude-Code environment), `~/.claude/plugins/installed_plugins.json` does not exist. The self-heal block called `readFileSync(ipPath)` unconditionally, which threw an error caught by the outer `try/catch` — **preventing the marker file from being written** (line 95 comes after the throw on line 55).

On the next hook invocation the marker still doesn't exist, so self-heal runs again. If `correctDir` also doesn't exist yet (first invocation), it triggers `cpSync` (full directory copy), which takes >5 s on macOS — exceeding the `spawnSync` timeout in the test. The `wget` test passed because by then `correctDir` already existed, skipping the expensive `cpSync`.

## How

**Fix 1** — In `#wrapWithFileContent`, explicitly specify UTF-8 for each affected language:
- **Python**: `open(FILE_CONTENT_PATH, "r", encoding="utf-8")`
- **Ruby**: `File.read(FILE_CONTENT_PATH, encoding: "utf-8")`
- **Perl**: `open(my $fh, '<:encoding(UTF-8)', $FILE_CONTENT_PATH)`
- **R**: `readLines(FILE_CONTENT_PATH, warn=FALSE, encoding="UTF-8")`
- **Defense-in-depth**: add `PYTHONUTF8=1` to `#buildSafeEnv`

JS/TS/Go/Rust/Elixir/PHP are unaffected — already UTF-8 by default.

**Fix 2** — Replace `process.exit(code)` with `process.exitCode = code` in `cli.ts`, allowing Node.js to drain all handles naturally before exiting.

Also fixed the shell runtime detection test to accept Windows full paths (e.g. `C:\Program Files\Git\usr\bin\bash.exe`) — the existing assertion only accepted `"bash"` or `"sh"`, which always failed on Windows even though the runtime works correctly.

**Fix 3** — Wrap the `installed_plugins.json` read/write in an `existsSync` guard in `pretooluse.mjs`. When the file is absent (CI, fresh installs, non-Claude-Code environments), self-heal still completes and writes the marker file, so the expensive setup only runs once.

## Test plan

- [x] `npm run test:all` passes
- [x] `/context-mode:doctor` — all checks PASS, clean exit (EXIT: 0, no assertion)
- [x] Tested in a live Claude Code session on Windows 11
- [x] CI passes on all three platforms (macOS, Ubuntu, Windows)

### Test output

```
=== executor.test.ts ===          76 passed, 0 failed
=== extract-snippet.test.ts ===   14 passed, 0 failed
=== fuzzy-search.test.ts ===      18 passed, 0 failed
=== hook-integration.test.ts ===  14 passed, 0 failed
=== intent-search.test.ts ===      5 passed, 0 failed
=== project-dir.test.ts ===        8 passed, 0 failed
=== search-fallback-integration.test.ts === 7 passed, 0 failed
=== search-wiring.test.ts ===     19 passed, 0 failed
=== store.test.ts ===             53 passed, 0 failed
=== stream-cap.test.ts ===        16 passed, 0 failed
=== subagent-budget.test.ts ===   11 passed, 0 failed
=== turndown.test.ts ===           9 passed, 0 failed
─────────────────────────────────────────────────────
Total: 250 passed, 0 failed (12 test files)
```

New UTF-8 tests added to `executor.test.ts`:
```
✓ execute_file: Python reads UTF-8 non-ASCII content (32ms)
✓ execute_file: JS reads UTF-8 non-ASCII content (109ms)
✓ execute_file: Ruby reads UTF-8 non-ASCII content (124ms)
✓ execute_file: Shell reads UTF-8 non-ASCII content (89ms)
```

### Before/After

**execute_file before** (Chinese Windows, file with Chinese content):
```
UnicodeDecodeError: 'gbk' codec can't decode byte 0x80 in position 575
```

**execute_file after**:
```
总行数: 135
OK - UTF-8 读取成功
```

**doctor before**:
```
└  Diagnostics complete!

Assertion failed: !(handle->flags & UV_HANDLE_CLOSING), file src\win\async.c, line 76
```

**doctor after**:
```
└  Diagnostics complete!

EXIT: 0
```

**hook-integration (macOS CI) before**:
```
✗ Bash + curl: redirected to echo via updatedInput (5024 ms)
  Error: Expected exit 0, got 1
```

**hook-integration (macOS CI) after**:
```
✓ Bash + curl: redirected to echo via updatedInput (38 ms)
```

## Local development setup

- [x] Pointed `installed_plugins.json` installPath to my local clone (modified cache directory in-place — equivalent)
- [x] Bumped version in `package.json` and confirmed with `/context-mode:doctor` — all checks PASS, v0.9.17

## Checklist

- [x] I've checked [existing PRs](https://github.com/mksglu/claude-context-mode/pulls) to make sure this isn't a duplicate
- [x] I've run the full test suite locally (250 tests, 0 failures)
- [x] New functionality includes tests (4 new UTF-8 execute_file tests + fixture)

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -67,14 +67,7 @@ export class PolyglotExecutor {
 
     try {
       const filePath = this.#writeScript(tmpDir, code, language);
-
-      let cmd: string[];
-      try {
-        cmd = buildCommand(this.#runtimes, language, filePath);
-      } catch (err: unknown) {
-        const msg = err instanceof Error ? err.message : String(err);
-        return { exitCode: 1, stdout: "", stderr: msg, timedOut: false };
-      }
+      const cmd = buildCommand(this.#runtimes, language, filePath);
 
       // Rust: compile then run
       if (cmd[0] === "__rust_compile_run__") {
@@ -86,8 +79,8 @@ export class PolyglotExecutor {
       try {
         rmSync(tmpDir, { recursive: true, force: true });
       } catch {
-        // On Windows, EBUSY/EPERM is common due to delayed handle release
-        // after child process exit. Silently ignore — OS cleans temp dirs.
+        // On Windows, bash may still hold file handles when rmSync runs.
+        // Ignore EPERM/EBUSY — the OS will clean up %TEMP% eventually.
       }
     }
   }
@@ -226,7 +219,22 @@ export class PolyglotExecutor {
       // Only .cmd/.bat shims need shell on Windows; real executables don't.
       // Using shell: true globally causes process-tree kill issues with MSYS2/Git Bash.
       const needsShell = isWin && ["tsx", "ts-node", "elixir"].includes(cmd[0]);
-      const proc = spawn(cmd[0], cmd.slice(1), {
+
+      // On Windows with Git Bash, pass the script as `bash -c "source /posix/path"`
+      // rather than `bash /path/to/script.sh`. This avoids MSYS2 path mangling
+      // while still allowing MSYS_NO_PATHCONV to protect non-ASCII paths in commands.
+      let spawnCmd = cmd[0];
+      let spawnArgs: string[];
+      if (isWin && cmd.length === 2 && cmd[1]) {
+        const posixPath = cmd[1].replace(/\\/g, "/");
+        spawnArgs = [posixPath];
+      } else {
+        spawnArgs = isWin
+          ? cmd.slice(1).map(a => a.replace(/\\/g, "/"))
+          : cmd.slice(1);
+      }
+
+      const proc = spawn(spawnCmd, spawnArgs, {
         cwd,
         stdio: ["ignore", "pipe", "pipe"],
         env: this.#buildSafeEnv(cwd),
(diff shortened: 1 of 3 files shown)


mksglu:
@xxnbyy It's going to merge to `next` branch and will be merge to `main` when some features and bugs it's done. LGTM. Thanks.

PR #44, opened 2026-03-03
fix(hooks): intent-aware Read guidance and git command denial in sandbox

rjkaes:
## Summary

- **Intent-aware Read guidance**: Replace the blanket "don't use Read for large files" rule with context-sensitive guidance. When the intent is to edit a file, Read is the correct tool (Edit requires file content in context). When the intent is analysis, `execute_file` is preferred. This prevents the agent from reading files twice — once with `execute_file` then again with `Read` before editing.

- **Git command denial in sandbox tools**: The MCP sandbox (`execute`, `batch_execute`) has no `.git` directory, so git commands fail with a confusing "not a git repository" error. Add detection that denies the call with a clear message redirecting to Bash, saving a wasted tool call.

Both changes update the shared `routing-block.mjs` module so guidance is consistent across `sessionstart.mjs` and `pretooluse.mjs` injection points.

## Test plan

- [ ] Verify that editing a file after `execute_file` analysis no longer triggers a redundant Read nudge
- [ ] Verify that running `git status` in `execute`/`batch_execute` is denied with a message pointing to Bash
- [ ] Verify that `sessionstart` routing block includes updated `forbidden_actions` and `READ_GUIDANCE`

### DIFF
--- a/hooks/pretooluse.mjs
+++ b/hooks/pretooluse.mjs
@@ -251,8 +251,19 @@ if (tool === "Task") {
   });
 }
 
-// ─── MCP execute: security check for shell commands ───
+// ─── MCP execute: git guard + security check for shell commands ───
 if (tool.includes("context-mode") && tool.endsWith("__execute")) {
+  // Git commands don't work in the sandbox (no .git directory)
+  if (toolInput.language === "shell" && /(^|\s|&&|\||\;)git\s/.test(toolInput.code ?? "")) {
+    outputAndExit({
+      hookSpecificOutput: {
+        hookEventName: "PreToolUse",
+        permissionDecision: "deny",
+        reason: "git commands do not work in the sandbox (no .git directory). Use Bash for all git operations (git log, git diff, git status, etc.).",
+      },
+    });
+  }
+
   if (security && toolInput.language === "shell") {
     const code = toolInput.code ?? "";
     const policies = security.readBashPolicies(process.env.CLAUDE_PROJECT_DIR);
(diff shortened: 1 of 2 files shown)


rjkaes:
@mksglu I'm not sure how to handle this, but here's what's happening.

When Claude decides to call `batch_execute` with a bare git call, the command fails:

```
⏺ context-mode-dev - batch_execute (MCP)(commands: [{"label":"Branch history","command":"git log --oneline main..HEAD"},{"label":"SESSION.md in
                                        commits","command":"git log --oneline --all -- SESSION.md"},{"label":"Merge base","command":"git merge-base main
                                        HEAD"}], queries: ["commits containing SESSION.md","merge base with main","full branch history"])
  ⎿  Executed 3 commands (13 lines, 0.3KB). Indexed 3 sections. Searched 3 queries.

     ## Indexed Sections
     … +28 lines (ctrl+o to expand)

⏺ The sandbox doesn't have the git repo. Let me use Bash directly for these git commands.
```

**But** if Claude decides to append a `cd` before the `git` call it succeeds:

```
❯ git diff next..HEAD

⏺ context-mode-dev - execute (MCP)(language: "shell", code: "cd /Users/rjk/src/opensource/claude-context-mode && git diff next..HEAD", intent: "diff changes between next branch and HEAD")
  ⎿  diff --git a/hooks/pretooluse.mjs b/hooks/pretooluse.mjs
     index 99747d2..3fbe518 100644
     --- a/hooks/pretooluse.mjs
     … +49 lines (ctrl+o to expand)
```

I was hoping the nudging would make a difference.  Perhaps the wording could be better?

(The unrelated Read Guidance is *genuinely* helpful and fixes the double-read filling the context.)

mksglu:
@rjkaes Thanks for both changes. Quick review:

**Read guidance** — merging this. It's genuinely the right fix for the double-read problem. No notes.

**Git command denial** — we're going a different direction here. The hard deny with `/(^|\s|&&|\||\;)git\s/` also blocks `cd /project && git status` which actually works. And as you noted, the nudging alone isn't reliable enough.

Instead of denying or nudging, we're fixing the root cause: sandbox commands (`execute`, `batch_execute`) will run in `CLAUDE_PROJECT_DIR` so git works naturally without any `cd` prefix. This makes the git guard unnecessary.

We'll cherry-pick the Read guidance change and apply the working directory fix separately. Your commits will be in the repo.

mksglu:
Merged both changes into `next` at commit 4970027:

1. **Read guidance** — cherry-picked as-is. Your wording is perfect.
2. **Git in sandbox** — instead of the deny guard, we fixed the root cause: shell commands now run in `CLAUDE_PROJECT_DIR` instead of the temp directory. Git works naturally without any `cd` prefix.

Your commits are attributed in the repo. Thanks for both catches — the double-read fix especially is a big quality-of-life improvement.

Closing this PR since the changes are applied. Feel free to verify on `next`.

PR #58, opened 2026-03-04
fix: normalize hook paths to forward slashes on Windows

### LINKED ISSUE #55: [Bug]: Hook paths with backslashes fail in Bash shell on Windows
### context-mode version

0.9.21

### Doctor output

```text
● context-mode doctor

  - Runtimes: 6/11 (javascript, typescript, python, shell, go, perl)
  - Performance: FAST (Bun)
  - Server test: PASS
  - Hooks: PASS (PreToolUse + SessionStart)
  - Plugin enabled: PASS
  - FTS5: PASS
  - npm (MCP): v0.9.21
  - Marketplace: v0.9.21
```

### Prompt that triggered the bug

```text
It is no prompt here, because this is a session start hook error.
```

### What happened?

When context-mode registers hooks in ~/.claude/settings.json, the paths use Windows backslashes:

  "command": "node C:\\Users\\xuhao\\AppData\\Local\\npm-cache\\_npx\\...\\hooks\\sessionstart.mjs"

  Since Claude Code uses Bash (MSYS2) as its shell on Windows, the backslashes are interpreted as
  escape characters and stripped, resulting in:

  Error: Cannot find module 'C:\UsersxuhaoAppDataLocalnpm-cache_npx16b0a7c005a6df70node_modulescontext
  -modehookssessionstart.mjs'

  Workaround:

  Manually change backslashes to forward slashes in settings.json:

  "command": "node C:/Users/xuhao/AppData/Local/npm-cache/_npx/.../hooks/sessionstart.mjs"

### Expected behavior

In the hook registration logic, normalize paths to use forward slashes (/) instead of backslashes,
  since Node.js on Windows supports forward slash paths and it avoids Bash escaping issues.

### Steps to reproduce

Just Start Claude Code in Windows Terminal.

### Debug logs

```text

```

### Operating System

Windows (Native)

### JS Runtime

_No response_

xxnbyy:
## Summary

- `resolve()` on Windows returns backslash paths (e.g. `C:\Users\...\hooks\pretooluse.mjs`)
- These are written directly into `settings.json` as hook `command` strings
- Claude Code executes hooks via Bash (MSYS2), which strips backslashes as escape characters
- Result: `Cannot find module 'C:Users...'` — hooks silently fail on Windows

## Fix

Added `toUnixPath()` helper and applied it to `hookScriptPath` in `upgrade()`:

```ts
export function toUnixPath(p: string): string {
  return p.replace(/\/g, "/");
}

// In upgrade():
const hookScriptPath = toUnixPath(resolve(pluginRoot, "hooks", "pretooluse.mjs"));
```

Node.js on Windows fully supports forward-slash paths, so this is safe cross-platform.

## Tests

Added `tests/cli-hook-path.test.ts` with 5 tests covering:

- backslash → forward slash conversion
- forward-slash paths left unchanged
- mixed slash handling
- pretooluse.mjs command string has no backslashes
- sessionstart.mjs command string has no backslashes

All 5 tests pass. `npm run typecheck` passes. `hook-integration.test.ts` (14/14) unaffected.

Closes #55

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -41,6 +41,11 @@ if (args[0] === "setup") {
  * Shared helpers
  * ------------------------------------------------------- */
 
+/** Normalize Windows backslash paths to forward slashes for Bash (MSYS2) compatibility. */
+export function toUnixPath(p: string): string {
+  return p.replace(/\\/g, "/");
+}
+
 function getPluginRoot(): string {
   const __filename = fileURLToPath(import.meta.url);
   const __dirname = dirname(__filename);
@@ -592,7 +597,7 @@ async function upgrade() {
 
   // Step 4: Fix hooks
   p.log.step("Configuring PreToolUse hooks...");
-  const hookScriptPath = resolve(pluginRoot, "hooks", "pretooluse.mjs");
+  const hookScriptPath = toUnixPath(resolve(pluginRoot, "hooks", "pretooluse.mjs"));
   const settings = readSettings() ?? {};
 
   const desiredHookEntry = {
--- a/tests/cli-hook-path.test.ts
+++ b/tests/cli-hook-path.test.ts
@@ -1,0 +1,97 @@
+/**
+ * CLI Hook Path Tests
+ *
+ * Verifies that hook command paths written to settings.json use forward slashes.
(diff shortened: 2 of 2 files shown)


mksglu:
cc: @xxnbyy

PR #72, opened 2026-03-05
fix: use pathToFileURL for dynamic imports on Windows

skoll43:
## What

Fix `ERR_UNSUPPORTED_ESM_URL_SCHEME` on Windows by using `pathToFileURL(...).href` instead of raw `join()` paths in dynamic `import()` calls.

The idea here is minimal and safe: this is a one-line fix per callsite using the canonical Node.js built-in for path-to-URL conversion. No logic changes, no new dependencies, no behavior differences on macOS/Linux. The only thing that changes is that Node's ESM loader gets a proper `file://` URL instead of a bare Windows path it refuses to load.

## Why

On Windows, Node.js ESM loader rejects absolute paths like `C:\...` in dynamic `import()` expressions with:

```
ERR_UNSUPPORTED_ESM_URL_SCHEME: Only URLs with a scheme in: file, data,
and node are supported. Received protocol 'c:'
```

This caused `ERR_UNSUPPORTED_ESM_URL_SCHEME` to be printed to the terminal after every hook invocation on Windows, and session tracking (PostToolUse, PreCompact, SessionStart) silently failed — `SessionDB` was never loaded.

## How

Wrap all dynamic import paths with `pathToFileURL(...).href` from `node:url`, which converts native OS paths to valid `file://` URLs on all platforms:

```js
// Before (fails on Windows)
await import(join(pluginRoot, scriptPath));

// After (works on all platforms)
await import(pathToFileURL(join(pluginRoot, scriptPath)).href);
```

**Affected files:**
- `src/cli.ts`: `hookDispatch()` used `await import(join(pluginRoot, scriptPath))`
- `hooks/vscode-copilot/sessionstart.mjs`: used `new URL(...).pathname` (gives `/C:/...` on Windows) + `await import(join(PKG_SESSION, ...))` ×3
- `hooks/vscode-copilot/posttooluse.mjs`: `await import(join(PKG_SESSION, ...))` ×2
- `hooks/vscode-copilot/precompact.mjs`: `await import(join(PKG_SESSION, ...))` ×2

No behavior change on macOS/Linux — `pathToFileURL` is the canonical Node.js way to convert paths to ESM-compatible URLs on all platforms.

## Affected platforms

- [ ] Claude Code
- [ ] Gemini CLI
- [x] VS Code Copilot
- [ ] OpenCode
- [ ] Codex CLI

> Note: `src/cli.ts` (`hookDispatch`) is shared across all platforms — if other platforms use hooks with absolute paths on Windows, they benefit from this fix too.

## TDD (required)

> This is a Windows-specific runtime bug triggered by Node.js ESM loader internals. There is no unit test path that can intercept `import()` scheme validation — the error is thrown deep inside `node:internal/modules/esm/load` before user code runs. Verification was done via manual hook invocation.

### RED (failing test)

Reproduced on Windows + Node.js v24.12.0:

```
> echo '{"tool_name":"Read","tool_input":{},"tool_response":"ok"}' | context-mode hook vscode-copilot posttooluse

node:internal/modules/esm/load:195
    throw new ERR_UNSUPPORTED_ESM_URL_SCHEME(parsed, schemes);
          ^

Error [ERR_UNSUPPORTED_ESM_URL_SCHEME]: Only URLs with a scheme in: file, data, and node are
supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs.
Received protocol 'c:'
    at throwIfUnsupportedURLScheme (node:internal/modules/esm/load:195:11)
    ...
  code: 'ERR_UNSUPPORTED_ESM_URL_SCHEME'
```

### GREEN (passing test)

After fix:

```
> echo '{"tool_name":"Read","tool_input":{},"tool_response":"ok"}' | context-mode hook vscode-copilot posttooluse
Exit: 0
```

- [x] I've considered whether my change involves platform-specific behavior — this fix is specifically for Windows path handling in ESM `import()`
- [x] I've asked an AI assistant to review my code for cross-platform issues — `pathToFileURL` is the Node.js built-in designed exactly for this conversion on all OS, no side effects on Unix

## Adapter checklist

- [x] Hook scripts work on all affected platforms (`hooks/vscode-copilot/`)
- [ ] Routing instruction files updated if tool names or behavior changed — N/A, no behavior change
- [ ] Adapter tests pass (`tests/adapters/`) — no adapter logic changed
- [x] `writeRoutingInstructions()` still works for all adapters — not touched

## Test plan

- [ ] `npm test` passes
- [x] `context-mode doctor` — all checks PASS on local build (Windows 11, Node v24.12.0)
- [x] Tested in a live VS Code Copilot session with local MCP server

### Before/After comparison

**Before:** Every tool call in VS Code Copilot printed `ERR_UNSUPPORTED_ESM_URL_SCHEME` and session events were not captured.

**After:** Hooks run silently, session tracking works, `context-mode doctor` exits cleanly.

### Test environment

| | |
|---|---|
| **OS** | Windows 11 |
| **Node.js** | v24.12.0 |
| **npm** | (global install) |
| **VS Code** | Agent mode with GitHub Copilot |
| **context-mode** | v1.0.5 (npm linked local clone) |
## Checklist

- [x] I've checked existing PRs to make sure this isn't a duplicate
- [ ] Tests came first (TDD: red then green) — N/A for this type of runtime error


skoll43:
Small note: CI runs Node 20, but this was only reproduced on Node 24.12.0 — so windows-latest might pass either way. If anyone happens to have Windows + Node 24 around, a quick context-mode hook vscode-copilot posttooluse smoke test would maybe help confirm.

mksglu (changes requested):
## Review — Request Changes

The fix is correct — `pathToFileURL(...).href` is the canonical Node.js solution for Windows ESM dynamic imports. No issues with the approach itself.

### Problem: Fix is incomplete

This PR only patches **VS Code Copilot hooks** and `src/cli.ts`, but the same `await import(join(...))` pattern exists across **all platform hooks**. These will break identically on Windows:

**Gemini CLI** (unfixed):
- `hooks/gemini-cli/aftertool.mjs:29` — `await import(join(PKG_SESSION, "extract.js"))`
- `hooks/gemini-cli/aftertool.mjs:30` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/gemini-cli/sessionstart.mjs:36` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/gemini-cli/sessionstart.mjs:56` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/gemini-cli/sessionstart.mjs:68` — `await import(join(PKG_SESSION, "db.js"))`

**Claude Code** (unfixed):
- `hooks/posttooluse.mjs:25` — `await import(join(PKG_SESSION, "extract.js"))`
- `hooks/posttooluse.mjs:26` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/precompact.mjs:26` — `await import(join(PKG_SESSION, "snapshot.js"))`
- `hooks/precompact.mjs:27` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/sessionstart.mjs:38` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/sessionstart.mjs:59` — `await import(join(PKG_SESSION, "db.js"))`
- `hooks/sessionstart.mjs:72` — `await import(join(PKG_SESSION, "db.js"))`

That's **12 additional callsites** with the exact same bug.

Also check whether `hooks/gemini-cli/sessionstart.mjs` and `hooks/sessionstart.mjs` use `new URL(".", import.meta.url).pathname` for `HOOK_DIR` — if so, they need the same `fileURLToPath()` fix you applied to the VS Code Copilot one.

### Ask

Please apply the `pathToFileURL(...).href` fix to all hook files across all adapters. The fix should be consistent — if a Windows user runs context-mode on any platform, every hook should work.

skoll43:
Applied [pathToFileURL] to all 13 remaining callsites across Claude Code and Gemini CLI hooks — 7 files updated in commit 3d85b3e

PR #73, opened 2026-03-05
chore: add .gitattributes to enforce LF line endings on Windows

skoll43:
What: Add .gitattributes to enforce LF line endings on all platforms.

Why: Without it, git clone on Windows with core.autocrlf=true (the default) converts LF→CRLF on checkout, making every unmodified file appear as changed. This is noisy and can cause issues with shell scripts and .mjs files.

How: Single line * text=auto eol=lf — the canonical Git configuration for this. No behavior change on macOS/Linux.


skoll43:
commit
[ef56650](https://github.com/mksglu/context-mode/commit/ef566503abb926e4f39a7d7f7b63f71b9a6f3f39)

fixed permissions

PR #77, opened 2026-03-06
feat: sql.js fallback when better-sqlite3 native compilation fails

Lets7512:
## Summary
`better-sqlite3` requires glibc 2.29+, Python 3.8+, and gcc to compile. Install fails on CentOS 7/8, Alpine, and systems without build tools. This adds sql.js as a fallback so context-mode works everywhere.

## Changes
- Created `src/sqljs-wrapper.ts` — drop-in wrapper around sql.js that matches better-sqlite3's sync API (prepare/run/get/all/iterate/pragma/transaction)
- Modified `src/db-base.ts` — `loadDatabase()` returns null on failure, `ensureDatabaseReady()` initializes sql.js when needed
- Modified `src/server.ts` — calls `ensureDatabaseReady()` at startup
- Modified `src/store.ts` — graceful FTS5 fallback (sql.js doesn't support FTS5, falls back to LIKE queries)
- Moved `better-sqlite3` to `optionalDependencies`, added `sql.js` as regular dependency
- Added 12 tests covering CRUD, transactions, persistence, and FTS5 limitation

## Trade-offs
- better-sqlite3: ~3-5x faster, native compilation required
- sql.js: universal (WASM/asm.js), zero native deps, slower but functional everywhere
- FTS5 full-text search unavailable in fallback mode (LIKE queries used instead)

## Test plan
- [x] `npx vitest run tests/db-provider.test.ts` — 12/12 passing

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "context-mode",
-  "version": "1.0.2",
+  "version": "1.0.5",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "context-mode",
-      "version": "1.0.2",
+      "version": "1.0.5",
       "license": "Elastic-2.0",
       "dependencies": {
         "@clack/prompts": "^1.0.1",
@@ -14,6 +14,7 @@
         "@modelcontextprotocol/sdk": "^1.26.0",
         "better-sqlite3": "^12.6.2",
         "picocolors": "^1.1.1",
+        "sql.js": "^1.14.1",
         "turndown": "^7.2.0",
         "turndown-plugin-gfm": "^1.0.2",
         "zod": "^3.25.0"
@@ -2571,6 +2572,12 @@
       "engines": {
         "node": ">=0.10.0"
       }
+    },
+    "node_modules/sql.js": {
+      "version": "1.14.1",
+      "resolved": "https://registry.npmjs.org/sql.js/-/sql.js-1.14.1.tgz",
+      "integrity": "sha512-gcj8zBWU5cFsi9WUP+4bFNXAyF1iRpA3LLyS/DP5xlrNzGmPIizUeBggKa8DbDwdqaKwUcTEnChtd2grWo/x/A==",
+      "license": "MIT"
     },
     "node_modules/stackback": {
       "version": "0.0.2",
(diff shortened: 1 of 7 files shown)


Lets7512:
Went through all 6 points — they're valid. The sql.js fallback doesn't hold up under scrutiny.

Evaluated alternatives against your concerns:

| Concern | sql.js (this PR) | sql.js-fts5 | @photostructure/sqlite | **prebuildify** | node:sqlite |
|---|---|---|---|---|---|
| 1. FTS5 | No | Yes | Yes | **Yes** | No |
| 2. WAL/concurrent | No | No | Yes | **Yes** | Yes |
| 3. Package size | +1.5MB | +1.5MB | +5MB | **~0** | 0 |
| 4. Persist overhead | High | High | None | **None** | None |
| 5. Type safety | Double cast | Double cast | Different API | **Full** | Different API |
| 6. Breaking changes | Yes | Yes | Rewrite | **None** | Rewrite |
| CentOS 7 (glibc 2.17) | Works | Works | No (needs 2.31) | **Yes** | Depends |
| Addresses all 6? | 0/6 | 1/6 | 4/6 | **6/6** | 3/6 |

**Prebuildify is the right approach.** Compile better-sqlite3 in CI using Docker containers targeting old glibc:

```
CI (GitHub Actions, one-time setup):
  Docker CentOS 7  → prebuilds/linux-x64/node.napi.glibc.node
  Docker Alpine    → prebuilds/linux-x64/node.napi.musl.node
  macOS runner     → prebuilds/darwin-x64/, darwin-arm64/
  Windows runner   → prebuilds/win32-x64/
  → npm publish (all prebuilts bundled)

Client (every install):
  npm install context-mode
  → node-gyp-build finds matching prebuild → done
  → no Docker, no Python, no GCC needed on client
```

This solves the original problem (CentOS 7/8 can't compile better-sqlite3) without a fallback layer. The binaries compiled against glibc 2.17 work on all newer systems since glibc is backward compatible.

I'll close this PR and open a new one with the prebuildify approach if that works for you.

Lets7512:
Created a separate PR with the prebuildify approach: #81

Instead of the sql.js WASM fallback, it ships pre-compiled `better_sqlite3.node` binaries built on manylinux2014 (glibc 2.17). At runtime, if the default native binding fails, it falls back to our prebuild via better-sqlite3's own `nativeBinding` constructor option. No extra dependencies needed.

PR #78, opened 2026-03-06
fix: prevent EALREADY crash when context-mode used as OpenCode plugin

Lets7512:
## Summary
Adding `"plugin": ["context-mode"]` to `opencode.json` alongside the MCP server config crashes OpenCode with `EALREADY: connection already in progress`.

## Root cause
OpenCode's `"type": "local"` MCP imports the package in-process via the `"."` export rather than spawning a subprocess. When both `mcp` and `plugin` entries reference context-mode, two Bun workers both import `cli.js` through `"."`, both start the MCP stdio transport, and the second one fails with EALREADY because stdin/stdout are already claimed.

## Fix
Changed `"."` and `"main"` in package.json from `cli.js` to `opencode-plugin.js`. This makes `import("context-mode")` return the plugin module (no MCP server side effects). The `bin` field still points to `cli.js`, so the MCP server starts correctly when OpenCode resolves the command.

- Plugin import: `"."` → `opencode-plugin.js` (safe, no stdio binding)
- MCP server: `bin` → `cli.js` (starts server normally)
- CLI usage: unchanged

## Test plan
- [x] OpenCode starts with both `mcp` and `plugin` configured — no crash
- [x] MCP tools available and functional
- [x] Plugin hooks load silently in background

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
   "main": "./build/opencode-plugin.js",
   "exports": {
     ".": "./build/opencode-plugin.js",
-    "./plugin": "./build/opencode-plugin.js"
+    "./plugin": "./build/opencode-plugin.js",
+    "./cli": "./build/cli.js"
   },
   "bin": {
     "context-mode": "./build/cli.js"
--- a/tests/package-exports.test.ts
+++ b/tests/package-exports.test.ts
@@ -1,0 +1,16 @@
+import { describe, test, expect } from "vitest";
+
+describe("Package exports", () => {
+  test("default export exposes ContextModePlugin factory", async () => {
+    const mod = await import("../src/opencode-plugin.js");
+    expect(mod.ContextModePlugin).toBeDefined();
+    expect(typeof mod.ContextModePlugin).toBe("function");
+  });
+
+  test("default export does not leak CLI internals", async () => {
+    const mod = (await import("../src/opencode-plugin.js")) as any;
+    expect(mod.toUnixPath).toBeUndefined();
+    expect(mod.doctor).toBeUndefined();
+    expect(mod.upgrade).toBeUndefined();
+  });
+});


mksglu (commented):
## Review

Good root cause analysis — the EALREADY crash from dual stdio binding is well understood and the fix direction is correct. A few things to address before merge:

### 1. Preserve explicit CLI entry point in exports

After this change, both `"."` and `"./plugin"` resolve to `opencode-plugin.js`. The MCP server entry (`cli.js`) becomes unreachable via programmatic import. We support multiple adapters (Claude Code, Gemini CLI, VS Code Copilot, OpenCode, Codex CLI) — all of them resolve the MCP server through `bin`, but we should keep the CLI accessible as an explicit export for completeness:

```json
"exports": {
  ".": "./build/opencode-plugin.js",
  "./plugin": "./build/opencode-plugin.js",
  "./cli": "./build/cli.js"
}
```

This way `import("context-mode")` is side-effect-free (plugin), `import("context-mode/cli")` explicitly starts the server, and `import("context-mode/plugin")` remains backward-compatible.

### 2. Tests required

This is a behavioral change to the package entry point — it needs test coverage. Please add a `tests/package-exports.test.ts` (or similar) that verifies:

1. **Default export is side-effect-free** — importing `"."` must not bind stdio or start an MCP server
2. **Default export exposes ContextModePlugin** — `typeof ContextModePlugin === "function"`
3. **CLI internals are not leaked** — e.g. `toUnixPath` should not be on the default export

Example:

```ts
import { describe, test, expect } from "vitest";

describe("Package exports", () => {
  test("default export does not start MCP server (side-effect-free)", async () => {
    const mod = await import("../src/opencode-plugin.js");
    expect(mod.ContextModePlugin).toBeDefined();
    expect(typeof mod.ContextModePlugin).toBe("function");
  });

  test("default export does not leak CLI internals", async () => {
    const mod = await import("../src/opencode-plugin.js") as any;
    expect(mod.toUnixPath).toBeUndefined();
  });
});
```

### 3. Minor: `toUnixPath` lives in `cli.ts`

Not blocking, but `toUnixPath` is a pure utility exported from a file with top-level side effects. Consider moving it to a shared utils module in a follow-up — keeps `cli.ts` as a pure entry point and makes the utility safely importable.

---

**Summary:** Fix direction is correct. Please add the `"./cli"` export and the test file, then this is good to merge.

Lets7512:
Addressed all items:

1. **Added `"./cli"` export** — `import("context-mode/cli")` now explicitly reaches the MCP server entry point. Exports map:
   - `"."` → `opencode-plugin.js` (side-effect-free)
   - `"./plugin"` → `opencode-plugin.js` (backward-compatible)
   - `"./cli"` → `cli.js` (MCP server)

2. **Added `tests/package-exports.test.ts`** — 2 tests:
   - Default export exposes `ContextModePlugin` as a function
   - Default export does not leak CLI internals (`toUnixPath`, `doctor`, `upgrade`)

3. **`toUnixPath` in `cli.ts`** — Noted for follow-up, not in scope here.

PR #97, opened 2026-03-11
fix: enforce context-mode routing for verbose Bash commands

benzntech:
## Summary
- Add hard-redirect patterns for Bash commands known to produce large output (test runners, build tools, package managers, docker/k8s, linters, etc.) instead of relying on soft advisory tips
- Strengthen BASH_GUIDANCE with `<priority_instructions>` block to override other plugin/workflow instructions
- Fixes issue where third-party plugins (e.g. GSD) ignore context-mode's soft `<tip>` guidance and use Bash directly for verbose commands

## Test plan
- [ ] Verify commands like `npm test`, `pytest`, `docker logs` get redirected to ctx_execute
- [ ] Verify safe commands (git, mkdir, rm, mv, cd) still pass through normally
- [ ] Verify GSD plugin workflows route verbose commands through context-mode


mksglu:
Hey @benzntech, thanks for the PR and for identifying a real pain point — third-party plugins (like GSD) bypassing our soft routing guidance is a legitimate issue.

After reviewing the implementation, I have some concerns about the hard-redirect approach that I think would cause more problems than it solves:

## 1. Command replacement breaks user intent

The `modify` action replaces the actual command with an `echo` message. This means the user's command **never executes** — they get a text message instead of output. The LLM then has to parse the echo, understand it needs to call a different tool, and retry. This adds a full round-trip and the original command context can get lost.

```js
// User asks for: npm test
// Actually runs: echo "context-mode: This command typically produces verbose output..."
```

If the LLM doesn't understand the redirect message (or retries with Bash), it enters an infinite loop.

## 2. False positives in the regex patterns

Several patterns match commands that produce minimal output:

- **`npm run dev`** / **`npm run build`** — These are long-running processes or commands where you need real-time output. They can't be meaningfully sandboxed.
- **`find . -name "foo.ts"`** — Often returns 1-2 lines
- **`make clean`** — Usually a few lines
- **`cat file | grep pattern`** — Piped output is typically filtered and short
- **`du -sh .`** — Single line output
- **`df -h`** — A handful of lines
- **`ps aux | grep node`** — Piped, usually 1-3 lines

Hard-blocking these would frustrate users who are running perfectly reasonable short-output commands.

## 3. String interpolation security risk

The `safeCmd` escaping only handles `\` and `"`, but Bash commands can contain `$()`, backticks, single quotes, newlines, and other special characters that could break or be interpreted unexpectedly inside the echo string:

```js
const safeCmd = command.replace(/\\/g, "\\\\").replace(/"/g, '\\"');
// Won't safely handle: docker logs $(docker ps -q) or commands with $VAR
```

## 4. The `<priority_instructions>` approach is better but risky

Adding `<priority_instructions>` to BASH_GUIDANCE is a more reasonable direction. However, claiming override authority over "any other plugin or workflow instructions" can cause unpredictable conflicts with other hooks in the user's setup. We intentionally keep our guidance as advisory `<tip>` blocks to be a good citizen in multi-plugin environments.

## Our current design philosophy

The PreToolUse hook uses a **tiered approach** by design:

1. **Hard actions** (`deny` / `modify`) — Only for tool aliasing (e.g., `mcp__context-mode__ctx_search` → `mcp__plugin_context-mode_context-mode__ctx_search`) where we know the redirect is safe and deterministic
2. **Soft guidance** (`context` with `<tip>`) — For Bash, Read, Grep where the command *might* produce large output but we can't be sure

Turning soft guidance into hard blocks would break this contract and create a brittle system that over-intercepts.

## What could work instead

If you're hitting issues with specific plugins ignoring our tips, I'd be interested in exploring:

- **Output-size-based post-hook**: Instead of guessing upfront, we could check the actual output size in `PostToolUse` and suggest sandbox for next time
- **Plugin-specific routing**: If GSD is a common case, we could add explicit GSD-aware routing hints

Thanks again for the contribution and for thinking about this problem. I'm closing this PR but happy to discuss alternative approaches — feel free to open an issue if you want to explore the ideas above!

PR #95, opened 2026-03-10
fix(executor): expose 'path' alias in ctx_execute_file sandbox scripts

flightlesstux:
## What

Expose a `path` variable (alias for `FILE_CONTENT_PATH`) in all `ctx_execute_file` sandbox scripts.

## Why

When calling `ctx_execute_file`, users naturally expect the `path` parameter to be accessible by that name inside their script. Instead, only `FILE_CONTENT_PATH` and `FILE_CONTENT` were injected, causing `NameError`/`ReferenceError` for any code that referenced `path` directly:

```python
# This failed with: NameError: name 'path' is not defined
with open(path) as f:
    data = json.load(f)
```

This was caught in a real session where Claude generated code using `path` after calling `ctx_execute_file` — a completely natural and expected variable name given the tool's parameter is called `path`.

## How

In `#wrapWithFileContent()` (`src/executor.ts`), added `path = FILE_CONTENT_PATH` (and language equivalents) as an alias in the boilerplate prepended to user code for all 11 supported languages:

| Language | Alias added |
|---|---|
| Python | `path = FILE_CONTENT_PATH` |
| JS/TS | `const path = FILE_CONTENT_PATH` |
| Shell | `path=$FILE_CONTENT_PATH` |
| Ruby | `path = FILE_CONTENT_PATH` |
| Go | `var path = FILE_CONTENT_PATH` |
| Rust | `let path = file_content_path` |
| PHP | `$path = $FILE_CONTENT_PATH` |
| Perl | `my $path = $FILE_CONTENT_PATH` |
| R | `path <- FILE_CONTENT_PATH` |
| Elixir | `path = file_content_path` |

`FILE_CONTENT_PATH` and `FILE_CONTENT` remain unchanged — `path` is purely additive.

## Affected platforms

- [x] All platforms / core MCP server

### RED (failing test)

Added 4 tests before touching the implementation:

```
× execute_file: Python exposes 'path' as alias for FILE_CONTENT_PATH 25ms
× execute_file: JS exposes 'path' as alias for FILE_CONTENT_PATH 65ms
× execute_file: Shell exposes 'path' as alias for FILE_CONTENT_PATH 11ms
× execute_file: Ruby exposes 'path' as alias for FILE_CONTENT_PATH 64ms
```

### GREEN (passing test)

```
✓ execute_file: Python exposes 'path' as alias for FILE_CONTENT_PATH 23ms
✓ execute_file: Shell exposes 'path' as alias for FILE_CONTENT_PATH 11ms
[...]

PR #99, opened 2026-03-11
feat(executor): multi-file and directory mode for ctx_execute_file

flightlesstux:
## What

Adds two new input modes to `ctx_execute_file`, making it possible to process multiple files or an entire directory in a single sandbox call.

## Why

Currently `ctx_execute_file` only accepts a single `path`. Real-world workflows often need to compare or correlate multiple files (e.g. two config files, a schema + data file), or process all files in a directory (e.g. all `*.ts` files in `src/`). Without this, callers must chain multiple tool calls and stitch results together in context — exactly what context-mode exists to prevent.

### New modes

| Parameter | Behaviour |
|---|---|
| `path` (existing) | Single file — injects `FILE_CONTENT_PATH`, `FILE_CONTENT`, `file_path` (unchanged) |
| `paths: string[]` | Multiple files — injects a `files` map keyed by the given paths |
| `dir: string` | Directory — globs all non-binary files ≤ 100 KB, injects `files` map, **auto-indexes into FTS5** for follow-up `ctx_search` |
| `glob: string` | Optional filter for `dir` mode (e.g. `**/*.ts`). Default: all non-binary files |

### `files` map per language

| Language | Variable |
|---|---|
| Python | `files: dict[str, str]` |
| JS/TS | `const files: Record<string, string>` |
| Shell | `FILE_COUNT`, `FILE_0_PATH`, `FILE_0_CONTENT`, `FILE_1_PATH`, … |
| Ruby | `files: Hash` |
| Go | `var files map[string]string` |
| Rust | `let mut files: HashMap<String, String>` |
| PHP | `$files: array` |
| Perl | `%files` |
| R | `files: list` |
| Elixir | `files: map` |

### dir mode auto-indexing

When `dir` is used, every resolved file (not just those injected into the sandbox) is indexed into the FTS5 store. This means `ctx_search` works immediately after the call — no separate `ctx_index` needed. Files exceeding the 1 MB injection budget are still indexed but not injected.

### Key implementation details (`src/executor.ts`)

- **`#resolveFiles()`** — resolves `paths`/`dir` opts into a content map + absolute path list. Skips binary files (null-byte scan of first 8 KB), caps per-file at 100 KB and total injection at 1 MB.
- **`#wrapWithMultipleFiles()`** — generates language-appropriate `files` boilerplate for all 11 languages.
- **`globToRegex()`** — lightweight glob→RegExp (no new dependencies). Handles `**/*.ext`, `*.ext`, `src/**`.
- **`isBinary()`** — null-byte heuristic.
- `executeFile()` return type extended with `resolvedPaths?: string[]` (server uses this for FTS5 indexing; single-file mode unchanged).

### Backward compatibility

The `path` parameter is now optional (was required). All existing single-file behaviour is unchanged. A runtime validation error is returned if none of `path`, `paths`, or `dir` is provided.

## Affected platforms

### RED (failing before implementation)

```
× paths: Python accesses both files via 'files' map
× paths: Shell accesses both files via FILE_0/FILE_1 indexed vars
× paths: Ruby accesses both files via 'files' map
× paths: returns resolvedPaths with absolute paths
× paths: binary files are skipped
× dir: Python sees all non-binary files via 'files' map
× dir: glob filter limits to matching files only
× dir: returns resolvedPaths for all matched files
× dir: empty result when glob matches nothing
```

### GREEN (after implementation)

```
✓ paths: Python accesses both files via 'files' map 36ms
✓ paths: Shell accesses both files via FILE_0/FILE_1 indexed vars 19ms
✓ paths: Ruby accesses both files via 'files' map 91ms
✓ paths: returns resolvedPaths with absolute paths 12ms
✓ paths: binary files are skipped 11ms
✓ dir: Python sees all non-binary files via 'files' map 32ms
✓ dir: glob filter limits to matching files only 31ms
✓ dir: returns resolvedPaths for all matched files 11ms
✓ dir: empty result when glob matches nothing 11ms
```

### Full test run

```
Tests  10 failed | 80 passed | 14 skipped (104)
```

Pre-existing failure count unchanged (10). 9 new tests all green.

- [x] Shell indexed vars use single-quoting (same pattern as existing `FILE_CONTENT_PATH` injection) — safe against `$`, backtick, `!` expansion and special chars in file content
- [x] Windows path separators normalised to `/` in `files` map keys (`rel.replace(/\\/g, "/")`)
- [x] `readdirSync({ recursive: true })` available Node 18.17+ (matches existing engine requirements)
- [x] No new runtime dependencies added

## Checklist

- [x] I've checked existing PRs to make sure this isn't a duplicate
- [x] No breaking changes — `path` single-file mode is fully backward compatible
- [x] `FILE_CONTENT_PATH`, `FILE_CONTENT`, `file_path` unchanged in single-file mode

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -719,7 +719,7 @@ export class PolyglotExecutor {
       case "go":
         return `package main\n\nimport (\n\t"fmt"\n\t"os"\n)\n\nvar FILE_CONTENT_PATH = ${escaped}\nvar file_path = FILE_CONTENT_PATH\n\nfunc main() {\n\tb, _ := os.ReadFile(FILE_CONTENT_PATH)\n\tFILE_CONTENT := string(b)\n\t_ = FILE_CONTENT\n\t_ = fmt.Sprint()\n${code}\n}\n`;
       case "rust":
-        return `use std::fs;\n\nfn main() {\n    let file_content_path = ${escaped};\n    let file_path = file_content_path;\n    let file_content = fs::read_to_string(file_content_path).unwrap();\n${code}\n}\n`;
+        return `#![allow(unused_variables, dead_code)]\nuse std::fs;\n\nfn main() {\n    let file_content_path = ${escaped};\n    let file_path = file_content_path;\n    let file_content = fs::read_to_string(file_content_path).unwrap();\n${code}\n}\n`;
       case "php":
         return `<?php\n$FILE_CONTENT_PATH = ${escaped};\n$file_path = $FILE_CONTENT_PATH;\n$FILE_CONTENT = file_get_contents($FILE_CONTENT_PATH);\n${code}`;
       case "perl":


mksglu:
## Review — Claude Code Engineer + Adapter Engineers + Context Mode Architect

Hey @flightlesstux, thanks for another well-structured PR! After review by three engineering teams, we want to share both the technical findings and a broader architectural perspective.

---

### The core question: does this need to exist?

context-mode already has dedicated tools that cover these workflows:

| What you'd use PR #99 for | Existing tool |
|---|---|
| Process 2-3 files together | `ctx_batch_execute(commands: ["cat a.ts", "cat b.ts"], queries: ["compare"])` — runs both, auto-indexes, searchable |
| Scan a directory | `ctx_execute(language: "shell", code: "find src/ -name '*.ts' \| head -50")` — full shell power, no file count limits |
| Index a directory into FTS5 | `ctx_index({ path: "src/" })` — purpose-built for exactly this, handles recursion, binary detection, and chunking |
| Compare config files | `ctx_execute(language: "python", code: "import json; a = json.load(open('a.json')); ...")` — full language stdlib available |

The `ctx_execute_file` tool was designed with a single responsibility: "inject one file's content into a sandbox and run code against it." This simplicity is a feature — it keeps the attack surface small and the mental model clear.

---

### If we were to proceed: technical blockers

Our review teams (Claude Code Engineer, Adapter Engineers, Context Mode Architect) identified 4 blockers that would need to be resolved:

**B1. Path traversal on all resolved paths**

Every file path that reaches `readFileSync` must be validated against the project root. Currently the server's deny-policy check only validates the top-level `dir` argument, not each file resolved from within. `resolve(projectRoot, "../../etc/passwd")` escapes the sandbox.

```ts
const abs = resolve(projectRoot, p);
if (!abs.startsWith(projectRoot + sep) && abs !== projectRoot) {
  throw new Error(`Path escapes project root: ${p}`);
}
```

**B2. Symlink traversal in dir mode**

`readdirSync({ recursive: true })` follows symlinks. A symlink to `/etc` or `../` escapes the project root. Fix: use `lstatSync` to detect and skip symlinks.

**B3. File count cap in dir mode**

No limit means `dir: "node_modules"` will attempt to stat/read 100K+ files. Needs a hard cap (~500 files) with a clear error telling the user to narrow with a `glob` filter.

**B4. Mutual exclusivity validation**

If more than one of `path`/`paths`/`dir` is provided, behavior is undefined (silently falls to single-file). Also `paths: []` (empty array) is truthy and passes the filter but is a no-op. Both cases should reject with a clear error.

---

### Why the blockers reinforce the architectural point

These 4 issues aren't implementation bugs — they're inherent complexity that comes with expanding a single-file tool into a multi-file orchestrator. The existing tools (`ctx_execute`, `ctx_batch_execute`, `ctx_index`) already solved these problems at the right abstraction layer:

- `ctx_execute` gives full language runtimes where *your code* controls file I/O — no path injection needed
- `ctx_batch_execute` handles parallel operations with built-in indexing
- `ctx_index` handles directory recursion with proper binary detection, chunking, and FTS5 integration

Adding the same capabilities to `ctx_execute_file` would duplicate logic that already exists and is battle-tested.

---

### Bottom line

We recommend closing this PR — not because the implementation is bad (it's well-structured and the TDD approach was solid), but because the feature adds complexity to a tool that benefits from being simple, while the use cases are already well-served by the existing toolkit.

Really appreciate the effort and the thorough PR description. Your `file_path` alias (#95) shipped in v1.0.18 and that was exactly the kind of focused improvement that makes context-mode better. Looking forward to more of those!

flightlesstux:
Thanks for the detailed review! We've addressed all 4 blockers in commit [0f8fd56](https://github.com/flightlesstux/context-mode/commit/0f8fd56):

**B1 — Path traversal**
Relative paths (e.g. `../../etc/passwd`) are now validated against `projectRoot` in all three modes (`path`, `paths`, `dir`). Absolute paths pass through unchanged to preserve existing behavior.

**B2 — Symlink traversal**
Replaced `readdirSync({ recursive: true })` with a new private `#safeReaddirFiles` helper that manually recurses using `lstatSync`, skipping any entry where `isSymbolicLink() === true`. Symlinks are never followed.

**B3 — File count cap**
`#safeReaddirFiles` stops collecting at 501 entries. If the limit is exceeded, a clear error is thrown pointing the user to the `glob` option:
> *"Directory contains more than 500 files, which exceeds the limit of 500. Use the 'glob' option to narrow the selection (e.g. glob: "**/*.ts")."*

**B4 — Mutual exclusivity**
`executeFile` now throws immediately if more than one of `path`/`paths`/`dir` is provided, and also rejects `paths: []` (empty array).

All 4 fixes are covered by new tests (B1–B4 labeled) and all 99 tests pass.

Regarding the architectural point — we hear you, and the existing tools (`ctx_batch_execute`, `ctx_index`, `ctx_execute`) do cover many of these workflows. That said, `ctx_execute_file` with multi-path support reduces the boilerplate for the common case of "run this code against these specific files" without requiring the user to compose shell commands or know the right tool for each scenario. We think the two can coexist cleanly now that the security surface is properly bounded.

mksglu:
@flightlesstux — first off, the implementation quality here is genuinely impressive. All four security blockers addressed with tests, clean code structure, and the `#safeReaddirFiles` symlink handling is exactly how I'd write it. This isn't a code quality issue at all.

But I want to walk through why we're not merging this, with technical specifics.

### 1. MCP tool schema is a per-interaction tax

In MCP, every registered tool's full JSON schema — parameters, descriptions, enums — is sent to the LLM on **every single interaction**. It's not loaded once; it's part of the system prompt on every turn. Adding `paths: string[]`, `dir: string`, and `glob: string` to `ctx_execute_file` increases the schema payload that all users pay on every message. For a context-saving tool, this is a direct contradiction of our core value proposition.

### 2. Multi-mode tools degrade LLM routing accuracy

When a tool has one clear purpose ("process this file with code"), the model routes to it reliably. When the same tool has 3 input modes with mutual exclusivity constraints, the model has to reason about which mode to use — and sometimes picks the wrong one. We've seen this empirically: simpler tool interfaces → higher routing accuracy → better context savings.

`ctx_execute_file(path)` is unambiguous. `ctx_execute_file(paths? | dir? | glob?)` requires the model to understand exclusivity rules that aren't natural to express in JSON Schema.

### 3. The injection model doesn't scale to N files

`ctx_execute_file` works by **injecting file content as string literals into generated code**. For a single file, this is clean — `FILE_CONTENT` is a variable. For N files, you're `JSON.stringify`-ing a map and embedding it as a code literal. This is why you need the 100KB/file and 1MB/total caps — the injection approach is fundamentally bounded.

Compare with the existing alternative:

```python
# ctx_execute — direct file I/O, no injection limits
import json, pathlib
files = {str(p): p.read_text() for p in pathlib.Path("src").rglob("*.json")}
# process files directly — no serialization overhead, no caps
```

Direct file I/O in `ctx_execute` has no artificial limits because it doesn't inject anything. The code reads what it needs.

### 4. The `killTree` fix already exists in our codebase

We already use `process.kill(-proc.pid, "SIGKILL")` with `detached: !isWin` — same pattern you have. So there's nothing to extract here either.

---

The boilerplate reduction argument is valid in isolation, but the tradeoffs (schema bloat × routing degradation × injection scaling limits × duplicated template maintenance for 11 languages × 2 modes) outweigh it. The existing tools cover these workflows without these costs.

I'm closing this PR. Thank you for the thorough work and especially for addressing all the security concerns — it shows real engineering discipline.

PR #101, opened 2026-03-12
fix(batch_execute): coerce string inputs for commands and queries params

flightlesstux:
## What

Add `z.preprocess()` coercion to `ctx_batch_execute` for both `commands` and `queries` parameters so LLM string inputs are normalized rather than rejected.

## Why

LLMs sometimes pass `commands` and `queries` as JSON strings instead of arrays (producing MCP error `-32602: Expected array, received string`), or pass `commands` as an array of plain strings instead of `{label, command}` objects. This caused hard failures with no useful recovery path. Fixes the same class of bug that `ctx_search` already handles via its own normalization at line 934.

## How

- Wrapped `commands` with `z.preprocess()` that handles two cases:
  1. JSON string → `JSON.parse()` → array
  2. Array of plain strings → auto-wrapped as `{label: "cmd N", command: str}`
- Wrapped `queries` with `z.preprocess()` that handles:
  1. JSON string → `JSON.parse()` → array
  2. Bare string → single-element array `[str]`
- Valid array inputs pass through unchanged — no breaking change.

## Affected platforms

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] VS Code Copilot
- [ ] OpenCode
- [ ] Codex CLI

## TDD (required)

New test file: `tests/batch-execute-coercion.test.ts` (9 tests, red-green-refactor).

### RED (failing test)

Tests `commandsSchemaStrict` and `queriesSchemaStrict` (plain `z.array()` without preprocess) to prove they reject string inputs — confirming the bug existed:

```
✓ strict schema rejects JSON string input
✓ strict schema rejects array of plain strings
✓ strict schema rejects JSON string input  (queries)
```

These pass because `safeParse(...).success === false` — i.e. the schema correctly fails on bad input.

### GREEN (passing test)

After applying `z.preprocess()`:

```
✓ tests/batch-execute-coercion.test.ts (9 tests) 3ms
```

All 9 tests pass, including coercion of JSON strings, plain string arrays, and valid passthrough.

Change is pure Zod schema preprocessing — no file I/O, no child processes, no path handling. Cross-platform safe.

## Adapter checklist

No hook or adapter changes — tool name and interface are unchanged.

## Test plan

- [x] `npm test` passes (844 passing, 24 pre-existing failures in `executor`/`turndown`/`stream-cap`/`project-dir` unrelated to this change)
- [x] `npm run typecheck` passes (clean build)
- [x] Tested in a live session with my local MCP server

### Test output

```
✓ tests/batch-execute-coercion.test.ts (9 tests) 3ms

Test Files  4 failed | 37 passed (41)
     Tests  24 failed | 844 passed | 14 skipped (882)
```

(4 failing files are pre-existing, unrelated to this PR)

### Before/After comparison

**Before:** Passing `commands` as a JSON string produced:
```
MCP error -32602: Input validation error: Expected array, received string
```

**After:** Same call succeeds and executes normally.

## Local development setup

- [x] Symlinked plugin cache to my local clone
- [x] Deleted `server.bundle.mjs` so `start.mjs` uses `build/server.js`

## Checklist

### DIFF
--- a/tests/batch-execute-coercion.test.ts
+++ b/tests/batch-execute-coercion.test.ts
@@ -1,0 +1,131 @@
+/**
+ * Tests for ctx_batch_execute input coercion (string → array normalization).
+ *
+ * LLMs sometimes pass `commands` and `queries` as JSON strings instead of
+ * arrays, or pass commands as plain strings instead of {label, command}
+ * objects. These tests verify that z.preprocess() coerces all three cases
+ * gracefully rather than throwing -32602 validation errors.
+ */
+
+import { describe, test, expect } from "vitest";
+import { z } from "zod";
+
+// ── Replicate the schemas from server.ts ─────────────────────────────────────
+
+const commandItemSchema = z.object({
+  label: z.string(),
+  command: z.string(),
+});
+
+const commandsSchemaStrict = z.array(commandItemSchema).min(1);
+
+const commandsSchemaCoerced = z.preprocess((val) => {
+  let arr = val;
+  if (typeof val === "string") {
+    try {
+      arr = JSON.parse(val);
+    } catch {
+      return val;
+    }
+  }
+  if (Array.isArray(arr)) {
+    return arr.map((item, i) =>
+      typeof item === "string" ? { label: `cmd ${i + 1}`, command: item } : item,
+    );
+  }
+  return arr;
+}, z.array(commandItemSchema).min(1));
+
+const queriesSchemaStrict = z.array(z.string()).min(1);
+
+const queriesSchemaCoerced = z.preprocess((val) => {
+  if (typeof val === "string") {
+    try {
+      const parsed = JSON.parse(val);
+      return Array.isArray(parsed) ? parsed : [val];
+    } catch {
+      return [val];
+    }
+  }
+  return val;
+}, z.array(z.string()).min(1));
+
+// ── commands coercion ─────────────────────────────────────────────────────────
+
+describe("ctx_batch_execute commands coercion", () => {
+  test("strict schema rejects JSON string input", () => {
+    const input = JSON.stringify([{ label: "test", command: "echo hi" }]);
+    const result = commandsSchemaStrict.safeParse(input);
+    expect(result.success).toBe(false);
+  });
+
+  test("strict schema rejects array of plain strings", () => {
+    const result = commandsSchemaStrict.safeParse(["echo hi", "echo bye"]);
+    expect(result.success).toBe(false);
+  });
+
+  test("coerced schema accepts JSON string of {label,command} objects", () => {
+    const input = JSON.stringify([{ label: "greet", command: "echo hi" }]);
+    const result = commandsSchemaCoerced.safeParse(input);
+    expect(result.success).toBe(true);
+    if (result.success) {
+      expect(result.data).toEqual([{ label: "greet", command: "echo hi" }]);
+    }
+  });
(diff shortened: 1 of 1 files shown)


mksglu:
@flightlesstux — this PR bundles two features, so let me address each separately.

### 1. Batch execute string coercion — already merged

PR #126 by @jeanmonet landed on `next` on 2026-03-15. It adds extracted `coerceJsonArray()` and `coerceCommandsArray()` helpers with `z.preprocess()` covering both `ctx_batch_execute` (commands + queries) and `ctx_search` (queries). Your inline implementation covers the same ground but with a narrower scope (batch_execute only) and a label format inconsistency (`"cmd 1"` vs the merged `"cmd_1"`).

The coercion feature is fully resolved — no action needed here.

### 2. Multi-file/directory mode for `ctx_execute_file` — closed in PR #99

This is architecturally the same feature we discussed and closed in #99. The security fixes (B1-B4) are solid, but the core architectural concerns remain:

**MCP schema is a per-interaction tax.** Every tool's full JSON schema — parameter names, types, descriptions, constraints — is serialized into the LLM system prompt on every single message. Adding `paths: string[]`, `dir: string`, and `glob: string` to `ctx_execute_file` increases this payload for all users on every turn. For a tool whose entire purpose is context savings, expanding the schema footprint is a direct tradeoff against our own value proposition. This isn't theoretical — we measure token consumption and every parameter adds to it.

**Multi-mode tools degrade routing accuracy.** When `ctx_execute_file` accepts one parameter (`path`), the model routes to it unambiguously. With three mutually exclusive input modes, the model must reason about which mode to use — and mutual exclusivity constraints don't map cleanly to JSON Schema (there's no `oneOf` for parameter groups in MCP tool schemas). In practice, this means the model sometimes picks `ctx_execute_file(paths: [...])` when `ctx_batch_execute` would be more appropriate, or constructs invalid calls with multiple modes set. Simpler tool interfaces → higher routing accuracy → fewer wasted turns.

**The injection model has fundamental scaling limits.** `ctx_execute_file` works by embedding file content as string literals in generated code — `FILE_CONTENT = "..."` for single files, `JSON.parse('{"a.json": "..."}')` for multi-file. This serialization-into-code approach is why you need the 100KB/file and 1MB/total caps. Compare with direct file I/O in `ctx_execute`:

```python
# No injection, no caps, no serialization overhead
import pathlib, json
files = {str(p): json.loads(p.read_text()) for p in pathlib.Path("src").rglob("*.json")}
# Full programmatic control — read what you need, skip what you don't
```

Direct I/O has no artificial limits because the code controls its own file access. The injection approach adds a serialization layer that doesn't scale to N files without increasingly aggressive caps.

**11 × 2 template maintenance.** `#wrapWithMultipleFiles()` adds a second code template for every supported language. Any new language addition requires both `#wrapWithFileContent()` and `#wrapWithMultipleFiles()` — and they must stay in sync. This is a real maintenance cost that compounds over time.

### 3. `killTree` / `detached` fix — already exists

Our `next` branch already has `process.kill(-proc.pid, "SIGKILL")` with `detached: !isWin` at executor.ts lines 18-29 and 226. Same pattern — no delta here.

---

All three components of this PR are either already shipped or architecturally precluded. I'm closing this — but I want to be clear that the implementation quality is high. The B1-B4 security hardening, cross-platform test coverage, and `#safeReaddirFiles` approach are all well-engineered. Thank you for the effort.

Issue #86, opened 2026-03-07
Add NODE_EXTRA_CA_CERTS to subprocess environment passthrough list


ayeganov:
## Problem

When running inside an environment that performs TLS interception (e.g., a corporate proxy or a security sandbox with an ephemeral CA), `ctx_fetch_and_index` fails with:

```
TypeError: fetch failed
  [cause]: Error: self-signed certificate in certificate chain
      at TLSSocket.onConnectSecure (node:_tls_wrap:1679:34)
```

## Root Cause

The `PolyglotExecutor.#buildSafeEnv()` method in `src/executor.ts` constructs a curated environment for spawned subprocesses. The passthrough list includes `SSL_CERT_FILE` and `CURL_CA_BUNDLE`, but **not `NODE_EXTRA_CA_CERTS`**.

Node.js's built-in `fetch()` (undici) relies on `NODE_EXTRA_CA_CERTS` to trust additional certificate authorities. When this env var is stripped from the subprocess environment, any TLS-intercepting proxy or custom CA in the chain causes fetch to reject the connection.

The parent process has `NODE_EXTRA_CA_CERTS` set and pointing to the system CA bundle (which includes the intercepting CA), but the subprocess never sees it.

## Suggested Fix

Add `NODE_EXTRA_CA_CERTS` (and optionally `REQUESTS_CA_BUNDLE` for Python `requests` library parity) to the passthrough array in `#buildSafeEnv()`:

```diff
       "SSL_CERT_FILE",
       "CURL_CA_BUNDLE",
+      "NODE_EXTRA_CA_CERTS",
+      "REQUESTS_CA_BUNDLE",
```

## Context

I ran into this while using context-mode inside a TLS-intercepting security sandbox. The `https` module (which respects `NODE_EXTRA_CA_CERTS` in the parent process) works fine, but any subprocess spawned by the executor that uses `fetch()` fails because the env var is missing.

I've patched this locally in both `server.bundle.mjs` and `src/executor.ts` and confirmed it resolves the issue.

Is this something you'd consider a valuable change to merge upstream? Happy to open a PR if so.

mksglu:
Fixed in v1.0.15 🎉

Added `NODE_EXTRA_CA_CERTS` and `REQUESTS_CA_BUNDLE` to the `#buildSafeEnv()` passthrough list in `src/executor.ts`.

Update via:
- **Claude Code Marketplace:** `/upgrade context-mode` or `/context-mode:ctx-upgrade`
[...]

Issue #22, opened 2026-03-01
The sandboxing doesn't play well with virtual environments like direnv and nix devshells


zaphar:
Tools like direnv or nix develop depend on setting a number of environment variables as well as access to paths that may not exist in the actual repository. Context mode doesn't perform well in those cases. Particularly it appears to be dropping env variables that are necessary and potentially access to paths that are necessary for the virtual environment to work.

mksglu:
Thanks for reporting this @zaphar. You're right — confirmed bug.

### Root cause

context-mode's executor uses a **whitelist-based environment** (`#buildSafeEnv()` in `src/executor.ts`). Only specific env vars (auth tokens, proxy settings, SSH agent) are passed through to sandboxed subprocesses. Everything else is dropped.

This means **direnv/nix devshell variables** like `DIRENV_DIR`, `NIX_PATH`, `IN_NIX_SHELL`, `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, `PKG_CONFIG_PATH`, `VIRTUAL_ENV`, etc. are all lost — so tools installed via nix or activated via direnv aren't reachable from the sandbox.

### Fix

Added 30+ virtual environment variables to the passthrough list:

- **direnv**: `DIRENV_DIR`, `DIRENV_FILE`, `DIRENV_DIFF`, `DIRENV_WATCHES`, `DIRENV_LAYOUT_DIR`
- **Nix**: `NIX_PATH`, `NIX_PROFILES`, `NIX_SSL_CERT_FILE`, `NIX_CC`, `NIX_STORE`, `IN_NIX_SHELL`, `LOCALE_ARCHIVE`
- **Library paths**: `LD_LIBRARY_PATH`, `DYLD_LIBRARY_PATH`, `LIBRARY_PATH`, `C_INCLUDE_PATH`, `CPLUS_INCLUDE_PATH`, `PKG_CONFIG_PATH`, `CMAKE_PREFIX_PATH`
- **Language runtimes**: `GOPATH`, `GOROOT`, `CARGO_HOME`, `RUSTUP_HOME`, `VIRTUAL_ENV`, `CONDA_PREFIX`, `PYTHONPATH`, `GEM_HOME`, `JAVA_HOME`, `SDKMAN_DIR`
- **Version managers**: `ASDF_DIR`, `ASDF_DATA_DIR`, `MISE_DATA_DIR`, `RBENV_ROOT`

[...]

PR #102, opened 2026-03-12
feat: configurable env var passthrough for sandboxed processes

dandaka:
## Summary

The hardcoded env whitelist in `#buildSafeEnv()` prevents users from passing custom environment variables (e.g. API tokens loaded via secret managers like Infisical, Vault, or Doppler) to sandboxed subprocesses. This forces users to fall back to reading secrets from files on disk — which is a **worse security posture** than inheriting them from process memory.

This PR adds a configurable `envPassthrough` option:

- **`envPassthrough: true`** — inherit ALL parent env vars (no filtering)
- **`envPassthrough: ["SLACK_*", "LINEAR_*"]`** — pass through vars matching glob patterns, in addition to the built-in whitelist
- **`CONTEXT_MODE_ENV_PASSTHROUGH` env var** — configure without code changes:
  - `"*"` → pass all
  - `"SLACK_*,LINEAR_*,MY_TOKEN"` → comma-separated patterns

Constructor option takes precedence over the env var.

### Why this matters

Many users manage secrets via secret managers that inject env vars into the process at startup. The current approach forces them to either:
1. Write secrets to `~/.claude/.env` (file on disk = persistent attack surface)
2. Add their specific vars to the hardcoded whitelist (requires forking)

Neither is ideal. Env vars in memory are scoped to the process tree with no disk exposure.

### Prior art

Issues [#86](https://github.com/mksglu/claude-context-mode/issues/86), [#29](https://github.com/mksglu/claude-context-mode/issues/29), [#22](https://github.com/mksglu/claude-context-mode/issues/22) all added specific vars to the whitelist. This PR provides a general solution so users don't need to fork or file issues for each new var.

## Changes

- `src/executor.ts` — Added `#envPassthrough` field, constructor option parsing, env var fallback (`CONTEXT_MODE_ENV_PASSTHROUGH`), and glob matching in `#buildSafeEnv()`
- `src/server.ts` — Comment noting env var is read automatically
- `tests/executor.test.ts` — 8 new tests covering exact names, glob patterns, pass-all mode, env var config, and precedence

## Test plan

- [x] All 90 existing tests pass
- [x] 8 new tests for envPassthrough (exact match, glob, multiple patterns, pass-all, env var config, precedence)
- [x] TypeScript typecheck passes
- [x] Default behavior unchanged — custom vars are still blocked without explicit opt-in

### DIFF
--- a/README.md
+++ b/README.md
@@ -315,7 +315,7 @@ Each `ctx_execute` call spawns an isolated subprocess with its own process bound
 
 Eleven language runtimes are available: JavaScript, TypeScript, Python, Shell, Ruby, Go, Rust, PHP, Perl, R, and Elixir. Bun is auto-detected for 3-5x faster JS/TS execution.
 
-Authenticated CLIs work through credential passthrough — `gh`, `aws`, `gcloud`, `kubectl`, `docker` inherit environment variables and config paths without exposing them to the conversation.
+Authenticated CLIs work through credential passthrough — `gh`, `aws`, `gcloud`, `kubectl`, `docker` inherit environment variables and config paths without exposing them to the conversation. For custom environment variables, see [Environment Variable Passthrough](#environment-variable-passthrough).
 
 When output exceeds 5 KB and an `intent` is provided, Context Mode switches to intent-driven filtering: it indexes the full output into the knowledge base, searches for sections matching your intent, and returns only the relevant matches with a vocabulary of searchable terms for follow-up queries.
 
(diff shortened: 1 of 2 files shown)


flightlesstux:
Hey @dandaka — great PR! This is exactly the right approach to a recurring problem. The generalized solution is much better than adding one-off vars to the whitelist forever, and the test coverage is really solid.

A couple of things worth addressing before we merge:

## Bug: inconsistency between `true` mode and array mode

In the `true` branch, there's a `!(key in env)` guard:

```ts
if (key && value !== undefined && !(key in env)) {
  env[key] = value;
}
```

This means the built-in whitelist wins on conflicts. But in the array/glob branch, `env[pattern] = process.env[pattern]` overwrites without that guard — so the parent env wins there. The comment even says _"user-set vars override base env"_ which is the opposite of what the `true` branch actually does. One of these needs to be made consistent with the other.

## `envPassthrough: true` / `CONTEXT_MODE_ENV_PASSTHROUGH=*` needs a warning

The pass-all mode silently disables all env filtering, which could inadvertently expose AWS credentials, SSH keys, or other sensitive vars that happen to be in the process environment. This is easy to trigger accidentally in CI or via a `.env` file. I think this option is still worth keeping — power users have legitimate reasons for it — but it needs a clear warning in the JSDoc so people understand the trade-off.

## README

Users running as an MCP server (the most common case) have no way to discover `CONTEXT_MODE_ENV_PASSTHROUGH` without reading source. A small section in the README would go a long way.

---

Really appreciate the work here — the design is right and this will save a lot of people from filing issues. Once these are addressed it's ready to go. 🙏

dandaka:
## Follow-up fixes (59c461e)

After reviewing the initial implementation, added hardening and documentation:

### Bug fixes
- **Regex escaping**: `?` was not escaped in the wildcard-to-regex conversion — a pattern like `FOO?BAR` would be treated as regex `?` (zero-or-one) instead of a literal character
- **Whitelist override protection**: Added `!(key in env)` guard on both glob-match and exact-match branches, so user passthrough can never override built-in whitelist vars like `PATH`

### Documentation
- **README**: Added cross-link from sandbox section to the "Environment Variable Passthrough" docs; fixed "glob" → "wildcard" terminology since only `*` is supported
- **JSDoc**: Added security warning about `envPassthrough: true` exposing all parent env vars; clarified that only `*` wildcards are supported

### Terminology
- Replaced "glob patterns" with "wildcard patterns" throughout — calling it "glob" implies support for `?`, `**`, `{a,b}` which are not implemented

mksglu:
Closing in favor of a different approach that solves the root problem.

Instead of adding an opt-in `envPassthrough` parameter, we replaced the entire 80-var allowlist with a 27-var denylist (committed to `next`). Parent env is now inherited by default — only vars proven to corrupt sandbox stdout or enable code injection are stripped.

This means:
- User env vars (SLACK_TOKEN, LINEAR_API_KEY, custom CI vars) work automatically
- No configuration needed — just works
- Issues #22, #29, #86 are permanently resolved
- Security maintained via deny of BASH_ENV, NODE_OPTIONS, PERL5OPT, LD_PRELOAD, etc. (each backed by CVE/MITRE/live testing)

Thanks for raising this — your PR and the discussion around it directly led to the better solution.

PR #104, opened 2026-03-12
feat: add ctx_delegate for distributed parallel analysis

mksglu:
## Summary

- Adds `ctx_delegate` MCP tool that spawns parallel Claude sub-agents via `claude --print`
- Each sub-agent receives pre-read file contents, performs single-turn analysis, returns a compressed summary
- Results indexed per-agent into FTS5 (`source: "delegate:<task-name>"`) for targeted `ctx_search` follow-up
- Zero external dependencies — uses `claude --print` (installed with Claude Code)
- Adds Delegate Analytics section to `ctx_stats` (compression%, speedup, tokens saved)

## How it works

1. MCP server pre-reads files from disk and embeds them in the sub-agent prompt
2. Spawns `claude --print --model claude-sonnet-4-6 --max-turns 1` as child processes
3. CPU-based concurrency (max 10 parallel sub-agents) with `Promise.race` limiter
4. Each result indexed separately: `source: "delegate:core-modules"`, `source: "delegate:api-layer"`, etc.
5. Depth guard (`CTX_DELEGATE_DEPTH` env var) prevents infinite delegation loops

## Example prompts

These prompts naturally trigger `ctx_delegate` without mentioning the tool name:

```
Analyze all TypeScript modules in src/ and describe the architecture
```

```
Review the entire adapter system — what does each platform support?
```

```
Document every MCP tool handler in server.ts with input/output examples
```

```
Find all error handling patterns across the codebase and check for inconsistencies
```

```
Analyze the session continuity system — how do events flow from hooks to resume snapshots?
```

## Real benchmark

From a live test analyzing 42 files across 5 parallel sub-agents:

| Metric | Value |
|--------|-------|
| Sub-agents spawned | 5 |
| Wall time | 42.5s |
| Sequential would be | 200.0s |
| **Speedup** | **4.7x** |
| Raw tokens (files) | ~96,806 |
| Summary tokens | ~2,686 |
| **Compression** | **97.2%** |
| **Tokens saved** | **~94,120** |

## Files changed

| File | Change |
|------|--------|
| `src/delegate.ts` | New — core delegation module (spawn, concurrency, file reading) |
| `src/server.ts` | Tool registration, stats tracking, per-agent FTS5 indexing |
| `src/session/db.ts` | Session DB updates for delegate event tracking |
| `hooks/sessionstart.mjs` | Injects delegate into tool selection hierarchy (Claude Code only) |
| `configs/claude-code/CLAUDE.md` | Delegate as #1 in routing hierarchy |
| `CLAUDE.md` | Root-level routing hierarchy update |
| `skills/context-mode/SKILL.md` | Trigger phrases for natural routing |

## Test plan

- [ ] Run `npm run build` — TypeScript compiles
- [ ] Run `npm test` — existing tests pass
- [ ] Local dev mode test (see [CONTRIBUTING.md](./CONTRIBUTING.md) for setup)
- [ ] Test natural trigger: ask Claude to "analyze all modules in src/" — should pick `ctx_delegate`
- [ ] Test fail-fast: pass a non-existent file path — should return immediately with `missingPaths`
- [ ] Test depth guard: verify sub-agents cannot delegate further
- [ ] Run `/context-mode:ctx-stats` after a delegate call — verify Delegate Analytics section appears
- [ ] Test `ctx_search(queries: ["..."], source: "delegate:core-modules")` — verify per-agent results are searchable


mksglu:
Hey @skoll43 @rjkaes @flightlesstux @kurtextrem @xxnbyy @vaban-ru @Demir-Utku WDYT?

flightlesstux:
This looks really exciting! The idea of spawning parallel sub-agents to handle distributed analysis is super clever — and those benchmark numbers speak for themselves (4.7x speedup with 97.2% compression is wild 🎯).

A few things I love about the design:
- The depth guard via `CTX_DELEGATE_DEPTH` is a thoughtful touch to prevent runaway delegation loops
- Per-agent FTS5 indexing with `source: "delegate:<task-name>"` makes follow-up searches really targeted
- Zero external dependencies keeps things clean

Would love to see this land! Happy to test it out locally if you need more eyes on the benchmarks.

flightlesstux:
Grabbed a coffee and thought about this more deeply — here are some ideas that could take `ctx_delegate` even further:

**1. Result caching**
Hash each file's content + task description. If nothing changed since the last run, skip re-spawning and return cached results instantly. For large codebases this would make repeated analysis nearly free.

**2. Synthesis step**
After all sub-agents finish, run one final orchestrator pass that reads all delegate summaries and produces a unified output. Right now the caller has to mentally stitch results together — a `synthesize: true` option would close that loop automatically.

**3. Progressive / streaming indexing**
Index each sub-agent's result as it completes rather than waiting for the whole batch. Users could start querying `source: "delegate:api-layer"` while other agents are still running in parallel.

**4. Smart file chunking**
Auto-split large files (e.g. >500 lines) into overlapping chunks and assign one agent per chunk. A 2000-line file going to a single agent limits both quality and parallelism — chunking would help both.

**5. Per-agent token budget**
Expose a `maxTokensPerAgent` option so users can tune the cost vs. depth tradeoff explicitly depending on how thorough they need the analysis to be.

**6. Persistent delegate history across sessions**
Store delegate run metadata (files analyzed, tasks, compression stats) in the session DB so `ctx_stats` can show historical trends across sessions, not just the current one.

Would love to see some of these land as follow-ups if the core gets merged!

rjkaes:
@flightlesstux Ideas 1, 2, and 4 are excellent ideas. Idea 1 in particular feels like a huge win.

@mksglu Love this concept. ♥️

src/delegate.ts
```diff
@@ -0,0 +1,361 @@
...
+  model: string,
+  timeout: number,
+  limit: number,
+): Promise<DelegateTaskResult[]> {
+  const results: DelegateTaskResult[] = [];
+  const queue = [...tasks];
+  const running = new Set<Promise<void>>();
+
+  while (queue.length > 0 || running.size > 0) {
+    while (running.size < limit && queue.length > 0) {
```
rjkaes: **issue (blocking):** `concurrency` of 0 or negative causes an infinite loop.

The Zod schema at the tool boundary does not enforce `.positive().int()` on the `concurrency` field. If `concurrency: 0` reaches this line, `running.size < limit` is never true, so no tasks are ever dequeued, and the outer `while` spins forever.

Add `.positive().int()` to the Zod schema for `concurrency`, and/or add a `Math.max(1, ...)` guard here.

src/server.ts
```diff
@@ -1398,6 +1411,158 @@ server.registerTool(
...
+                "File/directory paths to pre-read and embed in the prompt. " +
+                "Directories are read recursively (code files only). " +
+                "Paths relative to project root. Example: ['src/server.ts', 'src/adapters/']",
+              ),
+          }),
+        )
+        .min(1)
+        .max(20)
+        .describe("Array of tasks — each becomes one parallel sub-agent"),
+      model: z
```
rjkaes: **issue (blocking, security):** The `model` string is passed directly to `claude --print --model <value>` with no validation.

While `spawn` with an args array avoids shell injection, there's no check that the model string is a legitimate identifier. A value containing spaces or starting with `--` could confuse the CLI argument parser.

Validate against a pattern (e.g., `/^[a-z0-9._-]+$/i`) or use an allowlist of known model prefixes.

src/delegate.ts
```diff
@@ -0,0 +1,361 @@
...
+  timeout?: number;
+  /** Max concurrent sub-agents. Default: CPU count, max: 10 */
+  concurrency?: number;
+}
+
+export interface DelegateTaskResult {
+  name: string;
+  summary: string;
+  durationMs: number;
+  inputTokens: number;
```
rjkaes: **suggestion (non-blocking):** `inputTokens`, `outputTokens`, and `cacheReadTokens` are hardcoded to 0 in every code path and never populated from `claude --print` output.

This is structurally misleading: consumers see `number` fields suggesting real token counts. Consider either:
1. Removing them until `claude --print` provides usage data, or
2. Adding a JSDoc note marking them as placeholders.

Similarly, `totalPromptTokens`/`totalSummaryTokens` in `DelegateResult` are chars/4 estimates. Renaming to `estimatedPromptTokens`/`estimatedSummaryTokens` would signal this.

mksglu:
Closing this PR. After re-evaluating with fresh eyes, the honest assessment:

1. **@rjkaes's 13 review items are all valid** — prompt injection via CLI args, unbounded memory, process leaks, dead code paths, concurrency edge cases. Every one of them is a legitimate blocker.
2. **No real user demand.** Zero issues requesting this. `batch_execute` + `execute_file` + `search` already cover multi-file analysis effectively.
3. **Hidden cost multiplication.** Each sub-agent is a full API call — 5 parallel delegates = 5x cost, invisible to users. Context savings don't justify cost opacity.
4. **Zero tests** for a feature that spawns processes, manages concurrency pools, and handles timeouts.

The concept has potential but the implementation isn't ready and the need isn't proven. If demand emerges, we'll revisit with @rjkaes's feedback addressed from the start.

Thanks @rjkaes @flightlesstux for the thorough review — it was right.

Issue #103, opened 2026-03-12
[Feature]: Don't spawn so many node processes that use 100% cpu on MacOS


scottlepp:
### Platform

Claude Code

### context-mode version

v1.0.18

### Problem

<img width="751" height="302" alt="Image" src="https://github.com/user-attachments/assets/e9c2fe24-e950-4197-8e1c-9083cc8b3701" />

lots of node processes being spawned using 100%

### Proposed solution

not spawn a bunch of node processes using 100% cpu

mksglu:
Thanks for reporting this @scottlepp.

From your Activity Monitor screenshot, I can see 9 node processes each at ~100% CPU with hours of accumulated CPU time. These are long-running MCP server processes — not short-lived hooks (which execute in <100ms and exit).

This is abnormal — the MCP server should be idle most of the time, waiting for stdio input from Claude Code. It should **not** be consuming 100% CPU.

**To diagnose, could you run these commands and paste the output?**

```bash
# 1. Check how many context-mode processes are running
ps aux | grep -i "context-mode" | grep -v grep

# 2. Kill all of them
pkill -f "context-mode.*start.mjs"

# 3. Run diagnostics (restart Claude Code first)
# Then in Claude Code, run:
/context-mode:ctx-doctor
```

**Most likely cause:** Multiple MCP server instances accumulating across session restarts without the old ones being killed. Claude Code should clean these up, but sometimes orphaned processes survive.

**Immediate fix:** Kill all context-mode processes with `pkill -f "context-mode.*start.mjs"` and restart Claude Code.

I'll investigate the root cause — likely we need a startup guard that detects and kills stale MCP server processes from previous sessions.

keki:
tl;dr: Yes they seem to be orphaned processes.

I'm seeing the same issue, I ran the commands, ctx-doctor says everything looks good. 
```⏺ context-mode doctor
                                                                                                                                                                                        
  - Runtimes: 7/11 (javascript, typescript, python, shell, ruby, php, perl)                                                                                                             
  - Performance: FAST (Bun)                                                                                                                                                           
  - Server test: PASS                                                                                                                                                                   
  - PreToolUse hook: PASS                                                                                                                                                             
  - SessionStart hook: PASS                                                                                                                                                             
  - Hook script: PASS                                                                                                                                                                 
  - Plugin enabled: PASS                                                                                                                                                                
  - FTS5 / better-sqlite3: PASS                                                                                                                                                         
[...]

PR #112, opened 2026-03-13
fix: kill process group on Unix to prevent orphaned children

dandaka:
## Summary
- Spawned processes now use `detached: true` on Unix, giving each its own process group
- `killTree()` sends SIGKILL to `-pid` (entire group) instead of just the single process
- `cleanupBackgrounded()` also kills process groups instead of individual PIDs
- Prevents `while(true) {}` loops (and their children) from surviving as orphaned processes consuming 100% CPU

## Root cause
On Unix, `proc.kill("SIGKILL")` only kills the immediate process. If the runtime (e.g. bun) forks internally, the child survives and gets reparented to PID 1. Two such orphans were found consuming 92% CPU each, running for 90+ minutes.

## Test plan
- [x] Added test: infinite loop process is dead after timeout kill
- [x] Added test: child processes spawned by parent are killed with the group
- [x] All 6 timeout tests pass
- [x] All 16 stream-cap tests pass

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -14,14 +14,17 @@ import type { ExecResult } from "./types.js";
 
 const isWin = process.platform === "win32";
 
-/** Kill process tree — on Windows, proc.kill() only kills the shell, not children. */
+/** Kill process tree — on Windows uses taskkill /T; on Unix kills the process group. */
 function killTree(proc: ReturnType<typeof spawn>): void {
   if (isWin && proc.pid) {
     try {
       execSync(`taskkill /F /T /PID ${proc.pid}`, { stdio: "pipe" });
     } catch { /* already dead */ }
-  } else {
-    proc.kill("SIGKILL");
+  } else if (proc.pid) {
+    try {
+      // Kill entire process group (negative PID) to prevent orphaned children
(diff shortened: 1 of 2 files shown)


mksglu:
Thanks for this fix @dandaka! 🙏

Your process group kill change has been merged and we've built on top of it — we just pushed an additional **process lifecycle guard** (`src/lifecycle.ts`) to main that detects parent death via ppid comparison, stdin close, and OS signals (SIGTERM/SIGINT/SIGHUP). Combined with your executor-level process group kill, both server-level and child-process orphans are now covered.

Changes in `939c5ca`:
- `src/lifecycle.ts` — new module: periodic ppid check + stdin close + signal handlers
- `src/server.ts` — lifecycle guard wired to `gracefulShutdown()`
- `src/db-base.ts` — WAL checkpoint before close to prevent SQLite lock contention

Would appreciate if you could test once the new version is live on the marketplace — specifically:
1. Start a session, note the MCP server PID
2. Force-kill the host (e.g. `kill -9 <claude-pid>`)
3. Verify the MCP server exits within ~30s (no more zombie processes)

Thanks again for flagging issue #103 and contributing the fix!

Issue #113, opened 2026-03-13
[Feature]: Throttle advisory PreToolUse guidance to once-per-session


RussellZager:
### Platform

Claude Code

### context-mode version

v1.0.18

### Problem

The PreToolUse hooks for Read, Bash, and Grep inject identical advisory guidance (`<context_guidance>` blocks) on **every** tool call. When an agent makes 3 parallel Read calls, 3 identical system reminders are injected into the conversation context. Over a session with dozens of tool calls, this:

- Wastes ~150 tokens per reminder x N tool calls per session
- Slows model generation as it processes redundant information
- Causes the model to "lose the thread" mid-task (stalling)
- Contributes to context window pressure, triggering earlier compaction

Related issues:
- #79 -- reported invasive hooks; the concern about repeated Read/Grep tips specifically
- #103 -- node process spawning / CPU cost (related symptom -- each hook spawns a process)
- #67 -- dedup for knowledge base outputs (analogous pattern, already fixed in v1.0.0)

### Proposed solution

Show each advisory guidance type (READ_GUIDANCE, BASH_GUIDANCE, GREP_GUIDANCE) **at most once per session**. The deny/modify actions (WebFetch block, curl/wget redirect, Agent prompt injection) should always fire -- only the `{ action: "context" }` returns need throttling.

Implementation approach (already working as a local patch):

```javascript
// In routing.mjs -- add after existing imports
import { existsSync, writeFileSync, mkdirSync } from "node:fs";
import { tmpdir } from "node:os";
import { resolve } from "node:path";

const _guidanceDir = resolve(tmpdir(), `context-mode-guidance-${process.ppid}`);
try { mkdirSync(_guidanceDir, { recursive: true }); } catch {}

function guidanceShown(type) {
  const marker = resolve(_guidanceDir, type);
  if (existsSync(marker)) return true;
  try { mkdirSync(_guidanceDir, { recursive: true }); } catch {}
  try { writeFileSync(marker, "", "utf-8"); } catch {}
  return false;
}
```

Then wrap each advisory return:
```javascript
// Before:
return { action: "context", additionalContext: READ_GUIDANCE };
// After:
return guidanceShown("read") ? null : { action: "context", additionalContext: READ_GUIDANCE };
```

**Why process.ppid**: Each hook invocation is a child process. ppid is the Claude Code process PID, constant for the session. New session = new PID = fresh markers. OS cleans $TMPDIR.

### Alternatives considered

[...]

PR #115, opened 2026-03-13
fix(tests): reset guidance throttle between core-routing tests

RussellZager:
## Summary

- The v1.0.21 guidance throttle (#113) causes 3 test failures in `core-routing.test.ts` because throttle state leaks between tests within the same vitest file
- Tests expecting `context` action for Bash only succeed on the first call per type — subsequent tests get `null` due to once-per-session throttle
- Adds `beforeEach(() => resetGuidanceThrottle())` to reset state between tests, matching the pattern already used in `guidance-throttle.test.ts`

## Failing tests (before fix)

```
FAIL  core-routing.test.ts > Bash tool > allows mkdir with BASH_GUIDANCE context
      expected null not to be null (line 89)

FAIL  core-routing.test.ts > Bash tool > allows npm install with BASH_GUIDANCE context
      expected null not to be null (line 95)

FAIL  core-routing.test.ts > Bash tool > does not false-positive on gradle in quoted text
      expected null not to be null (line 138)
```

## Test plan

- [x] All 28 core-routing tests pass after fix
- [x] Full test suite: no new failures introduced (6 pre-existing failures unrelated to this change)

### DIFF
--- a/tests/hooks/core-routing.test.ts
+++ b/tests/hooks/core-routing.test.ts
@@ -1,4 +1,4 @@
-import { describe, it, expect, beforeAll } from "vitest";
+import { describe, it, expect, beforeAll, beforeEach } from "vitest";
 
 // Dynamic import for .mjs module
 let routePreToolUse: (
@@ -12,6 +12,7 @@ let routePreToolUse: (
   additionalContext?: string;
 } | null;
 
+let resetGuidanceThrottle: () => void;
 let ROUTING_BLOCK: string;
 let READ_GUIDANCE: string;
 let GREP_GUIDANCE: string;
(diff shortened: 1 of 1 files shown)


mksglu:
Hey @RussellZager — thanks for catching this! You were absolutely right that the throttle fix broke these tests.

We couldn't merge this PR directly because it only covered part of the problem. Here's the full picture:

### What your PR fixed
`core-routing.test.ts` — the in-memory `Set` in `guidanceOnce()` persisted across tests within the same vitest file. Your `beforeEach` + `resetGuidanceThrottle()` fix was correct for this.

### What else was broken
The throttle uses a **hybrid approach** (in-memory Set + file-based O_EXCL markers keyed by `process.ppid`). This caused failures in two more places your PR didn't cover:

1. **`vscode-hooks.test.ts`** — hooks run as **subprocesses** via `spawnSync`. The subprocess's `process.ppid` = test runner's `process.pid`, so file markers persisted between tests. Needed a `beforeEach` that cleans `context-mode-guidance-${process.pid}` directory.

2. **Cross-file parallelism race** — vitest runs test files in parallel threads. All threads share the same `process.ppid`, so `core-routing.test.ts` (thread 1) and `guidance-throttle.test.ts` (thread 2) would race on the same marker directory. Thread 1 creates a `grep` marker → thread 2's Grep test hits `O_EXCL` and gets `null` instead of guidance.

### The fix we landed
Three commits on main:

- **`routing.mjs`**: Append `VITEST_WORKER_ID` to the marker directory path when running in vitest, so each parallel worker gets its own isolated marker directory. Zero impact on production (env var doesn't exist outside vitest).
  ```js
  const _guidanceId = process.env.VITEST_WORKER_ID
    ? `${process.ppid}-w${process.env.VITEST_WORKER_ID}`
    : String(process.ppid);
  const _guidanceDir = resolve(tmpdir(), `context-mode-guidance-${_guidanceId}`);
  ```

- **All 3 test files** (`core-routing`, `hook-integration`, `vscode-hooks`): Updated `beforeEach` cleanup to use the worker-aware directory path.

- **Bonus**: Fixed `executor.ts` Rust template (`#[allow(unused_variables)]`) and lifecycle test Windows skip — unrelated to your PR but were also failing on CI.

CI is now green on all 3 platforms (macOS, Ubuntu, Windows). Thanks again for the report — it led us to a deeper fix! 🤝

Closing this PR since the fix is already on main.

PR #116, opened 2026-03-13
feat: add verification script for guidance throttle

RussellZager:
## Summary

- Adds `scripts/verify-guidance-throttle.mjs` — a standalone script users can run after upgrading to verify the once-per-session guidance throttle (#113) is working correctly
- Exercises all throttle behaviors: first-show, suppression, deny/modify bypass, per-type independence, and reset
- Prints a color-coded pass/fail checklist with exit code 0/1

## Usage

```bash
node scripts/verify-guidance-throttle.mjs
```

## Example output

```
--- Throttle: first call shows guidance ---

  ✓  Read: first call returns context guidance
  ✓  Bash: first call returns context guidance
  ✓  Grep: first call returns context guidance

--- Throttle: subsequent calls are suppressed ---

  ✓  Read: second call returns null (throttled)
  ✓  Bash: second call returns null (throttled)
  ✓  Grep: second call returns null (throttled)

--- Deny/modify actions are NEVER throttled ---

  ✓  WebFetch: first call is denied
  ✓  WebFetch: second call is still denied

--- Per-type independence ---

  ✓  Bash guidance not affected by Read throttle
  ✓  Grep guidance not affected by Read throttle

--- Reset simulates new session ---

  ✓  Read: guidance returns after reset

──────────────────────────────────────────────────
✓ All 11 checks passed — guidance throttle is working correctly.
```

## Motivation

Users who maintain local patches or custom hook configurations want a quick way to verify the throttle works after upgrading. This script provides that without needing to run the full test suite.

Related: #113

### DIFF
--- a/scripts/verify-guidance-throttle.mjs
+++ b/scripts/verify-guidance-throttle.mjs
@@ -1,0 +1,90 @@
+#!/usr/bin/env node
+/**
+ * Verify that the guidance throttle is working correctly.
+ *
+ * Run after upgrading context-mode to confirm that advisory guidance
+ * (Read/Bash/Grep tips) fires once per session, not on every tool call.
+ *
+ * Usage:
+ *   node scripts/verify-guidance-throttle.mjs
+ *
+ * Expected output: all checks PASS.
+ */
+
+import { routePreToolUse, resetGuidanceThrottle } from "../hooks/core/routing.mjs";
+
+let passed = 0;
+let failed = 0;
+
+function check(label, actual, expected) {
+  const ok = actual === expected;
+  const icon = ok ? "\x1b[32m\u2713\x1b[0m" : "\x1b[31m\u2717\x1b[0m";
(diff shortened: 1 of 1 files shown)


mksglu:
Hey @RussellZager — thanks for putting this together! We appreciate the effort.

However, we already have comprehensive test coverage for the guidance throttle in our test suite:

- `tests/guidance-throttle.test.ts` — 8 tests covering per-type throttle, cross-process file markers, reset, and deny passthrough
- `tests/hooks/core-routing.test.ts` — 28 tests including guidance routing
- `tests/hooks/vscode-hooks.test.ts` — subprocess-level hook verification

These run on CI across macOS, Ubuntu, and Windows — all green as of v1.0.21. A standalone verification script would duplicate what the test suite already validates, so we don't need this at this time.

Thanks again for the contributions on #113 and #115 — they helped us find and fix the parallel test isolation issue!

PR #126, opened 2026-03-15
fix: add z.preprocess coercion for array params to handle double-serialization

### LINKED ISSUE #125: Add z.preprocess coercion for array params to handle double-serialization
## Problem

Claude Code sometimes passes MCP array parameters as **stringified JSON strings** instead of native arrays, causing `-32602` validation errors:

```
MCP error -32602: Input validation error: Invalid arguments for tool ctx_batch_execute: [
  { "code": "invalid_type", "expected": "array", "received": "string",
    "path": ["commands"], "message": "Expected array, received string" }
]
```

**What happens:** The model generates `commands: "[\"cat file1\", \"ls\"]"` (a string containing JSON) instead of `commands: ["cat file1", "ls"]` (a native array).

This affects `ctx_batch_execute` (both `commands` and `queries` params) and `ctx_search` (`queries` param).

### Root cause

Two contributing factors:
1. **Claude Code HTTP transport bug** ([anthropics/claude-code#34520](https://github.com/anthropics/claude-code/issues/34520)) — serialization layer double-encodes arrays for HTTP MCP servers
2. **Model behavior** — for stdio plugins (like context-mode), the LLM itself sometimes wraps array values in quotes. This is intermittent and hard to prevent with prompt engineering alone.

## Proposed fix

Add `z.preprocess()` coercion to all array input schemas. This is a **zero-risk defensive fix** — native arrays pass through unchanged, stringified arrays get auto-parsed.

### Helper functions

```typescript
function coerceJsonArray(val: unknown): unknown {
  if (typeof val === "string") {
    try {
      const parsed = JSON.parse(val);
      if (Array.isArray(parsed)) return parsed;
    } catch { /* not valid JSON, let zod handle the error */ }
  }
  return val;
}

function coerceCommandsArray(val: unknown): unknown {
  const arr = coerceJsonArray(val);
  if (Array.isArray(arr)) {
    return arr.map((item, i) =>
      typeof item === "string" ? { label: `cmd_${i + 1}`, command: item } : item
    );
  }
  return arr;
}
```

### Schema changes

```typescript
// ctx_search
queries: z.preprocess(coerceJsonArray, z.array(z.string()).optional()
  .describe("...")),

// ctx_batch_execute
commands: z.preprocess(coerceCommandsArray, z.array(z.object({...})).min(1)
  .describe("...")),
queries: z.preprocess(coerceJsonArray, z.array(z.string()).min(1)
  .describe("...")),
```

### Also consider: routing-block.mjs

Adding a `CRITICAL` note in the `<tool_selection_hierarchy>` section of `routing-block.mjs` helps reduce the frequency of model-side double-serialization:

```
- CRITICAL: commands and queries MUST be native arrays, NOT stringified JSON.
  CORRECT: commands: ["cmd1", "cmd2"], queries: ["q1", "q2"]
  WRONG:   commands: "[\"cmd1\", \"cmd2\"]" ← causes -32602 validation error
```

## Testing

We tested locally with real Zod `z.preprocess`:
- **10/10 unit tests** covering: normal arrays, stringified arrays, null, undefined, malformed JSON, non-array JSON strings, plain command strings → objects
- **6/6 integration tests** with actual Zod schemas: stringified queries, normal queries, undefined/optional, stringified string commands, stringified object commands, normal arrays

All pass. The fix is backward-compatible — no behavior change for correctly-typed inputs.

## Environment

- context-mode v1.0.18
- Claude Code on Linux (stdio MCP plugin)
- Errors observed with Claude Opus 4.6 and Sonnet 4.6

jeanmonet:
## Summary

- Adds defensive `z.preprocess()` coercion to `ctx_search` (`queries`) and `ctx_batch_execute` (`commands`, `queries`) that silently parses stringified JSON arrays before Zod validation
- Handles both **double-serialization** (arrays wrapped in quotes) and **schema simplification** (model sending plain command strings instead of `{label, command}` objects)
- Native arrays pass through unchanged — zero behavior change for correctly-typed inputs

## Problem

Claude Code sometimes passes MCP array parameters as stringified JSON strings instead of native arrays, causing `-32602` validation errors. Two contributing factors:

1. [Claude Code HTTP transport bug](https://github.com/anthropics/claude-code/issues/34520) — serialization layer double-encodes arrays for HTTP MCP servers
2. Model behavior — for stdio plugins (like context-mode), the LLM itself intermittently wraps array values in quotes

## Changes

Single file: `src/server.ts`

- `coerceJsonArray()` — if input is a string, tries `JSON.parse`; returns parsed array or original value
- `coerceCommandsArray()` — extends `coerceJsonArray` to also map plain command strings to `{label, command}` objects
- Wraps `ctx_search.queries` and `ctx_batch_execute.{commands, queries}` with `z.preprocess()`

## Test plan

- [x] 10/10 unit tests covering: normal arrays, stringified arrays, null, undefined, malformed JSON, non-array JSON strings, plain command strings → objects
- [x] 6/6 integration tests with real Zod `z.preprocess` and actual schemas
- [x] Applied and validated locally on v1.0.18 (Linux, stdio plugin mode)
- [ ] Upstream CI

Closes #125

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -847,6 +847,36 @@ const SEARCH_WINDOW_MS = 60_000;
 const SEARCH_MAX_RESULTS_AFTER = 3; // after 3 calls: 1 result per query
 const SEARCH_BLOCK_AFTER = 8; // after 8 calls: refuse, demand batching
 
+/**
+ * Defensive coercion: parse stringified JSON arrays.
+ * Works around Claude Code double-serialization bug where array params
+ * are sent as JSON strings (e.g. "[\"a\",\"b\"]" instead of ["a","b"]).
+ * See: https://github.com/anthropics/claude-code/issues/34520
+ */
+function coerceJsonArray(val: unknown): unknown {
+  if (typeof val === "string") {
+    try {
+      const parsed = JSON.parse(val);
+      if (Array.isArray(parsed)) return parsed;
+    } catch { /* not valid JSON, let zod handle the error */ }
+  }
+  return val;
+}
+
+/**
+ * Coerce commands array: handles double-serialization AND the case where
+ * the model passes plain command strings instead of {label, command} objects.
+ */
+function coerceCommandsArray(val: unknown): unknown {
+  const arr = coerceJsonArray(val);
+  if (Array.isArray(arr)) {
+    return arr.map((item, i) =>
+      typeof item === "string" ? { label: `cmd_${i + 1}`, command: item } : item
+    );
+  }
+  return arr;
+}
+
 server.registerTool(
   "ctx_search",
   {
@@ -855,10 +885,10 @@ server.registerTool(
       "Search indexed content. Pass ALL search questions as queries array in ONE call.\n\n" +
       "TIPS: 2-4 specific terms per query. Use 'source' to scope results.",
     inputSchema: z.object({
-      queries: z
+      queries: z.preprocess(coerceJsonArray, z
         .array(z.string())
         .optional()
-        .describe("Array of search queries. Batch ALL questions in one call."),
+        .describe("Array of search queries. Batch ALL questions in one call.")),
       limit: z
         .number()
         .optional()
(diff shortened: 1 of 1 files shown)


mksglu:
Thanks for this clean contribution! 🎯

Solid defensive fix for the Claude Code double-serialization issue. The `z.preprocess` approach is the right pattern — transparent when input is already correct, graceful coercion when it's not.

Merging to `next` for integration testing.

PR #128, opened 2026-03-15
feat: worktree session isolation via CONTEXT_MODE_SESSION_SUFFIX

Setteld:
## Problem

When running two Claude Code instances simultaneously in git worktrees that share
the same git root, context-mode can resolve to the same `projectDir` → same
session DB hash → cross-contaminated session events at `SessionStart`.

## Solution

- New `src/git.ts` utility: `getWorktreeSuffix()` detects if CWD is a non-main
  git worktree and returns `__${basename(cwd)}` as a suffix.
- `CONTEXT_MODE_SESSION_SUFFIX` env var allows explicit override for CI/no-git
  environments (zero-config for single-worktree users).
- Session DB path gains the suffix: `{hash}__worktree-name.db` → fully isolated.

## Impact

- Zero impact on existing single-worktree setups (suffix is empty string).
- `execSync` has 2s timeout + `stdio: ignore` → no performance impact in normal use.
- Allows parallel Claude Code sessions across multiple worktrees without session bleed.

## Testing

Added `tests/git.test.ts` covering: main worktree (no suffix), empty env var,
and explicit env var override.

### DIFF
--- a/README.md
+++ b/README.md
@@ -295,6 +295,40 @@ Global `~/.codex/AGENTS.md` applies to all projects. Project-level `./AGENTS.md`
 
 </details>
 
+<details>
+<summary><strong>Build Prerequisites</strong> <sup>(CentOS, RHEL, Alpine)</sup></summary>
+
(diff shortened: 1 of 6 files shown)


mksglu:
@Setteld — all 6 review items addressed cleanly in a single commit. Nice work.

Merging into `next`. We'll do a small follow-up to move `getWorktreeSuffix()` from the standalone `git.ts` into `src/session/db.ts` where the session path logic lives — avoids a single-function file.

Thanks for the contribution!

PR #121, opened 2026-03-14
feat: add Antigravity support via self-instrumentation

winsorllc:
# feat: add Antigravity support via self-instrumentation

## Type of Change
- [x] New feature (non-breaking change which adds functionality)
- [x] Documentation update

## Description
This PR introduces support for **Antigravity** (Google DeepMind's agentic IDE) to the `context-mode` ecosystem.

### The Challenge
Antigravity is a standalone Electron-based environment that currently lacks a public plugin API or MCP hook infrastructure. Traditional automatic hooks (PostToolUse, PreCompact, etc.) cannot be injected into the daemon process.

### The Solution: Self-Instrumentation
We've implemented a **"Manual Hook"** architecture where the AI agent proactively instruments itself. The agent calls a set of Python-based utility scripts via its native `run_command` tool to:
1.  **Summarize context** before tool usage (e.g., smart file reading, directory listing).
2.  **Log structured events** to a persistent SQLite SessionDB (WAL mode).
3.  **Perform compaction recovery** by detecting daemon-level message count drops and restoring state from snapshots.

### New Components (in `contrib/antigravity`)
- `ctx_read.py` / `ctx_dir.py` / `ctx_summary.py`: Context-efficient pre-processors.
- `session_db.py` / `event_extract.py`: SQLite event store and tool-to-category mapping.
- `snapshot_builder.py` / `ctx_session.py`: Continuity management and snapshot generation.
- `ctx_instrument.py`: The unified entry point for "manual hooks".
- `ctx_doctor.py`: Comprehensive healthy-check utility.
- `SKILL.md`: Documentation for the agent on how to use these tools.

## Testing
- Tested in Antigravity version 1.19.4.
- Verified 11/11 health checks pass.
- Measured 50-90% context savings for large file reads and command outputs.
- Verified session continuity after conversation compaction.

## Checklist
- [x] I have scrubbed all local absolute paths and sensitive data.
- [x] I have followed the architectural patterns of `context-mode`.
- [x] Documentation has been updated.
- [x] All Python scripts are compatible with Python 3.10+.

Issue #120, opened 2026-03-14
[Feature]: Antigravity Workflows & Kiro Hooks Support


thedevbob005:
### Platform

All platforms

### context-mode version

1.0.22

### Problem

No support for Google Antigravity and Kiro yet.

### Proposed solution

Google Antigravity does not have hooks but has Workflows that should be able to do similarly.
Kiro has hooks support but a proper guide to create them needs to be added to readme.

winsorllc:
Hi! I just submitted PR #121 which addresses exactly this. It includes a full integration for Google Antigravity using the 'Self-Instrumentation' pattern via Workflows and Python utility scripts. Check out the \contrib/antigravity\ folder in that PR for the implementation.

PR #127, opened 2026-03-15
feat: add Antigravity adapter (mcp-only paradigm)

juxtapo9090:
## Summary

Adds first-class [Antigravity](https://blog.google/technology/google-deepmind/) (Google DeepMind agentic IDE) support following the established TypeScript adapter pattern.

## Paradigm: `mcp-only`

Antigravity is a standalone Electron-based IDE that lacks a public plugin API or MCP hook infrastructure. This adapter follows the **Codex CLI pattern** — all hook capabilities are `false`, and routing instructions (`ANTIGRAVITY.md`) provide soft enforcement (~60% compliance).

### New files (4)
- `src/adapters/antigravity/index.ts` — `AntigravityAdapter` class implementing `HookAdapter`
- `src/adapters/antigravity/hooks.ts` — Stub module (empty `HOOK_TYPES`)
- `src/adapters/antigravity/config.ts` — Thin re-exports
- `configs/antigravity/ANTIGRAVITY.md` — Routing instructions tailored for Antigravity tools

### Modified files (2)
- `src/adapters/detect.ts` — Add `ANTIGRAVITY_SESSION_ID`/`ANTIGRAVITY_PROJECT_DIR` detection (placed before Gemini CLI to avoid `~/.gemini/` collision)
- `src/adapters/types.ts` — Add `"antigravity"` to `PlatformId` union

### Test files (2)
- `tests/adapters/antigravity.test.ts` — 22 tests covering identity, capabilities, parse/format, config paths, diagnostics
- `tests/adapters/detect.test.ts` — +4 tests for Antigravity detection and `getAdapter()`

## Detection

| Signal | Confidence |
|--------|------------|
| `ANTIGRAVITY_SESSION_ID` env var | High |
| `ANTIGRAVITY_PROJECT_DIR` env var | High |
| `~/.gemini/antigravity/` directory | Medium |

## Config

- **Settings:** `~/.gemini/antigravity/settings.json` (JSON format)
- **Sessions:** `~/.gemini/antigravity/context-mode/sessions/`
- **Routing:** `ANTIGRAVITY.md` at project root

## Verification

- ✅ `tsc --noEmit` — zero errors
- ✅ 169/169 adapter tests pass (8 files)
- ✅ Full suite 887/890 (3 pre-existing failures in `session-hooks-smoke.test.ts`)
- ✅ Tested live — context-mode MCP server running on Antigravity with `ctx_execute`, `ctx_search`, `ctx_stats` confirmed working

PR #129, opened 2026-03-15
feat: add an Antigravity adapter

winsorllc:
This PR reworks Antigravity support into the TypeScript adapter shape requested in the review on #121 and the feature request in #120.

## What changed

- adds a dedicated TypeScript adapter under `src/adapters/antigravity/`
- wires Antigravity into platform detection and adapter selection
- uses Antigravity MCP config at `~/.gemini/antigravity/mcp_config.json`
- adds soft-enforcement assets under `configs/antigravity/`:
  - `GEMINI.md`
  - `.agent/workflows/context-mode.md` template
  - `mcp_config.json` example
- updates README install/docs and compatibility tables for Antigravity
- adds targeted tests for adapter behavior, detection, and hookless routing-instruction flow
- fixes the new adapter tests to be CI-safe under Vitest workers on macOS/Linux

## Why this shape

- stays in TypeScript/Node instead of introducing a parallel Python runtime
- follows the existing `src/adapters/<name>/` pattern
- uses context-mode's existing MCP/routing path instead of `run_command` wrappers
- avoids fabricated Antigravity env vars and instead uses verified filesystem/config signals

## Current scope and limitation

- Antigravity is implemented as `mcp-only`
- routing is soft only via MCP + `.agent/workflows/context-mode.md` + project `GEMINI.md`
- no native Antigravity pre/post tool hooks are assumed here, so session continuity is not enabled yet

## Verification

- `npm run typecheck`
- `npx vitest run tests/adapters/detect.test.ts tests/adapters/antigravity.test.ts tests/routing-instructions.test.ts`

## Remaining gap

Live integration proof from an actual Antigravity session is still needed before supported-platform claims should be treated as fully verified. Unit tests here cover adapter behavior and documented config paths, not end-to-end runtime confirmation inside Antigravity.

This should address issue #120 without reviving the Python `contrib/` approach from #121.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -254,6 +254,30 @@ After rebuilding, restart your Claude Code session. The MCP server reloads on se
 
 We follow test-driven development. Every PR must include tests.
 
+## Antigravity Adapter Notes
+
+If your PR touches `src/adapters/antigravity/`, keep these constraints explicit:
+
+- Treat Antigravity as `mcp-only` until there is real evidence of hook support.
+- Use `~/.gemini/antigravity/mcp_config.json` for MCP registration.
+- Use project `GEMINI.md` plus `.agent/workflows/context-mode.md` for soft routing guidance.
+- Do not introduce Antigravity-specific env vars, tool names, or config paths unless you can verify them from a real Antigravity session or primary documentation.
+- Prefer filesystem- or config-based detection over guessed environment variables.
+
+Before opening or updating an Antigravity PR, run:
+
+```bash
+npm run typecheck
+npx vitest run tests/adapters/detect.test.ts tests/adapters/antigravity.test.ts tests/routing-instructions.test.ts
+```
+
+If you have live Antigravity access, include integration proof in the PR:
+- screenshot, log, or GIF showing context-mode MCP tools visible in Antigravity
+- one successful tool invocation such as `ctx_execute` or `ctx_search`
+- the actual `mcp_config.json` entry used for registration
+
+Do not claim full Antigravity support from unit tests alone.
+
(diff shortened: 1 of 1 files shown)


winsorllc (commented):
please review

mksglu:
@winsorllc — closing this PR. Your approach was significantly more honest than #127 (correct `mcp_config.json` path, no fabricated env vars, proper `GEMINI.md` filename), and I want to acknowledge that.

However, there are still issues I can't merge around:

1. **`ANTIGRAVITY_HOME`** — `getAntigravityHome()` uses this env var, but there's zero evidence it exists. Not in Antigravity's system prompt leak (21 tools, no env var references), not in official docs, not in community guides. It's the same class of problem as #127's fabricated env vars — just one instead of three.

2. **Detection logic** — filesystem-based detection (`~/.gemini/antigravity/` exists) works but is fragile. If a user installs Antigravity once then switches to Gemini CLI, the directory persists and detection is wrong.

3. **No integration proof** — neither PR demonstrated context-mode actually running inside Antigravity.

I've implemented Antigravity support using a fundamentally different approach: **MCP protocol-level detection** via `clientInfo.name` from the initialize handshake. The MCP spec requires clients to send `{ name: "antigravity-client" }` during init — zero config, zero env var guessing, verified against the [Apify MCP Client Capabilities Registry](https://github.com/apify/mcp-client-capabilities). The adapter + client-map + detection changes are already passing 917 tests.

Your README install section and CONTRIBUTING.md guardrails were well-written — I'll incorporate similar documentation when the release ships.

When the next version is out, I'd really appreciate your help **testing it in a live Antigravity session**. That's the missing piece none of us can cover with unit tests. I'll tag you on the release.

Thanks for the careful work.

PR #133, opened 2026-03-16
refactor: modularize server, fix security issues, add tests

kianwoon:
## Summary

Closes #18 (fork PR reference)

- **Server modularization**: Split monolithic `server.ts` (~1700 lines) into 13 focused modules under `src/tools/` (11 tool files) and `src/server/` (4 server helpers). Zero behavioral changes.
- **Security fixes**: Non-greedy regex for nested parenthesis matching in `parseBashPattern`/`parseToolPattern`, silent `catch {}` blocks now log to stderr, background process cap (10 max, evicts oldest zombie)
- **Type safety**: Replaced `any` with `Record<string, unknown>` in security policy parsing
- **New tests**: 8 edge-case tests covering nested parens, chained commands, globstar patterns

## New files

| File | Purpose |
|------|---------|
| `src/server/session-stats.ts` | Session event tracking |
| `src/server/intent-search.ts` | Intent-driven search logic |
| `src/server/security-wrapper.ts` | Security policy enforcement wrapper |
| `src/server/snippet-extractor.ts` | Markdown snippet extraction |
| `src/tools/*.ts` | 11 tool modules extracted from server.ts |
| `src/plugin-root.ts` | Plugin root discovery (bundled + dev mode) |
| `tests/security.test.ts` | 8 new edge-case tests |

## Modified files

| File | Changes |
|------|---------|
| `src/server.ts` | ~1700 → ~90 lines (orchestrator only) |
| `src/executor.ts` | Background PID cap, CWD fix |
| `src/security.ts` | Non-greedy regex, `any` → typed |

## Test plan

- [x] `npx vitest run tests/` — all existing + 8 new tests pass
- [x] No behavioral changes — pure refactor + targeted fixes
- [x] Please run full test suite on CI across platforms

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -49,9 +49,6 @@ export class PolyglotExecutor {
   /** PIDs of backgrounded processes — killed on cleanup to prevent zombies. */
   #backgroundedPids = new Set<number>();
 
-  /** Maximum number of backgrounded processes before oldest is evicted. */
-  static #MAX_BACKGROUND_PIDS = 10;
-
   constructor(opts?: {
     maxOutputBytes?: number;
     hardCapBytes?: number;
@@ -236,18 +233,7 @@ export class PolyglotExecutor {
         if (background) {
           // Background mode: detach process, return partial output, keep running
           resolved = true;
-          if (proc.pid) {
-            // Evict oldest backgrounded process if limit reached
-            if (this.#backgroundedPids.size >= PolyglotExecutor.#MAX_BACKGROUND_PIDS) {
-              const oldest = this.#backgroundedPids.values().next().value;
-              if (oldest !== undefined) {
-                try { process.kill(-oldest, "SIGTERM"); } catch { /* already dead */ }
-                try { process.kill(oldest, "SIGTERM"); } catch { /* already dead */ }
-                this.#backgroundedPids.delete(oldest);
-              }
-            }
-            this.#backgroundedPids.add(proc.pid);
-          }
+          if (proc.pid) this.#backgroundedPids.add(proc.pid);
           proc.unref();
           proc.stdout!.destroy();
           proc.stderr!.destroy();
--- a/src/tools/batch-execute.ts
+++ b/src/tools/batch-execute.ts
@@ -40,27 +40,26 @@ export function registerBatchExecuteTool(server: McpServer, deps: ToolDeps): voi
           .preprocess(
             (v) => coerceCommandsArray(v),
             z.array(
-            z.object({
-              label: z
-                .string()
-                .describe(
-                  "Section header for this command's output (e.g., 'README', 'Package.json', 'Source Tree')",
-                ),
-              command: z
-                .string()
-                .describe("Shell command to execute"),
-            }),
+              z.object({
+                label: z
+                  .string()
+                  .describe(
+                    "Section header for this command's output (e.g., 'README', 'Package.json', 'Source Tree')",
+                  ),
+                command: z
+                  .string()
+                  .describe("Shell command to execute"),
+              }),
+            ).min(1),
           )
-          .min(1)
           .describe(
             "Commands to execute as a batch. Each runs sequentially, output is labeled with the section header.",
           ),
         queries: z
           .preprocess(
             (v) => coerceJsonArray(v),
-            z.array(z.string())
+            z.array(z.string()).min(1)
           )
-          .min(1)
           .describe(
             "Search queries to extract information from indexed output. Use 5-8 comprehensive queries. " +
             "Each returns top 5 matching sections with full content. " +


mksglu:
## Background PID Cap — Questions Before Merge

@kianwoon the modularization and stderr logging changes look good. One thing I want to understand better before merging: the background PID eviction cap (10 max).

### Context

The current codebase already handles background process cleanup via `cleanupBackgrounded()` on exit/SIGINT/SIGTERM. MCP server sessions are short-lived (per-conversation). The `background: true` option is niche — typical usage is 1-2 processes, and the tool description already discourages heavy use.

### Questions

1. **Can you reproduce a scenario where 10+ background processes accumulate in a single session?** What was the actual use case that motivated this cap? Did you hit this in practice or is it theoretical?

2. **SIGTERM reliability**: The eviction sends SIGTERM, but the evicted PID is immediately removed from `#backgroundedPids`. If the process ignores SIGTERM, it becomes an untracked zombie that `cleanupBackgrounded()` can't reach on shutdown. The existing `killTree()` uses SIGKILL for this reason. Was SIGTERM a deliberate choice here?

3. **Silent eviction**: When the oldest process is killed, the caller (Claude) gets no feedback. If a user's dev server suddenly dies because a new background process evicted it, there's no way to know why. Should the execute response include an eviction notice?

4. **Testing**: There are currently zero tests for background process management. If this cap is solving a real problem, we should be able to write a test that reproduces the accumulation scenario. If it's purely theoretical defense-in-depth, the test would just be exercising synthetic conditions.

Not blocking on these — just want to understand the motivation so we can decide whether to merge as-is or tighten it up. Thanks!

kianwoon:
Thanks for the thorough review @mksglu — addressing each question below.

### 1. Can you reproduce 10+ accumulation?

To be transparent: the motivation was theoretical defense-in-depth rather than a reproduced scenario. I haven't hit 10+ background processes in practice. The realistic accumulation path is a single heavy session where the agent spins up multiple background servers/watchers — but whether agent teams compound this depends on how Claude Code shares MCP servers across teammates, which is an undocumented implementation detail I can't verify. So I should be upfront: this is a safety net, not a fix for a known bug.

### 2. SIGTERM reliability

Valid concern. Two issues: (1) eviction removes the PID from tracking immediately, so if SIGTERM is ignored, the orphan is invisible to `cleanupBackgrounded()` for the rest of the session. (2) Worth noting `cleanupBackgrounded()` itself also uses SIGTERM — but that's less risky since it runs at shutdown (orphans get reparented to init). The eviction mid-session is worse. Fix: use the `killTree()` SIGKILL pattern for eviction instead.

### 3. Silent eviction

Agreed — silent eviction is bad UX. The execute response should include a warning like `[evicted background process PID ${oldest} to stay within limit]` so the caller (and user) knows what happened. This is a one-line addition to the execute response.

### 4. Testing

Fair point. If we keep the cap, I'll add a test that:
- Spawns 11+ background processes
- Verifies the oldest was killed and removed from tracking
- Verifies the newest is still tracked

If we decide to remove the cap instead, no test needed.

mksglu:
## Architect Review — Follow-up on @kianwoon's Response

Reviewing the response at #issuecomment-4072081190 against the fix commit `4a95208`. Three gaps between what was discussed and what was actually shipped:

---

### 1. SIGTERM vs SIGKILL Contradiction

@kianwoon's response says:

> The eviction mid-session is worse. Fix: use the `killTree()` SIGKILL pattern for eviction instead.

But commit `4a95208` does the **opposite** — changes to `SIGTERM`. The review approval confirmed SIGTERM (matching `cleanupBackgrounded()`), but the contributor's own analysis argues mid-session eviction is fundamentally different from shutdown cleanup:

- **Shutdown** (SIGTERM OK): process tree dies anyway, orphans reparent to init
- **Mid-session eviction** (SIGTERM risky): if ignored, PID is already deleted from `#backgroundedPids` — becomes an invisible orphan for the rest of the session

**Recommendation**: Either:
- (a) Use `killTree()` with SIGKILL for eviction (as kianwoon's analysis suggests), or
- (b) Keep SIGTERM but **don't delete from `#backgroundedPids` until confirmed dead** (check with `process.kill(pid, 0)` after a short delay)

Current code does neither — it's the worst of both: sends SIGTERM (may be ignored) AND immediately removes tracking.

---

### 2. Silent Eviction — Acknowledged but Not Implemented

kianwoon agreed:

> Silent eviction is bad UX. The execute response should include a warning like `[evicted background process PID ${oldest} to stay within limit]`

This one-liner was not included in `4a95208`. A user's dev server can silently die with zero diagnostic signal. This should be a prerequisite for merge.

---

### 3. Background PID Cap — Keep or Remove?

kianwoon was transparent: this is **theoretical defense-in-depth**, not fixing a real bug. The existing architecture already handles cleanup:
- `cleanupBackgrounded()` on exit/SIGINT/SIGTERM
- MCP sessions are short-lived (per-conversation)
- `background: true` is niche (1-2 processes typical)

**My recommendation**: Remove the PID cap entirely from this PR. Ship the modularization (which is clean) and the security fixes separately. The cap introduces complexity (eviction logic, signal handling, tracking consistency) for a scenario that can't be reproduced. If we want it later, it deserves its own PR with:
- Eviction notice in response
- Proper signal handling (SIGTERM → wait → SIGKILL)
- Tests for the accumulation scenario
- Configurable limit (not hardcoded 10)

---

### Summary

| Item | Status | Action |
|------|--------|--------|
| CWD revert | Fixed | — |
| z.preprocess restore | Fixed | — |
| SIGTERM/tracking inconsistency | Open | Fix or remove cap |
| Silent eviction warning | Acknowledged, not shipped | Must ship before merge |
| Background PID tests | Not added | Required if cap stays |

The modularization itself is solid work. Recommend separating the PID cap into its own PR to unblock the refactor merge.

kianwoon:
Done — pushed `a470769` with two changes:

1. **Stripped the background PID cap entirely** — removed `#MAX_BACKGROUND_PIDS`, eviction logic, and SIGTERM kill. Back to the simple `if (proc.pid) this.#backgroundedPids.add(proc.pid)` with existing `cleanupBackgrounded()` handling shutdown. As recommended, the cap deserves its own PR with proper signal handling (SIGTERM → wait → SIGKILL), eviction notice in response, tests, and a configurable limit.

2. **Fixed a pre-existing zod schema bug** in `batch-execute.ts` — `.min(1)` was chained on `z.preprocess()` (returns `ZodType`) instead of the inner `z.array()`. This was causing `tests/core/search.test.ts` to fail on import. Applied the same fix to the `queries` field which had the same pattern.

All 36 test suites pass (1086 tests).

The modularization + security fixes are now clean for review.

mksglu:
Hey @kianwoon — this PR has merge conflicts with the `next` branch. Could you rebase onto the latest `next` and resolve the conflicts?

```bash
git fetch origin
git rebase origin/next
# resolve conflicts
git push --force-with-lease
```

Once the conflicts are resolved, we'll review and move forward. Thanks!

PR #178, opened 2026-03-24
feat: configurable fetch backend with browser fallback

### LINKED ISSUE #165: [Feature]: Make fetch_and_index backend configurable (browser fallback for JS-rendered pages)
## Problem

`fetch_and_index` uses Node's `fetch()` (plain HTTP client) which fails on:
- SPAs and JS-rendered pages (Gatsby, Next.js, etc.) — returns empty or minimal HTML
- Sites that block non-browser user agents — returns 403/404
- Pages behind Cloudflare/anti-bot protection

This results in frequent `HTTP 404` or empty content errors that break research workflows.

## Current workaround

I'm using a user-level `PreToolUse` hook in `~/.claude/settings.json` that denies `fetch_and_index` and redirects to:
- **`gh` CLI** for GitHub URLs (structured JSON, handles auth)
- **`agent-browser`** (Vercel's headless Chromium CLI) for everything else

And I edited `routing-block.mjs` to update the guidance, but that gets overwritten on upgrade.

## Proposed solution

Make the fetch backend configurable, ideally via a setting in `~/.claude/settings.json` or a config file. Options could include:

1. **`fetch`** (current default) — plain HTTP, fast but limited
2. **`agent-browser`** — headless Chromium via CLI, handles JS rendering
3. **`gh`** — automatic for GitHub URLs, uses `gh` CLI for structured JSON
4. **Custom command** — user-specified shell command for fetching

Alternatively, a simpler approach: add a browser fallback in `fetch_and_index` itself — if `fetch()` fails (non-200 or empty content), try `agent-browser` before returning an error.

## Environment

- context-mode: 0.9.22 (upgrading to latest)
- agent-browser: 0.21.4
- Platform: macOS (Darwin)
- Claude Code

## Additional context

The `routing-block.mjs` guidance currently tells Claude to "use fetch_and_index instead of WebFetch", but there's no way for users to override this guidance without editing cache files (which get overwritten on upgrade). A user-configurable override for this guidance would also help.

waynehoover:
## Summary

Makes `fetch_and_index` backend configurable via environment variables, with optional browser fallback for JS-rendered pages and sites behind anti-bot protection.

Closes #165

### Environment variables

- **`CTX_FETCH_BACKEND`**: `fetch` (default) | `browser` | `agent-browser` | custom shell command (e.g. `curl -sL $CTX_URL`)
- **`CTX_FETCH_FALLBACK`**: `browser` | `agent-browser` | `auto` — retries with headless browser when primary fetch fails/returns empty

### Changes

- `buildBrowserFetchCode()` — uses `agent-browser` CLI via `execFileSync` (no shell, prevents URL injection)
- `buildCustomFetchCode()` — passes URL via `CTX_URL` env var (never interpolated into shell command)
- `executeFetchCode()` — extracted helper for DRY primary/fallback retry logic
- `buildFetchCodeForBackend()` — dispatches to the right code builder based on config
- `getFetchBackend()` / `isFallbackEnabled()` / `detectBrowserCli()` — config + detection utilities
- Lazy-cached `agent-browser` detection (no startup penalty when not installed)
- Tracks effective backend (reports "via browser" when fallback activated)

### Security

- Browser backend: `execFileSync('agent-browser', ['open', url])` — no shell
- Custom backend: URL passed as `CTX_URL` env var, `{{url}}` in template becomes `$CTX_URL`
- Detection: `execFileSync("agent-browser", ["--version"])` — no shell

## Test plan

- [x] `getFetchBackend()` returns correct backend for all env var values
- [x] `isFallbackEnabled()` respects env var settings
- [x] `detectBrowserCli()` returns boolean without throwing
- [x] Source invariants: `execFileSync` used (not `execSync`), URL passed via env var
- [x] Full test suite passes (1259 tests, 0 failures)
- [x] TypeScript type-check passes
- [x] Build succeeds


mksglu:
## Architecture Review — 6 Specialist Agents | Verdict: REQUEST CHANGES

Thanks @waynehoover — the feature design is solid and the security model for URL injection is well done. However, 6 specialist reviews (Security Engineer, OS Architect, LLM Architect, Context Mode Architect, Performance Engineer, Staff Engineer) found blocking issues that need fixes before merge.

### What Passed Review

| Concern | Status |
|---------|--------|
| URL injection (browser) | SAFE — `execFileSync` with array args, no shell |
| URL injection (custom) | SAFE — `CTX_URL` env var, never interpolated |
| Pipeline integrity | PRESERVED — all backends produce identical indexing output |
| Schema compatibility | UNCHANGED — `url`, `source`, `force` params |
| Cache design | CORRECT — backend not in cache key |
| Happy path perf | ZERO overhead when env vars unset |
| Architecture | CORRECT — env var for infra config is right for v1 |

### Blocking Issues (must fix)

**1. CI not running**
No checks reported on the branch. Tests must pass on all 3 platforms before merge.

**2. Remove bundle files from PR**
`server.bundle.mjs` and `cli.bundle.mjs` should not be in the PR — CI auto-builds on merge. Creates 328 lines of noise + merge conflict risk.

**3. Windows `.cmd` shim resolution** (OS Architect)
`execFileSync("agent-browser", ...)` fails on Windows because `.cmd` shim extensions aren't resolved without `shell: true`. Need Windows-aware binary resolution (e.g., check for `agent-browser.cmd` on win32).

**4. Orphaned Chromium processes** (OS Architect)
`agent-browser open` spawns headless Chromium. If the executor kills the subprocess on timeout, the Chromium process may orphan. Add `agent-browser close` in a `finally` block.

**5. No SIGTERM/SIGINT cleanup** (OS Architect)
When the MCP server exits, active browser fetches leave Chromium processes orphaned. Consider a cleanup handler.

**6. Browser timeout mismatch** (Staff + Performance)
When `CTX_FETCH_BACKEND=browser`, outer executor timeout is 30s but inner `execFileSync` calls sum to 45s → premature kill. Align to 60s like the fallback path. Also: worst-case 90s (30s primary + 60s fallback) is too long — consider 30s+30s=60s max.

**7. Unquoted `$CTX_URL` in custom backend** (Security)
`{{url}}` → `$CTX_URL` without quotes causes word splitting on URLs with spaces:
```diff
- const cmd = template.replace("{{url}}", "$CTX_URL");
+ const cmd = template.replace("{{url}}", '"$CTX_URL"');
```

### Medium Issues (strongly recommended)

**8. Silent browser error discard** (LLM Architect)
When both fetch and browser fallback fail, the browser error is swallowed. The LLM only sees the original fetch error. Log or append the browser error.

**9. No byte cap on browser content** (Performance)
Browser-rendered SPAs can produce 3-10x more HTML than static fetch. Add a size guard before indexing (e.g., truncate to 512KB).

**10. `agent-browser get text body` semantics** (LLM Architect)
Clarify: does this return `innerText` (plain text) or `innerHTML` (markup)? If plain text, the Turndown pass in `buildBrowserFetchCode` is wasteful.

### Positive Notes

The overall architecture is correct — backend selection is cleanly separated from the indexing pipeline, security model uses `execFileSync` (no shell) consistently, and the lazy-cached detection avoids startup penalty. The source-invariant tests are a creative regression guard. Looking forward to the next revision.

waynehoover:
## Architecture Review — All Issues Addressed

Pushed `d0b8508` addressing all blocking and medium issues from the 6-specialist review.
Follow-up `720564e` fixes CI — updated security tests to match Windows shim and quoting changes. All 3 platforms now pass.

### Blocking Issues — Fixed

| # | Issue | Fix |
|---|-------|-----|
| 1 | CI not running | ✅ CI passes on all 3 platforms (`720564e` updated security tests to match new shim + quoting logic). |
| 2 | Bundle files in PR | Reverted `cli.bundle.mjs` and `server.bundle.mjs` to main — net PR diff is now 0 lines for bundles. |
| 3 | Windows .cmd shim | Added `agentBrowserBin()` helper that probes `agent-browser.cmd`/`.exe` on win32, caches result. All call sites use it. |
| 4 | Orphaned Chromium | Added `finally { closeBrowser() }` in `buildBrowserFetchCode` — always calls `agent-browser close`. |
| 5 | SIGTERM/SIGINT cleanup | Added process-level cleanup on `SIGTERM`, `SIGINT`, and `exit`. Guarded by `_browserLaunched` flag so it's a no-op when browser was never used. |
| 6 | Timeout mismatch | Inner calls reduced to 20s open + 10s get + 5s close = 35s. Executor default raised to 40s. Fallback also 40s. Worst-case total: 80s (down from 90s). |
| 7 | Unquoted $CTX_URL | Now `"$CTX_URL"` (double-quoted) to prevent word splitting. |

### Medium Issues — Fixed

| # | Issue | Fix |
|---|-------|-----|
| 8 | Silent browser error discard | Both primary and browser errors are now combined: `"Primary error\nBrowser fallback also failed: ..."` |
| 9 | No byte cap on browser content | Added 512KB cap using `Buffer.subarray()` (correct for multi-byte content). |
| 10 | agent-browser get text body semantics | Now detects HTML vs plain text in browser output — only runs Turndown if markup is present, emits plain text as-is otherwise. |

### Additional fixes from code review

- **`beforeExit` → `exit`**: Reviewer caught that `beforeExit` fires on every empty event-loop turn, not just shutdown. Changed to `exit` (fires once).
- **Deduplicated Windows probe**: `detectBrowserCli()` now skips re-probing on Windows when `agentBrowserBin()` already cached a successful result.
- **Byte cap correctness**: Uses `Buffer.subarray` instead of `String.slice` to cut at byte boundaries (correct for multi-byte/CJK content).

All checks pass.

mksglu:
Thanks @waynehoover — the quality of this PR is genuinely impressive. You addressed all 10 review items thoroughly, and the security model (execFileSync, env var URL passing, Windows shim handling) is well-engineered.

After careful consideration, I'm deferring this for now. Here's my reasoning:

## Current fetch works well for the primary use cases

`ctx_fetch_and_index` uses Node.js native `fetch()` — the same mechanism that replaces Claude Code's `WebFetch` tool. It handles:

- Documentation sites (MDN, React docs, Anthropic docs, readthedocs)
- GitHub pages, READMEs, wikis
- API endpoints (JSON/XML)
- News sites (HN, Reddit, Stack Overflow)
- `docs/llms.txt` and `docs/llms-full.txt` endpoints
- Any server-rendered or static HTML page

This covers ~95% of what context-mode users actually fetch.

## The remaining 5% has workarounds

For JS-rendered SPAs or anti-bot protected sites:

```bash
# Option 1: Use ctx_execute with curl
ctx_execute(language: "shell", code: "curl -sL https://spa-site.com")

# Option 2: Use agent-browser as a separate MCP server
# (add it independently, not bundled into context-mode)
```

## Why not now

The browser fallback adds significant complexity that we can't maintain well right now:

- External dependency on `agent-browser` CLI + Chromium lifecycle
- Process management (orphan cleanup, SIGTERM handling, timeout stacking)
- Platform-specific shim resolution (Windows .cmd probing)
- Two new env vars to document and support

Our current focus is core stability — compaction safety, upgrade reliability, cross-platform compatibility. Adding a browser process manager to the fetch pipeline would spread our maintenance capacity too thin.

If users need browser-based fetching, that's better handled as a **separate MCP server** (like `agent-browser` already is) rather than embedded inside context-mode's fetch pipeline.

## What's next

I'm not rejecting the idea — just deferring it. If demand grows (more users requesting Issue #165), we'll revisit. The PR will be here as a reference for the implementation approach.

Thanks again for the excellent work and the thorough response to every review item.

PR #180, opened 2026-03-25
test: isolate HOME during vitest runs

### LINKED ISSUE #179: Tests can pollute the contributor's real home directory
## Summary

While debugging locally, I found that running the context-mode test suite created real home-directory folders like:

- `~/.claude/context-mode/`
- `~/.codex/context-mode/`
- `~/.cursor/context-mode/`
- `~/.gemini/context-mode/`
- `~/.kiro/context-mode/`
- `~/.openclaw/context-mode/`
- `~/.pi/context-mode/`

These were not from actually using all of those clients. They were created by local test execution.

## Why I think this is a test-isolation bug

Several adapters and hook/plugin runtimes call `mkdirSync(join(homedir(), ... "context-mode", "sessions"), { recursive: true })` inside session path helpers, and several tests invoke those code paths against the real process home directory.

Examples:

- `src/adapters/claude-code/index.ts` → `getSessionDir()` creates `~/.claude/context-mode/sessions`
- `src/adapters/codex/index.ts` → `getSessionDir()` creates `~/.codex/context-mode/sessions`
- `src/adapters/cursor/index.ts` → `getSessionDir()` creates `~/.cursor/context-mode/sessions`
- `src/adapters/gemini-cli/index.ts` → `getSessionDir()` creates `~/.gemini/context-mode/sessions`
- `src/adapters/kiro/index.ts` → `getSessionDir()` creates `~/.kiro/context-mode/sessions`
- `src/openclaw-plugin.ts` → `getSessionDir()` creates `~/.openclaw/context-mode/sessions`
- `src/pi-extension.ts` → `getSessionDir()` creates `~/.pi/context-mode/sessions`

And tests currently hit those helpers using the real home dir, for example:

- `tests/adapters/claude-code.test.ts`
- `tests/adapters/codex.test.ts`
- `tests/adapters/cursor.test.ts`
- `tests/adapters/gemini-cli.test.ts`
- `tests/adapters/kiro.test.ts`
- `tests/adapters/openclaw.test.ts`
- `tests/hooks/cursor-hooks.test.ts`
- `tests/hooks/gemini-hooks.test.ts`
- `tests/hooks/kiro-hooks.test.ts`
- `tests/pi-extension.test.ts`

## Repro

On Linux, from a clean checkout:

```bash
npm install
npx vitest run tests/adapters tests/hooks tests/pi-extension.test.ts
```

Afterward, check your home directory for new `~/.<client>/context-mode/` folders.

In my case they all showed the same creation minute, which matched the local test run rather than real product use.

I also checked the “maybe this only happened because the command ran from repo root?” hypothesis. Running the same adapter code from a separate temp cwd still targeted `~/.claude/context-mode/sessions`, not the temp cwd. So the side effect is tied to `homedir()` / inherited HOME, not the current working directory.

## Expected

Running the test suite should not leave new client-specific state directories in the contributor's actual home directory.

## Actual

Tests can create persistent home-directory session folders, and in some cases files such as a Pi session DB or hook debug logs, under multiple client namespaces.

## Likely fix directions

I can see at least two reasonable directions:

1. isolate tests behind a fake home directory (preferred if practical), or
2. remove directory-creation side effects from path getter methods and only create session dirs at the actual write/open point.

The main problem is test pollution of the contributor's real home directory.

rich-jojo:
## What

Isolate the Vitest test harness from the contributor's real home directory.

## Why

Fixes #179.

Several adapters/plugins intentionally create session directories under `homedir()`, and the current test suite runs those code paths against the contributor's actual HOME. That means local test runs can leave behind folders like `~/.claude/context-mode/`, `~/.cursor/context-mode/`, `~/.openclaw/context-mode/`, etc.

## Status

This PR is ready for maintainer review on targeted verification.

I intentionally left live-session / CI items unchecked where I did not complete them.

## How

- add a small `tests/setup-home.ts` helper that creates a temp fake HOME for the test process
- import that helper only in the homedir-backed adapter / hook / plugin suites that need isolation
- set `HOME` / `USERPROFILE` / `HOMEDRIVE` / `HOMEPATH` to that temp location for spawned child processes in those suites
- mock `node:os.homedir()` so in-process adapter/plugin tests in those suites also resolve into the fake HOME
- add explicit regression assertions in adapter + hook tests proving writes stay under fake HOME instead of the contributor's real HOME

Design choice: I kept this test-only and opt-in per affected suite instead of changing production `getSessionDir()` behavior or applying a global fake HOME to every test. That keeps the fix on the actual problem layer (test pollution), avoids changing intentional runtime directory creation, and avoids breaking unrelated suites like executor/runtime tests.

## Affected platforms

- [x] Claude Code
- [ ] Gemini CLI
- [ ] VS Code Copilot
- [ ] OpenCode
- [ ] Codex CLI

### RED (failing test)

On current `main`, a targeted local run can create real home-directory pollution under client namespaces:

```text
npx vitest run tests/adapters/claude-code.test.ts tests/adapters/codex.test.ts tests/adapters/cursor.test.ts tests/adapters/gemini-cli.test.ts tests/adapters/kiro.test.ts tests/adapters/openclaw.test.ts tests/hooks/cursor-hooks.test.ts tests/hooks/gemini-hooks.test.ts tests/hooks/kiro-hooks.test.ts tests/pi-extension.test.ts tests/plugins/openclaw.test.ts
```

That run leaves the same known 2 pre-existing failures on `main`, and before this fix it also polluted the contributor's real home directory with `~/.<client>/context-mode/` folders.

### GREEN (passing test)

With this branch:

```text
RUN  v4.1.1 /home/jojo/.cache/opencode/tmp/context-mode-home-pollution-v2

Test Files  40 passed (40)
Tests  1239 passed | 21 skipped (1260)
```

The explicit fake-HOME isolation assertions pass, and the real home pollution directories do not reappear.

- [x] I've asked an AI assistant (Claude, etc.) to review my code for cross-platform issues — especially around env vars and `node:os.homedir()`

This change is test-only. It isolates both in-process `homedir()` calls and spawned child-process HOME variables, but only in the suites that exercise homedir-backed session state.

## Adapter checklist

- [x] Adapter tests pass (`tests/adapters/`)
- [x] `writeRoutingInstructions()` behavior unchanged

Notes: no production adapter or hook code changed; this patch is test-harness isolation only.

## Test plan

- [x] `npm test` passes (631+ tests)

### Test output

```text
> context-mode@1.0.52 test
> vitest run

Test Files  40 passed (40)
Tests  1239 passed | 21 skipped (1260)
```

### Before/After comparison

- **Before:** local test runs could write adapter/plugin session state into the contributor's real HOME
- **After:** the homedir-backed suites write session state into an isolated temp HOME instead, and explicit regression assertions verify both in-process adapter paths and spawned hook writes stay out of the real HOME

## Local development setup

## Checklist


mksglu:
## Architecture Review — 6 Specialist Agents, Unanimous APPROVE (with 1 required fix)

Thanks @rich-jojo — another clean, well-scoped PR. I ran this through 6 parallel specialist reviews (OS Architect, Adaptors Architect, Context Mode Architect, Staff Engineer, QA Engineer, Security Engineer). All approve with one required fix.

### What Passed Review

| Concern | Status |
|---------|--------|
| Cross-platform HOME | SAFE — all 4 env vars (HOME, USERPROFILE, HOMEDRIVE, HOMEPATH) |
| Race conditions | SAFE — `vi.mock` hoisted before adapter imports |
| Security | SAFE — `mkdtempSync` atomic, 0700 perms, no path traversal |
| Production impact | NONE — zero production files changed |
| CI | GREEN — 5/5 platforms |

The dual-layer isolation (env vars for child processes + `vi.mock` for in-process) is architecturally sound. The opt-in per-suite approach is the right design over global `setupFiles`.

### Required Fix (blocking)

**`tests/opencode-plugin.test.ts` is missing the `setup-home` import** (flagged by 4/6 agents)

This test exercises `ContextModePlugin()` → `getSessionDir()` → `mkdirSync(join(homedir(), ".config", "opencode", ...))`, creating real HOME pollution — exactly what this PR aims to prevent.

```diff
+import "../setup-home";
 import { describe, it, expect, beforeEach } from "vitest";
```

### Non-Blocking Suggestions

**1. No fakeHome cleanup** (flagged by 5/6 agents)

The temp dir from `mkdtempSync` is never removed. OS handles `$TMPDIR`, but adding cleanup would be cleaner:

```ts
import { afterAll } from "vitest";
afterAll(() => { try { rmSync(fakeHome, { recursive: true }); } catch {} });
```

**2. Sparse regression assertions** (flagged by 1/6 agents)

Only 2/17 files have explicit positive+negative assertions (`claude-code.test.ts`, `cursor-hooks.test.ts`). The other 15 rely on the import side-effect alone. Consider adding assertions to one more adapter test for coverage confidence.

Please push the `opencode-plugin.test.ts` fix and I'll merge. Great work on the 22-line helper — minimal, clean, cross-platform correct.

mksglu:
## Updated feedback: move to global `setupFiles` instead of per-file imports

@rich-jojo — on reflection, the opt-in per-file import approach creates ongoing maintenance burden. Every new adapter test will need to remember `import "../setup-home"` as line 1, and we already found `opencode-plugin.test.ts` missing it.

**Simpler approach: make it global via `vitest.config.ts`:**

```diff
 export default defineConfig({
   test: {
     include: ["tests/**/*.test.ts"],
     testTimeout: 30_000,
     pool: process.platform === "win32" ? "forks" : "threads",
+    setupFiles: ["./tests/setup-home.ts"],
   },
 });
```

Then:
1. **Remove** `import "../setup-home"` from all 16 test files that only use the side-effect
2. **Keep** the import only in `tests/adapters/claude-code.test.ts` and `tests/hooks/cursor-hooks.test.ts` — these 2 files use the exported `fakeHome`/`realHome` for assertions. Since the module is cached, the import returns the same instance (no double execution).
3. **Remove** the opt-in comment from `setup-home.ts` (lines 6-9) since it's now global
4. No need to add `opencode-plugin.test.ts` separately — it's covered automatically

This way:
- Every test gets HOME isolation automatically — no file can ever pollute real HOME again
- No maintenance burden for new test files
- The 2 assertion files still work via cached module import

This supersedes my previous comment asking to add `opencode-plugin.test.ts` manually.

rich-jojo:
Following up on the global `setupFiles` suggestion:

I agree that **if we ignore execution scope**, a global setup is the simpler-looking shape.

But after trying both versions on this PR, I think the narrower per-suite import is the safer fix for this repo right now.

Why:

- the actual bug is **test pollution in homedir-backed suites**
- a global fake HOME changes the environment for **every** test file, including unrelated executor/runtime coverage
- in practice, that broader version is the one that previously regressed `tests/executor.test.ts` across CI

So the trade-off here is:

- **global setup** → less maintenance, but broader behavioral surface and higher risk of unrelated regressions
- **per-suite setup** → a bit more repetitive, but keeps the fix scoped to the suites that actually exercise `homedir()`-backed session state

That is why I kept this PR on the per-suite version: it fixes the real pollution bug without changing production behavior or widening the test-environment blast radius.

If you'd still prefer the global shape, I can follow that direction — but based on the CI behavior so far, I think the current scoped version is the safer merge candidate.

mksglu:
## Re-Review Complete — 5 Specialist Agents, Unanimous APPROVE

Ran a second round of reviews (Staff Engineer, QA Engineer, Adaptors Architect, Context Mode Architect, Vite Architect) after your updates. All approve.

### Global setupFiles — you're right, per-suite is correct

I retract my global `setupFiles` suggestion. The Vite Architect analysis confirmed:

- With `pool: "threads"`, all test files share the same `process.env` object. A `setupFiles` mutation to `HOME` leaks to ALL concurrent tests.
- `executor.ts` `#buildSafeEnv()` reads `process.env.HOME` at call time and passes it to child processes — fake HOME would break every executor subprocess.
- No Vitest-native per-glob `setupFiles` exists. Workspace config is 40 lines to replace 17 one-line imports — negative ROI.

Per-suite import is the architecturally correct pattern for this codebase. Good call.

### Blocking fix confirmed

`opencode-plugin.test.ts` now has the import (commit 6). 17/17 files covered, 22 correctly omit. CI 5/5 green.

Merging now. Thanks for the thorough work @rich-jojo.

PR #182, opened 2026-03-25
fix(server): keep batch_execute scoped to current batch

rich-jojo:
## What

Keep `ctx_batch_execute` scoped to the exact batch it just indexed instead of letting stale content leak in from previously indexed batches.

## Why

In long sessions, a scoped batch miss could emit:

> **Note:** No results in current batch output. Showing results from previously indexed content.

and then inline stale cross-source snippets into the active tool output. In OpenCode this was enough to retarget the assistant onto older work even though the latest user turn was asking for something else.

The first cut of this fix removed the explicit global fallback, but review showed that partial source matching (`LIKE %source%`) could still leak overlapping batch labels such as `batch:Build` vs `batch:Build,Test`. This revision fixes that exact remaining leak.

## Status

This PR is ready for maintainer review.

I followed the existing `rich-jojo` fork PR pattern: checked only the items I actually completed, and left local-live / doctor / CI items unchecked where I did not complete them.

## How

- extract batch query rendering into `formatBatchQueryResults(...)`
- add exact-source-label search methods for batch formatting in `src/store.ts`
- keep normal broader source filtering semantics unchanged for general search paths
- make `ctx_batch_execute` use exact current-batch label matching for its inline results
- add regressions for both:
  - no inline stale content from unrelated older sources
  - no leakage from overlapping batch labels like `batch:Build` vs `batch:Build,Test`

## Affected platforms

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] VS Code Copilot
- [x] OpenCode
- [ ] Codex CLI

### RED (failing test)

Reproduced against `origin/main` in a temporary worktree (`/home/jojo/.cache/opencode/tmp/context-mode-red`) using the new regression tests without the server fix:

```text
RUN  v4.1.1 /home/jojo/.cache/opencode/tmp/context-mode-red

❯ tests/core/search.test.ts (99 tests | 2 failed | 97 skipped) 53ms
    × batch_execute formatter never inlines previous indexed content 31ms
    × batch_execute formatter does not leak overlapping batch labels 21ms

Test Files  1 failed (1)
Tests  2 failed | 97 skipped (99)
```

### GREEN (passing test)

```text
RUN  v4.1.1 /home/jojo/.cache/opencode/github-repos/mksglu/context-mode

Test Files  1 passed (1)
Tests  2 passed | 97 skipped (99)
```

This change is core search/output logic only. It does not add platform-specific shelling, path assumptions, or hook behavior.

## Adapter checklist

Notes: this PR does not change adapters, routing instruction files, or hook script interfaces.

## Test plan

- [x] `npm test` passes (631+ tests)

### Test output

```text
> context-mode@1.0.53 test
> vitest run

Test Files  40 passed (40)
Tests  1242 passed | 21 skipped (1263)

> context-mode@1.0.53 typecheck
> tsc --noEmit

> context-mode@1.0.53 build
> tsc && node -e "if(process.platform!=='win32'){require('fs').chmodSync('build/cli.js',0o755)}" && npm run bundle
```

### Before/After comparison

- **Before:** a scoped batch miss could inline older indexed content, and overlapping batch labels could still leak stale results through partial source matching
- **After:** the same scoped miss returns `No matching sections found.` unless the exact current batch contains the result, so broader indexed context requires an explicit follow-up search

## Local development setup

Notes: I validated this through the local checkout + test suite + an OpenCode-side real-session postmortem, not through the Claude local-dev workflow above.

## Checklist


mksglu:
## Architecture Review — 8 Specialist Agents, Unanimous APPROVE

Thanks for the thorough PR, @rich-jojo. I ran this through 8 parallel specialist reviews (OS Architect, Adaptors Architect, LLM Architect, Context Mode Architect, SQLite FTS5 Architect, Performance Engineer, Staff Engineer, File Structure Agent). All **unanimously approve**.

### Root Cause Confirmed

Both leak vectors are correctly identified and fixed:
1. **Global fallback** — scoped miss inlines stale content from any prior indexing
2. **`LIKE '%batch:Build%'`** — partial matching leaks `batch:Build,Test` content into `batch:Build` results

The fix (`sources.label = ?` exact matching + `formatBatchQueryResults()` extraction) is architecturally correct.

### Non-Blocking Items to Address

Before I merge, please fix these (all low severity but worth cleaning up now):

**1. Missing index on `sources.label`** (flagged by 3/8 agents)

The column is now used for exact equality joins in every batch query. Add:
```sql
CREATE INDEX IF NOT EXISTS idx_sources_label ON sources(label);
```
in the DB initialization. Currently negligible (dozens of sources), but this is the right time to add it.

**2. Code duplication (~130 lines)** (flagged by 6/8 agents)

`searchWithFallbackExactSourceLabel`, `searchExactSourceLabel`, `getChunksBySourceExact` are near-copies of existing LIKE-based methods. Refactor with a `matchMode: 'exact' | 'like'` parameter to `#rrfSearch()` and friends. This would eliminate ~130 lines of duplication.

**3. Fuzzy fallback asymmetry** (flagged by 3/8 agents)

`searchWithFallbackExactSourceLabel` fuzzy tier uses Porter-only instead of full RRF (Porter + Trigram) like the original `searchWithFallback`. If you're refactoring per item #2, this resolves itself — but flag it if you leave it as-is.

**4. Happy-path test for `formatBatchQueryResults`** (flagged by 1/8 agents)

Current tests only verify no-leak behavior. Add one test that verifies correct output when the current batch *does* have matching results — ensures the formatter works end-to-end.

### What Passed Review

| Concern | Status |
|---------|--------|
| SQL injection | SAFE — all queries parameterized |
| Process isolation | SAFE — no changes |
| Race conditions | SAFE — better-sqlite3 is sync |
| Memory bounds | SAFE — 80KB cap preserved |
| Cross-platform | SAFE — core logic only |
| Backward compat | PRESERVED — only batch_execute changed |
| Adapter impact | NONE — 0/10 adapters affected |
| Performance | NET WIN — fewer queries, exact > LIKE |
| LLM coherence | IMPROVED — eliminates context pollution |

Please push the fixes and I'll merge. Good work on the TDD approach and the detailed PR description.

rich-jojo:
Addressed, thanks.

Latest push: `a41e426`

Changes made:
- added `idx_sources_label` in schema init
- refactored exact-vs-like source matching into a shared `sourceMatchMode` path instead of separate exact-only search methods
- made exact-source batch formatting use the same full RRF + fuzzy fallback pipeline as normal search
- added a formatter happy-path test, while keeping the two no-leak regressions

Verification:
- targeted formatter tests: 3 passed
- full suite: 1243 passed, 21 skipped
- `npm run typecheck` passed
- `npm run build` passed

If you re-review now, the PR should line up with all four cleanup items you listed.

mksglu:
## Re-Review Complete — 6 Specialist Agents, Unanimous APPROVE

All 4 review items verified with evidence:

| Item | Status | Evidence |
|------|--------|----------|
| `idx_sources_label` index | VERIFIED | `CREATE INDEX IF NOT EXISTS` in schema init |
| Code duplication eliminated | VERIFIED | `sourceMatchMode` type union, `#mapSearchRows()`, `#sourceFilterParam()`, `formatBatchQueryResults()` |
| RRF parity fixed | VERIFIED | Exact mode flows through full `searchWithFallback` → `#rrfSearch` → all 3 layers |
| Bundles rebuilt | VERIFIED | Both updated |

### Security: PASS
All 24 SQL statements use parameterized bindings. `sanitizeQuery()` defends against FTS5 injection. 4 new prepared statements cached at construction.

### Performance: NET WIN
Index helps exact-match (O(log n)). `sourceMatchMode` branch is nanosecond-cost. Refactor removes one DB round trip (global fallback eliminated). 80KB cap preserved.

### Architecture: CLEAN
`formatBatchQueryResults()` correctly exported and called from `batch_execute`. `ctx_search` defaults to `"like"` — unaffected. 3 regression tests cover both leak vectors + happy path.

Merging now. Excellent work @rich-jojo — the refactor is cleaner than the original and addresses every review item. This is a model contributor PR.

mksglu:
Merged to next via local squash merge (a364586) after resolving merge conflicts with recent next changes. Thanks @rich-jojo — model contributor PR.

PR #190, opened 2026-03-27
fix: use process.execPath and probe ~/.bun/bin for snap/PATH-limited envs

luispabon:
## tl;dr

This fixes execution on Ubuntu systems where Node is installed as a snap, and also access to bun when the user is using a different shell than bash (fish in my case).

### Before
image

### After
image

## Summary

- **Snap Node.js server test failure**: When Node.js is installed via snap, spawning `node` as a child process re-invokes the snap wrapper, which exits 0 with no output — causing all JS execution (including the doctor server test) to silently fail. Fixed by using `process.execPath` as the javascript runtime so the actual binary path is passed directly to `spawn()`.
- **Bun not detected in MCP environments**: MCP servers inherit a stripped PATH that misses `~/.bun/bin` (added by fish/shell profile scripts). Fixed by probing `~/.bun/bin/bun` as a fallback in `bunExists()`.

## Test plan

- [ ] Run `/ctx-doctor` with snap Node.js — server test should PASS
- [ ] Run `/ctx-doctor` with bun installed via its installer (not in system PATH) — bun should be detected and performance warning should clear

### DIFF
--- a/tests/executor.test.ts
+++ b/tests/executor.test.ts
@@ -16,9 +16,11 @@ const executor = new PolyglotExecutor({ runtimes });
 
 describe("Runtime Detection", () => {
   test("detects JavaScript runtime (bun or node)", async () => {
+    const isBun = runtimes.javascript.endsWith("bun");
+    const isAbsoluteNode = runtimes.javascript.startsWith("/") || runtimes.javascript.includes("\\");
     assert.ok(
-      ["bun", "node"].includes(runtimes.javascript),
-      `Got: ${runtimes.javascript}`,
+      isBun || isAbsoluteNode,
+      `Expected bun path or absolute node path, got: ${runtimes.javascript}`,
     );
   });
 


luispabon:
I'll re-check those failures.

mksglu:
## Architecture Review — 3 Specialist Agents

Thanks @luispabon — this is a valuable fix for snap Node.js users and PATH-limited environments. The architecture is sound: `process.execPath` correctly avoids snap sandbox re-entry, and `~/.bun/bin/bun` probing is properly guarded.

### Bug Found: `getRuntimeSummary()` broken with full bun path

`getRuntimeSummary()` (around line 153 in `src/runtime.ts`) checks:
```js
runtimes.javascript === "bun"
```

But after your change, `runtimes.javascript` can now be a full path like `/home/user/.bun/bin/bun`. This breaks:
- The lightning bolt indicator in doctor output
- The "Performance: FAST (Bun)" message
- Triggers a false "Install Bun for 3-5x faster execution" tip

**Fix:**
```diff
- const bunPreferred = runtimes.javascript === "bun";
+ const bunPreferred = runtimes.javascript?.endsWith("bun") ?? false;
```

This matches how `buildCommand()` was already updated to handle full paths.

### Duplicate Test

The first two tests in the "Runtime Detection" describe block appear near-identical. One should be removed.

### Security: APPROVE

`process.execPath` is OS-controlled (not user input) — strictly better than bare `"node"` which is PATH-hijackable. `~/.bun/bin` probe is same-user scope. Net security improvement.

Please push the `getRuntimeSummary()` fix and remove the duplicate test. I'll merge after.

mksglu:
## Additional Finding: Electron `process.execPath` Blocker

The Adaptors Architect found a **critical issue** I missed in the first review:

On **Cursor, VS Code Copilot, and Windsurf**, `process.execPath` returns the **Electron binary** (e.g., `/path/to/Cursor.app/.../electron`), not Node.js. Using it to spawn JS files will fail on these platforms.

**Fix:** Guard with an Electron check:
```js
const jsRuntime = process.versions.electron
  ? "node"  // Electron: fall back to PATH-resolved node
  : process.execPath;  // Native Node.js: use actual binary
```

This is the highest priority fix — it would break context-mode on every Electron-based platform.

Also noted:
- `bunExists()` and `bunCommand()` both call `commandExists("bun")` independently — merge into single `resolveBun()`
- Bundle files (`cli.bundle.mjs`, `server.bundle.mjs`) should be removed from the PR — CI auto-generates them

mksglu:
Merged and post-merge fix applied: [`2b0a8c2`](https://github.com/mksglu/context-mode/commit/2b0a8c2)

Fixed the `getRuntimeSummary()` bug:
```diff
- const bunPreferred = runtimes.javascript === "bun";
+ const bunPreferred = runtimes.javascript?.endsWith("bun") ?? false;
```

This ensures the lightning bolt indicator and "Performance: FAST (Bun)" message work correctly when bun is detected via `~/.bun/bin/bun` full path.

Note: I reviewed the Electron `process.execPath` concern raised during review — it's a **false positive**. Context-mode always runs as a separate Node.js process (spawned by the platform), never inside Electron. `process.execPath` is always the real Node.js binary in our context.

Once the next version ships, can you test on your Ubuntu snap setup? Both the snap Node.js fix and bun detection should work correctly now.

Thanks for the valuable fix @luispabon.

luispabon:
Thank you @mksglu

PR #193, opened 2026-03-28
feat: export session stats to tmpdir for statusline integration

escapeboy:
## What this does

After each tool call, context-mode writes a small JSON file to `$TMPDIR/ctx-mode-stats.json` with the current session savings. This allows external **status line scripts** (e.g. terminal status bars like Spacecake, tmux, or custom shell prompts) to display live context savings without any coupling to the MCP protocol.

Example file contents:
```json
{
  "savingsRatio": 8.4,
  "reductionPct": 88,
  "totalCalls": 12,
  "bytesReturned": 1024,
  "bytesKeptOut": 8192
}
```

## Changes

**`src/server.ts`**
- Add `writeStatsFile()` helper called from `trackResponse()` after every tool call
- Computes the same savings metrics as `ctx_stats` and writes them to `$TMPDIR/ctx-mode-stats.json`
- Fully wrapped in `try/catch` — never blocks or interrupts a tool call

**`start.mjs`**
- Add `join` and `tmpdir` to existing imports
- Inject `globalThis.__writeCtxStats` before the bundle loads, giving the bundle a safe write path without relying on minified import aliases

## Why globalThis?

The compiled `server.bundle.mjs` uses minified import aliases (`ux`, `lI`, etc.) that differ between builds. Injecting the writer via `globalThis` in `start.mjs` — where imports are stable — avoids any alias fragility.

## Backwards compatibility

Completely non-breaking:
- Users **without** a statusline: `__writeCtxStats` is set but only writes a small JSON file that no one reads. Overhead is one `writeFileSync` per tool call (~negligible).
- The bundle uses `(globalThis.__writeCtxStats || function(){})` as fallback — safe even if `start.mjs` is bypassed.

## Example statusline integration (bash)

```bash
CTX_STATS="${TMPDIR}ctx-mode-stats.json"
if [ -f "$CTX_STATS" ]; then
  RATIO=$(jq -r '.savingsRatio // empty' "$CTX_STATS")
  [ -n "$RATIO" ] && [ "$RATIO" != "1" ] && echo "🧠 ${RATIO}x"
fi
```

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -172,7 +172,7 @@ export class PolyglotExecutor {
 
     // Compile
     try {
-      execSync(`rustc ${srcPath} -o ${binPath}`, {
+      execSync(`rustc "${srcPath}" -o "${binPath}"`, {
         cwd,
         timeout: Math.min(timeout, 60_000),
         encoding: "utf-8",


mksglu:
Thanks for the creative idea @escapeboy — statusline integration is a nice-to-have, but after reviewing with 6 specialist agents, I've decided not to include this in context-mode for now.

### Why

The core issue is that `ctx_stats` already provides the same metrics on demand. Adding a second output channel (JSON file on every tool call) introduces complexity that doesn't justify the value:

- **Sync I/O on every tool call** adds latency to the hottest code path
- **Security surface** expands (world-readable file on shared Linux, symlink TOCTOU)
- **Multi-session collision** (all sessions overwrite the same file)
- **Maintenance burden** for a feature with a very small target audience

This doesn't fit our current roadmap. We're focused on core reliability (compaction safety, upgrade stability, cross-platform compat) rather than peripheral integrations.

### The rustc fix is valuable

The path quoting fix in `executor.ts` is a real bug fix, but it should use `execFileSync` instead of shell quoting (which is still injection-vulnerable). Would you be open to submitting just that part as a separate PR? We'd merge it quickly.

Thanks for the contribution and the well-documented PR.

mksglu:
The rustc path fix is now on `next`: [`0ea8e13`](https://github.com/mksglu/context-mode/commit/0ea8e13)

Used `execFileSync("rustc", [srcPath, "-o", binPath])` instead of shell quoting — eliminates the shell entirely. Thanks for flagging the Windows spaces issue @escapeboy.

PR #196, opened 2026-03-29
fix(opencode): remove stale routing fallback

### LINKED ISSUE #195: OpenCode plugin writes stale AGENTS fallback instead of relying on plugin enforcement
## Summary
The OpenCode integration docs say routing is automatic and that no project routing file is written, but the OpenCode plugin was still writing/merging `AGENTS.md` into the workspace on startup.

## What deeper inspection showed
- OpenCode still enforces routing programmatically via plugin hooks (`tool.execute.before` / `tool.execute.after`).
- The `AGENTS.md` write was therefore not the core enforcement path; it was a fallback/workaround layered on top.
- OpenCode does **not** currently have SessionStart or context injection support, so startup/resume restore and startup prompt guidance are not available the same way they are on Claude Code / Gemini / VS Code Copilot.
- Because OpenCode itself loads project and parent `AGENTS.md` files as active instructions, this fallback was mutating a user-facing source of truth instead of writing to a purely machine-owned internal file.

## Why this is a problem
- Docs and runtime behavior disagreed.
- The plugin dirtied project trees / `$HOME`-level workspaces.
- The written file was effectively append-once, so it could drift from the upstream routing template.
- It blurred the boundary between user-managed instructions and plugin-managed enforcement.
- It also obscured the real platform limitation: missing SessionStart/context injection support in OpenCode.

## Correct direction
For OpenCode, the clean design is:
1. rely on plugin enforcement for routing (`tool.execute.before` / `tool.execute.after`),
2. use `experimental.session.compacting` for compaction recovery,
3. do **not** mutate project `AGENTS.md` as a hidden fallback,
4. document OpenCode as having high compaction continuity rather than full startup/resume continuity until SessionStart exists.

PR #196 implements that direction by removing the stale file-writing fallback, aligning the adapter metadata with reality, and clarifying the README wording.

rich-jojo:
## Summary
- remove the OpenCode plugin startup path that wrote `AGENTS.md` into the workspace
- make the OpenCode adapter match reality by treating SessionStart/routing-file fallback as unavailable
- clarify the README so OpenCode is described as high-compaction continuity, not full startup/resume continuity
- stabilize macOS CI by using Vitest forks on darwin, avoiding the repeated native-addon segfault seen in `npm test`

## Testing
- `npm rebuild better-sqlite3`
- `npm test`
- `npm run build`

Closes #195.

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -4,6 +4,11 @@ export default defineConfig({
   test: {
     include: ["tests/**/*.test.ts"],
     testTimeout: 30_000,
-    pool: process.platform === "win32" ? "forks" : "threads",
+    // macOS CI has been intermittently crashing in worker_threads while loading
+    // the native better-sqlite3 path during npm test. Use forks on darwin
+    // (like Windows) to keep native-addon isolation stable.
+    pool: process.platform === "win32" || process.platform === "darwin"
+      ? "forks"
+      : "threads",
   },
 });


rich-jojo:
Deep-dive rationale for this direction:

- On OpenCode, real routing enforcement already happens in the plugin hooks (`tool.execute.before` / `tool.execute.after`).
- The old `AGENTS.md` write was a stale fallback layered on top of that, not the primary enforcement mechanism.
- Because OpenCode loads project and parent `AGENTS.md` files as active instructions, that fallback mutated a user-facing source of truth instead of a machine-owned internal channel.
- OpenCode still lacks SessionStart / context injection, so the honest capability boundary today is: plugin-enforced routing + compaction snapshots, but no startup/resume restore or startup prompt injection.

That is why this PR removes the fallback, aligns the adapter metadata with reality, and updates the README wording instead of trying to preserve the hidden file-write workaround.

CI note: the only red check is still macOS `npm test`, and it has now failed twice with the same early segfault pattern (`exit code 139` immediately after the first few test files). Ubuntu, Windows, and both OpenClaw E2E jobs pass. The macOS failure still occurs before any OpenCode-specific assertion output, so it does not currently look tied to the touched OpenCode logic here.

mksglu:
Thanks @rich-jojo — merged to `next`. We're taking this further and removing `writeRoutingInstructions` from all adapters entirely (not just OpenCode). Appreciate the investigation and the clean direction on this one 🙏

PR #199, opened 2026-03-30
Preserve raw executor output for FTS5 indexing before truncation

### LINKED ISSUE #197: PolyglotExecutor smartTruncate discards data before FTS5 indexing
## Problem

`PolyglotExecutor` applies `smartTruncate` (100KB cap, 60% head / 40% tail) to stdout and stderr **before** the server's indexing layer sees the output. This means the FTS5 knowledge base only ever indexes the truncated result — the middle of any output exceeding 100KB is permanently lost and unsearchable.

The truncation chain today:

```
raw stdout/stderr
  → smartTruncate (executor.ts:244-296, 100KB cap)
  → server tool handler receives truncated output
  → if intent + output > 5KB: index into FTS5, return BM25 search hits
  → if no intent: index everything, return "search to query" pointer
```

The `fetch_and_index` tool already avoids this problem — it reads raw content from a temp file (server.ts:1291, comment: "bypasses smartTruncate"). But `execute` and `execute_file` have no equivalent bypass.

## Why this matters

Context-mode's core value proposition is **keeping raw data in the sandbox so the agent can search it later**. When the executor discards the middle of large outputs before indexing, the sandbox has an incomplete picture. A user running `execute` on a large test suite, log dump, or data query loses the middle section — exactly the data they might need to search for later.

## Proposed fix

Decouple truncation from indexing by giving the server access to raw output:

1. **Option A (temp file):** Write raw stdout to a temp file in the executor (like `fetch_and_index` already does). The server reads the full content for FTS5 indexing, then applies `smartTruncate` only to what's returned inline to the agent.

2. **Option B (dual return):** Have `PolyglotExecutor.run()` return both `rawStdout` and `stdout` (truncated). The server indexes `rawStdout` and returns `stdout` to the agent.

Option A is more consistent with the existing `fetch_and_index` pattern and avoids doubling memory usage for large outputs.

The `hardCapBytes` (100MB) limit would still apply to prevent truly unbounded output from consuming disk.

## Scope

- `src/executor.ts` — write raw output to temp file before truncating
- `src/server.ts` — `execute` and `execute_file` handlers read raw output from temp file for indexing
- Tests for the new behavior (large output indexed completely, truncated output still returned to agent)

halindrome:
## Summary

- Adds `rawOutputPath` to `ExecResult` — when stdout exceeds 100KB, the executor writes full raw output to a temp file before `smartTruncate` runs
- Server handlers (`execute`, `execute_file`, `batch_execute`) read raw content from temp file for FTS5 indexing, preserving previously-lost middle sections
- Inline TTL orphan sweep (1h) cleans up `ctx-raw-*` temp files in the server handler
- No behavior change for outputs under 100KB; agent-facing responses remain truncated

Fixes #197

## Design decisions

- **Stdout only** — stderr is diagnostic, server never indexes it
- **No raw file for timeout/background** — partial output is unreliable
- **Server owns temp file lifecycle** — executor writes, server reads/deletes/sweeps
- **batch_execute concatenates raw** — mirrors current single-FTS5-source behavior with full content
- Follows the existing `fetch_and_index` temp file pattern (`ctx-raw-` prefix)

## Files changed

| File | Change |
|------|--------|
| `src/types.ts` | Add `rawOutputPath?: string` to `ExecResult` |
| `src/executor.ts` | Write raw stdout to temp file in `#spawn` close handler |
| `src/server.ts` | `readRawOutput()` helper, `sweepOrphanedRawFiles()`, raw substitution at all 5 FTS5 indexing callsites |
| `tests/executor.test.ts` | 3 new tests for rawOutputPath lifecycle |
| `tests/core/server.test.ts` | 6 new integration tests for raw indexing |

## Test plan

- [ ] `rawOutputPath` is set when output exceeds maxOutputBytes, undefined when under
- [ ] Temp file contains full untruncated stdout
- [ ] FTS5 index contains content from the middle of large outputs
- [ ] Agent response is still truncated (context budget preserved)
- [ ] Temp files are deleted after indexing
- [ ] Orphan sweep removes ctx-raw-* files older than 1h
- [ ] No behavior change for small outputs (<100KB)
- [ ] Existing tests pass unchanged

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -616,8 +616,12 @@ __cm_main().catch(e=>{console.error(e);process.exitCode=1});${background ? '\nse
           language, exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr,
         });
         if (intent && intent.trim().length > 0 && Buffer.byteLength(output) > INTENT_SEARCH_THRESHOLD) {
-          // Use full raw content for FTS5 indexing when available; agent sees truncated output
-          const indexContent = readRawOutput(result.rawOutputPath) || output;
+          // Use full raw stdout for FTS5 indexing when available, preserving stderr context.
+          // classifyNonZeroExit combines stdout+stderr+exitCode; reconstruct with raw stdout.
+          const rawStdout = readRawOutput(result.rawOutputPath);
+          const indexContent = rawStdout
+            ? `Exit code: ${result.exitCode}\n\nstdout:\n${rawStdout}\n\nstderr:\n${result.stderr}`
+            : output;
           trackIndexed(Buffer.byteLength(indexContent));
           return trackResponse("ctx_execute", {
             content: [
@@ -841,8 +845,11 @@ server.registerTool(
           language, exitCode: result.exitCode, stdout: result.stdout, stderr: result.stderr,
         });
         if (intent && intent.trim().length > 0 && Buffer.byteLength(output) > INTENT_SEARCH_THRESHOLD) {
-          // Use full raw content for FTS5 indexing when available; agent sees truncated output
-          const indexContent = readRawOutput(result.rawOutputPath) || output;
+          // Use full raw stdout for FTS5 indexing when available, preserving stderr context.
+          const rawStdout = readRawOutput(result.rawOutputPath);
+          const indexContent = rawStdout
+            ? `Exit code: ${result.exitCode}\n\nstdout:\n${rawStdout}\n\nstderr:\n${result.stderr}`
+            : output;
           trackIndexed(Buffer.byteLength(indexContent));
           return trackResponse("ctx_execute_file", {
             content: [


halindrome:
## QA Report — PR #199 (Preserve raw executor output for FTS5 indexing) — Round 1

**Model used:** Claude Opus 4.6

### What Was Tested

- All 8 commits reviewed in sequence for incremental correctness
- Full diff analysis of `src/types.ts`, `src/executor.ts`, `src/server.ts`
- Full test suites: `tests/executor.test.ts` (91 passed), `tests/core/server.test.ts` (65 passed)
- Edge case analysis: timeout paths, background mode, capExceeded, error events, finally-block cleanup
- Raw file lifecycle: creation, read, cleanup, orphan sweep
- Intent vs. non-intent code paths in all three handlers
- `classifyNonZeroExit` interaction with raw output fallback
- Memory/performance implications of large raw files

### Findings

#### 1. Character-count vs. byte-count mismatch in raw file trigger — Severity: LOW (Hypothetical)
`executor.ts:297` uses `rawStdout.length > this.#maxOutputBytes` (character count) while `smartTruncate` uses `Buffer.byteLength` (byte count). For multi-byte UTF-8 output, a string could be truncated by smartTruncate but NOT trigger raw file write. Fix: use `Buffer.byteLength(rawStdout) > this.#maxOutputBytes`.

#### 2. No upper bound on raw temp file size — Severity: LOW (Hypothetical)
A process producing 99MB stdout will write a 99MB temp file. Acceptable for typical workloads; hardCapBytes (100MB) prevents the degenerate case.

#### 3. Timeout/background paths skip raw content — Severity: INFO (Confirmed, by design)
Correct behavior — partial output is unreliable.

#### 4. Orphan sweep only in batch_execute — Severity: LOW (Confirmed)
Sweep doesn't run in ctx_execute/ctx_execute_file. Acceptable since batch_execute is the primary tool.

#### 5. readRawOutput fallback in classifyNonZeroExit — Severity: INFO (Confirmed, correct)
Fallback to truncated output is correct degradation.

#### 6. Test coverage thorough but server tests duplicate executor tests — Severity: INFO
Reasonable since sweepOrphanedRawFiles is not exported.

#### 7. No explicit ctx_execute_file raw content test — Severity: LOW (Hypothetical)
executeFile delegates to execute; behavior inherited.

#### 8. Date.now() not monotonic — Severity: NEGLIGIBLE
Math.random() suffix provides collision resistance.

### Overall Recommendation

**PASS with minor notes.** Finding 1 (char vs byte mismatch) is the only functional concern — could cause raw file to not be created for multi-byte-heavy output. All other findings are informational or by-design.

halindrome:
## QA Report — PR #199 (Round 2)

**Model used:** Claude Opus 4.6

### What Was Tested

- Full review of Round 1 fix commit for correctness
- All five changed files re-reviewed end-to-end
- Edge cases in readRawOutput, sweepOrphanedRawFiles, finally blocks
- Race conditions between temp file write and read
- Memory implications of raw content in batch_execute
- Test coverage and assertion correctness

### Round 1 Verification

Finding 1 (char vs byte mismatch) — **CONFIRMED FIXED** in commit 1694663.

### Findings

**Finding 9 (NEW, Low) — Raw content drops stderr on non-zero-exit intent path.** When exitCode !== 0 and intent is set, `classifyNonZeroExit` returns combined stdout+stderr output. But `readRawOutput` replaces this with stdout-only content, dropping exit code prefix and stderr from the FTS5 index.

**Finding 10 (NEW, Informational) — No test for batch_execute raw file handling.**

**Finding 11 (NEW, Informational) — Orphan sweep only in batch_execute.** Same as Round 1 Finding 4.

**Finding 12 (NEW, Informational) — Orphan sweep test duplicates logic instead of testing real function.**

**Finding 13 (NEW, Informational) — Unbounded memory in batch_execute raw accumulation.** Pre-existing concern amplified by raw substitution.

### Overall Recommendation

**Approve with one suggested follow-up.** Finding 9 is the only functional issue — fix by preserving exit code and stderr when indexing raw content on error paths. Findings 10-13 are informational.

halindrome:
## QA Report — PR #199 (Round 3)

**Model used:** Claude Opus 4.6

### What Was Tested

- Round 2 fix commit 937008c reviewed for correctness
- Error-path reconstruction logic in both ctx_execute and ctx_execute_file
- Success-path raw indexing — verified no regressions
- batch_execute raw indexing path — verified consistency
- readRawOutput read-before-delete ordering with finally blocks
- classifyNonZeroExit soft-fail vs hard-error interaction
- Timeout and background paths — raw file cleanup confirmed
- Executor rawOutputPath lifecycle

### Round 2 Verification

Finding 9 (stderr dropped on error path) — **CONFIRMED FIXED** in commit 937008c. Correctly preserves exit code, full untruncated stdout, and stderr in FTS5 index.

### Findings

**Finding 14 (Informational, Low) — Soft-fail FTS5 index format inconsistency.** When classifyNonZeroExit returns soft-fail (e.g., grep exit 1), the raw path wraps content with `Exit code:` prefix that the non-raw path omits. Cosmetic only; accept as-is.

**Finding 15 (Informational, Low) — No test for error-path reconstruction.** The `Exit code: N\n\nstdout:\n...\n\nstderr:\n...` format introduced in 937008c has no dedicated test. Low risk since it's simple string interpolation.

**Finding 16 (Informational, Low) — Timeout path doesn't use raw file.** Pre-existing behavior, not introduced by this PR.

### Overall Recommendation

**APPROVE.** All prior fixes confirmed. Three new informational findings, none blocking. The implementation is solid across all code paths.

rich-jojo:
@halindrome Please follow the PR template

mksglu:
Hey @halindrome! 👋

Thanks for tackling this — the data-loss problem in #197 is real and your analysis was spot on. The dual-path architecture (temp file for raw + truncated for agent) was well-designed.

We solved it differently though: instead of adding temp files to preserve raw content alongside truncated output, we removed `smartTruncate` from the executor entirely — both stdout and stderr. The server now auto-indexes large outputs (>100KB) into FTS5 and returns only a pointer message. This eliminates:

- Temp file writes/reads/cleanup
- Dual return (raw + truncated)
- Orphan sweep overhead
- Memory doubling from `readFileSync`
- The `smartTruncate` function and `maxOutputBytes` option (fully removed from codebase)

The fix is simpler and addresses the root cause (why truncate data we want to index?) rather than working around it.

Landed on `next` in [`ee628da`](https://github.com/mksglu/context-mode/commit/ee628da) and [`a6a1fc0`](https://github.com/mksglu/context-mode/commit/a6a1fc0).

Appreciate the detailed investigation and the clean PR — it helped us understand the problem deeply! 🙏

PR #205, opened 2026-03-31
feat: add support for JSONC configuration files and update related error messages

FloSchl8:
## What

OpenCode supports both `opencode.json` and `opencode.jsonc` as config formats ([Docs](https://opencode.ai/docs/config/)). This PR adds JSONC support to the OpenCode adapter so context-mode can read the config regardless of which format the user has chosen.

## Why

When a user configures OpenCode using `opencode.jsonc`, context-mode fails to detect the plugin configuration:

```
■  Plugin configuration: FAIL — Could not read opencode.json
│    Run: context-mode upgrade
```

The adapter only looked for `opencode.json` and never attempted `opencode.jsonc`, causing a false-negative on an otherwise valid setup.

## How

Updated the OpenCode adapter's config detection logic to check for both `opencode.json` and `opencode.jsonc` (in that order). If the `.json` file is not found, it falls back to `.jsonc`. JSONC is a superset of JSON and Node's `JSON.parse` handles it cleanly after stripping comments - or the existing config parser already handles it if comments are absent.

## Affected platforms

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] VS Code Copilot
- [x] OpenCode
- [ ] Codex CLI

### RED (failing test)

Sry no strict TDD here. Tests were still successful afterwards.

## Adapter checklist

- [ ] Hook scripts work on all affected platforms
- [ ] Routing instruction files updated if tool names or behavior changed

## Test plan

- [ ] `npm test` passes (631+ tests)
- [x] `context-mode doctor` — all checks PASS on my local build
- [x] Tested in a live session with my local MCP server

### Before/After comparison

**Before:** `■  Plugin configuration: FAIL — Could not read opencode.json`

**After:** `■  Plugin configuration: OK — opencode.jsonc detected`

## Local development setup

## Checklist

- [ ] I've compared output quality before and after my change

### DIFF
--- a/package.json
+++ b/package.json
@@ -81,6 +81,7 @@
     "@clack/prompts": "^1.0.1",
     "@mixmark-io/domino": "^2.2.0",
     "@modelcontextprotocol/sdk": "^1.26.0",
+    "jsonc-parser": "^3.3.1",
     "picocolors": "^1.1.1",
     "turndown": "^7.2.0",
     "turndown-plugin-gfm": "^1.0.2",
--- a/src/adapters/opencode/index.ts
+++ b/src/adapters/opencode/index.ts
@@ -27,6 +27,7 @@ import {
 } from "node:fs";
 import { resolve, join } from "node:path";
 import { homedir } from "node:os";
+import { parse as jsoncParse } from "jsonc-parser";
 
 import type {
   HookAdapter,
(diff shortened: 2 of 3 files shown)


mksglu:
Hey @FloSchl8, thanks for putting this together! 👋

Great catch on the JSONC support gap — OpenCode does indeed support both formats and we should too.

I've implemented this directly on `next` with a few adjustments:

**What changed vs your PR:**

1. **No `jsonc-parser` dependency** — replaced with a 3-line `stripJsonComments` function (~180 bytes vs 208KB package). We keep context-mode's dependency tree lean.

2. **Path priority fixed** — `.json` is checked before `.jsonc` at each precedence level (project → .opencode/ → global). Your PR had `.jsonc` first, which would break users with both files.

3. **TDD coverage** — added 4 tests:
   - Reads `.jsonc` with comments stripped
   - Prefers `.json` over `.jsonc` when both exist
   - `configureAllHooks` works with `.jsonc` (write-back to same file)
   - Error message mentions both formats

4. **Dynamic error messages** — `"Could not read opencode.json or opencode.jsonc"` (also works for KiloCode)

**Note on comments:** `writeSettings()` uses `JSON.stringify`, so user comments in `.jsonc` files will be stripped on write-back (e.g., after `configureAllHooks`). This is a known limitation — preserving comments would require AST-aware editing which adds significant complexity. For now this is acceptable since config writes are rare (only during `ctx upgrade`).

Thanks for identifying this issue — it helped improve context-mode for OpenCode users! 🙏

FloSchl8:
Hey @mksglu 

yeah it was really bothering me ending up with an empty config file except the context-mode plugin und ctx doctor afterwards saying it can"t find a config file...

Sure JS/TS isn't my first language so no worries.

PR #209, opened 2026-04-01
fix: use real OS temp dir for shell TMPDIR instead of project root

vytautas-petrikas:
## Summary

- `os.tmpdir()` reads the `TMPDIR` env var, which may already be set to the project root by the user's shell. This caused shell commands to write temp/cache dirs (`node-compile-cache/`, `nx-native-file-cache-*/`, `tsx-*/`) into the working tree.
- Resolve the real OS temp dir via `getconf DARWIN_USER_TEMP_DIR` on macOS (`/tmp` fallback on Linux), bypassing `TMPDIR` env var.
- Thread a separate `sandboxTmpDir` parameter through `#spawn` so shell commands get the correct `TMPDIR` even though their `cwd` is the project root.

## Test plan

- [x] Added test: `shell TMPDIR points to OS temp dir, not project root`
- [x] All existing tests pass (no regressions)
- [x] Manual: run a shell command via context-mode and verify `$TMPDIR` points to `/var/folders/.../T/.ctx-mode-*`, not the project dir

### DIFF
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -14,6 +14,21 @@ import type { ExecResult } from "./types.js";
 
 const isWin = process.platform === "win32";
 
+// os.tmpdir() reads TMPDIR from the environment, which may be set to the
+// project root by the user's shell. Resolve the real OS temp dir by reading
+// the original TMPDIR before any override, falling back to /tmp on Unix and
+// the process.env.TEMP/TMP on Windows.
+const OS_TMPDIR = (() => {
+  if (isWin) return process.env.TEMP ?? process.env.TMP ?? tmpdir();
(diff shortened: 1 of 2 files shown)


mksglu:
Thanks @vytautas-petrikas — solid catch. The bug was real: `#buildSafeEnv(cwd)` was passing `projectRoot` as TMPDIR for shell commands, causing node/nx/tsx temp files to pollute the working tree.

We've applied your fix to `next` with two improvements:
- **`execFileSync` instead of `execSync`** — avoids shell interpretation at startup (no shell injection vector)
- **`mktemp -u -d` instead of `dirname $(mktemp -u)`** — cleaner, one fewer subshell

Both your tests plus a cross-platform writability test are included. Co-authored-by preserved. Appreciate the contribution 🙏

PR #210, opened 2026-04-01
Refactor/opencode improvements

### LINKED ISSUE #208: [Discussion]: CLI upgrade full support for Opencode/Kilocode
### Platform

Opencode / Kilocode

### context-mode version

1.0.54

### Full doctor output (REQUIRED)

```text
context-mode diagnostics
 Runtimes: 7/11 (64%) — javascript, shell, typescript, python, ruby, go, perl
 Performance: FAST (Bun)
 Server test: PASS
 FTS5 / SQLite: PASS — native module works
 Hook script: PASS — /Users/mickey/.cache/kilo/node_modules/context-mode/hooks/pretooluse.mjs
 Version: v1.0.54
All systems operational.
```

Currently, I am working on following:

### Problem 1: Hardcoded platform references found

Location: src/adapters/opencode/index.ts:357, 381, 403, 423
Problem: Error messages hardcoded "opencode.json", breaking support for other platforms like kilo.
Solution: Use this.platform and this.name for dynamic references.

### Problem 2: Config file selection logic

Problem: readSettings() stopped at the first valid config file, potentially missing configs with the context-mode plugin registered.
- I am using global config file to register context-mode plugin. In some projects I also have local config which could be override or extension of global file settings. Doctor and upgrade CLI options in this case don't check global file and plugin array has been reported as missing or added to project file even I have it registered globally. 

**Solution:** Scan all config paths, prioritize configs with context-mode plugin or global config, fallback to first valid config.

### Problem 3: Incorrect plugin root path

Problem: CLI uses project folder as root, while Opencode/Kilocode install from npm to user's global .cache directory and upgrade should install there latest downloaded version of plugin.
Solution: Added cachePluginRoot() for opencode-based platforms, fallback to defaultPluginRoot().

### Problem 4: Unnecessary operations during upgrade

Problem: Rebuilding native addons and running doctor on prebuilt npm packages; unnecessary for opencode/kilo.
Solution: Skip rebuild/doctor for opencode/kilo platforms; early return if already on latest version.

### Problem 5: Best backup strategy for config modification

Problem: CLI backups and updates wrong config file. Also, it overwrites old backup file even there is no update to config present.
Solution: this is tricky one. First I turned off this step in CLI for Opencode/Kilocode and moved calling to adapter's backup function inside of adapter alone. Backup is done only if setting file is going to change but with this solution CLI show's nothing (silent mode). To solve this silent mode I need to return somehow these information to main CLI process. I would do that but the main strategic decision needs to be made first. And this approach could be ignored depending on that decision.

## Opencode/Kilocode Strategy

- we need to decide how CLI will treat Opencode/Kilocode platform. I think we need to ask users where they want plugin to be installed. Locally or globally? This way CLI should know where to look exactly for context-mode configuration depending on users need.
- another dilemma I have is should we use `npm` plugin mode at all. CLI also allow us to install and update plugin through plugin folder either globally either locally. It is anyway clone repository in tmp, build it and install it overwriting the old version installed by `npm`. 
- we should not need CLI at all. According to this [PR](https://github.com/anomalyco/opencode/pull/16998), plugin array should contain "contex-mode@latest" which should update automatically whenever new version of `npm` package is released. **Note:** I tested this in Kilocode and it does not work yet, so at least for now CLI is still only option for easy update. **Update**: this [PR](https://github.com/Kilo-Org/kilocode/pull/8028) should bring fix from Opencode repository into Kilocode

BTW I have fixes almost ready for everything here, so this issue is not invite for the implementation. I just need clear requirements to know how to finish my contribution.

mikij:
## What

This commit refactors the settings reading logic in the OpenCode adapter to prioritize configs with context-mode plugins and adds proper cache-based plugin root detection for Opencode/Kilocode platforms. Messages have been updated to use dynamic platform names instead of hardcoded references.

## Why

Closes #208 

## Implementation Approach

**What changed:**

1. **Settings reading priority** (`src/adapters/opencode/index.ts:300-326`)
   - Refactored to prioritize configs with context-mode plugin over first valid config
   - Only falls back to global config if no plugin-enabled config exists

2. **Platform-aware plugin root detection** (`src/cli.ts:119-135`)
   - Added `cachePluginRoot()` for Opencode/Kilocode's npm cache installation path
   - Added platform detection to route to appropriate root directory

3. **Dynamic platform naming** (`src/adapters/opencode/index.ts:357,382,391,403,422`)
   - Replaced hardcoded "opencode.json"/"OpenCode" with `this.platform`/`this.name`

**Why:**

- Plugin prioritization ensures correct config is used for registration checks
- Cache-based detection fixes upgrade flows for Opencode/Kilocode platforms that install plugins in `.cache`
- Dynamic naming enables single adapter to support both platforms

## Affected platforms

- [ ] Claude Code
- [ ] Gemini CLI
- [ ] VS Code Copilot
- [x] OpenCode / KiloCode
- [ ] Codex CLI

Doctor response after change:
```
┌   context-mode doctor 
│
●  Platform: KiloCode (high confidence — KILO or KILO_PID env var set)
│
◇  Diagnostics complete
│
◇  Runtimes ─────────────────────────────────────────────────────────────────────────────────────────────────────╮
│                                                                                                                │
│    JavaScript: bun (1.3.11) ⚡                                                                                 │
│    TypeScript: bun (1.3.11)                                                                                    │
│    Python:     python3 (Python 3.9.6)                                                                          │
│    Shell:      bash (GNU bash, version 3.2.57(1)-release (arm64-apple-darwin25))                               │
│    Ruby:       ruby (ruby 2.6.10p210 (2022-04-12 revision 67958) [universal.arm64e-darwin25])                  │
│    Go:         go (unknown)                                                                                    │
│    Perl:       perl (This is perl 5, version 34, subversion 1 (v5.34.1) built for darwin-thread-multi-2level)  │
│                                                                                                                │
├────────────────────────────────────────────────────────────────────────────────────────────────────────────────╯
│
◆  Performance: FAST — Bun detected for JS/TS execution
│
●  Language coverage: 7/11 (64%) — javascript, shell, typescript, python, ruby, go, perl
│
◇  Testing server initialization...
│
◆  Server test: PASS
│
◇  Checking KiloCode hooks configuration...
│
◆  Plugin registration: PASS — context-mode found in plugin array
│
■  SessionStart hook: FAIL — SessionStart not supported in KiloCode (see issues #14808, #5409)
│
◇  Checking hook script...
│
◆  Hook script exists: PASS — /Users/mickey/.cache/kilo/node_modules/context-mode/hooks/pretooluse.mjs
│
◇  Checking KiloCode plugin registration...
│
◆  Plugin enabled: PASS — context-mode found in plugin array
│
◇  Checking FTS5 / SQLite...
│
◆  FTS5 / SQLite: PASS — native module works
│
◇  Checking versions...
│
◆  npm (MCP): PASS — v1.0.54
│
◆  KiloCode: PASS — v1.0.54
│
└  Diagnostics complete!
```


mikij:
@mksglu some tests seems unstable. I got this then run again and everything went well. Then I notice something similar in CI / test for linux. Though it is different hook

image

mksglu:
> @mksglu some tests seems unstable. I got this then run again and everything went well. Then I notice something similar in CI / test for linux. Though it is different hook
> 
> image

CI is OK now

mksglu:
@mikij You're doing great job!

mikij:
> @mikij You're doing great job!

Thanks

mksglu:
@mikij I added a SKILL. 

Use it, please.

```
npx skills add mksglu/context-mode --skill context-mode-ops
```

Use that SKILL and review your PR within `ultrathink` keyword on Claude Code.

mikij:
> @mikij I added a SKILL.
> 
> Use it, please.
> 
> ```
> npx skills add mksglu/context-mode --skill context-mode-ops
> ```
> 
> Use that SKILL and review your PR within `ultrathink` keyword on Claude Code.

I will try with KiloCode as I am not using Claude

mikij:
@mksglu Not sure but you can add coderabbitai. It has free tier for open source projects on GH

mikij:
@mksglu look it happened again

mikij:
@mksglu review discovered two issues. one was addressed in this [commit](https://github.com/mksglu/context-mode/pull/210/commits/53c7f980cfcfa560a5e3992ded08ac50e6ff7203). the second discovered missing parallel for `process.env.OPENCODE_PROJECT_DIR`, but in OpenCode [documentation](https://opencode.ai/docs/cli/#environment-variables) I could not find that variable is defined. Should I remove it and use just `process.cwd()`?

Kilocode repo search shows no trace of it also:
image

mksglu (commented):
Thanks for this, @mikij! Solid work overall. A few minor things I noticed:

**1. Unused import: `realpathSync`**
`src/adapters/opencode/index.ts` imports `realpathSync` from `node:fs` but it's never used anywhere in the file. Dead import, should be removed.

**2. `isGlobalConfig` positional detection is fragile**
The `configPath === configPaths.at(-1) || configPath === configPaths.at(-2)` pattern relies on global paths always being the last two entries. If someone adds or reorders paths in `paths()`, this silently breaks. A `Set` of global paths or an explicit `isGlobalPath()` helper would be more robust.

**3. `.jsonc` paths only for Kilo, not OpenCode**
Kilo gets `kilo.jsonc` and `.kilo/kilo.jsonc`, but OpenCode's `paths()` doesn't include `opencode.jsonc` variants. If JSONC support is being added, it should be consistent across both platforms.

mksglu:
@mikij RE: the environment variable documentation concern.

The most reliable and up-to-date documentation is the source code itself. If OpenCode/KiloCode have added `XDG_CACHE_HOME` handling or `.kilo/` paths in their codebase and it works in practice, that's our reference. They may not have updated their docs yet, but the code is the contract.

Let's write our adapter to match what their code actually does. If it's in their source and it runs, we treat it as the spec.

mikij:
@mksglu  .kilo is new standard for OpenCode based KiloCode implementation. .kilocode is old directory name and it is standard for RooCode based VSCode extension. I found that in their discord. I tested .kilo and it works and renamed all .kilocode folders to .kilo in my projects. I am not waiting they announce for deprecation of .kilocode folder. That is why I changed this in the PR too.

About XDG_CACHE_HOME that is the standard I think we can cover in our code. I see in KiloCode it is used only in some test env settings. Similar finding in OpenCode repo. It seems they don't using it so I will remove even I don't think that could cause problems.

image

mksglu:
@mikij Once you've read through all the review comments and replies:

If the changes are proven and tested (which they appear to be based on CI and our review), no issue there. Just make sure you update the README and docs/platform-support.md to reflect any new config paths, JSONC support, or KiloCode changes you've introduced.

Docs ship with code, not after.

mikij:
> @mikij Once you've read through all the review comments and replies:
> 
> If the changes are proven and tested (which they appear to be based on CI and our review), no issue there. Just make sure you update the README and docs/platform-support.md to reflect any new config paths, JSONC support, or KiloCode changes you've introduced.
> 
> Docs ship with code, not after.

Ofc I am taking care about that. I knew for README and I am checking it. But good to know that there is some another too

mksglu:
@mikij Code duplication review. There are several repeated patterns that should be cleaned up:

**1. XDG_CONFIG_HOME / APPDATA resolution is copy-pasted 3 times**
Same 5-line block in `opencode/index.ts` getSessionDir(), `opencode-plugin.ts` getSessionDir(), and tests. Extract a shared `getConfigDir()` utility.

**2. `getSessionDir()` duplicated between adapter and plugin**
`opencode-plugin.ts:81-91` reimplements what `OpenCodeAdapter.getSessionDir()` already does. The plugin imports the adapter at line 29, just call `adapter.getSessionDir()` instead.

**3. `getDBPath()` same story**
`opencode-plugin.ts:93-99` duplicates `OpenCodeAdapter.getSessionDBPath()`. Reuse the adapter method.

**4. Cache path logic duplicated between cli.ts and adapter**
`cachePluginRoot()` in cli.ts has Windows LOCALAPPDATA support, but `getInstalledVersion()` in the adapter hardcodes `~/.cache` without Windows handling. Extract a shared `getCachePluginDir(platform)` so both use the same logic (and the adapter gets the Windows fix for free).

**5. Plugin-check predicate repeated 3 times**
`validateHooks()`, `checkPluginRegistration()`, and `readSettings()` all independently check `plugins.some(p => p.includes("context-mode"))`. Extract a `hasContextMode(settings)` helper.

**6. Test scaffolding boilerplate (5+ copies)**
The tmpdir + mkdirSync + path resolution block is copy-pasted across multiple test cases. A `createTestFixture(platform)` helper would clean this up.

**7. Remove unused `realpathSync` import**

These are suggestions for a follow-up cleanup. The core logic is solid.

mksglu:
Merged to `next` — thanks @mikij! Solid work on the settings priority, cache detection, and .kilo migration.

The code duplication suggestions I noted are follow-up work, not blocking. I'll handle the docs update on my side.

Closes #208. 🙏

mikij:
@mksglu anyway I am working to remove these duplications, cleaner code is always better to have

mksglu:
@mikij All 7 code review items are now fixed on `main` + `next`:

1. ✅ Removed unused `realpathSync` import
2. ✅ Replaced fragile `isGlobalConfig` (.at(-1)/.at(-2)) with Set-based membership check
3. ✅ JSONC verified for both OpenCode and Kilo (your fix was correct)
4. ✅ Deduplicated `getSessionDir()` in opencode-plugin — now delegates to adapter
5. ✅ Deduplicated `getDBPath()` in opencode-plugin — now delegates to adapter
6. ✅ Extracted `hasContextModePlugin()` helper — replaces 3 duplicated checks
7. ✅ Cleaned up unused imports in opencode-plugin.ts

1295/1295 tests pass. Cutting a release soon — please pull latest `next`, test on your KiloCode setup, and confirm everything works. Tracking and testing is your responsibility going forward.

PR #222, opened 2026-04-02
Fix: Support openclaw.json in OpenClaw installer

vicpon:
## Problem

The OpenClaw installer script (`scripts/install-openclaw-plugin.sh`) fails because it expects `runtime/openclaw.runtime.json` but OpenClaw v2026.3.28 only creates `openclaw.json` as its configuration file.

## Solution

Modified the preflight check to support both:
1. `runtime/openclaw.runtime.json` (existing behavior)
2. `openclaw.json` (OpenClaw's current default)

## Changes

- Updated `scripts/install-openclaw-plugin.sh` to check for both file locations
- Added a fallback message when `openclaw.json` is used
- Maintained backward compatibility with existing installs

## Testing

Verified the fix works with:
- OpenClaw v2026.3.28
- Node.js v24.14.0
- Docker container environment

## Notes

The fix is minimal and only affects the installer script's file detection logic. No other functionality is changed.

### DIFF
--- a/scripts/install-openclaw-plugin.sh
+++ b/scripts/install-openclaw-plugin.sh
@@ -17,14 +17,20 @@ if [ ! -d "$OPENCLAW_STATE_DIR" ]; then
   echo "✗ OPENCLAW_STATE_DIR ($OPENCLAW_STATE_DIR) does not exist. Is OpenClaw installed?" >&2
   exit 1
 fi
-if [ ! -f "$OPENCLAW_STATE_DIR/runtime/openclaw.runtime.json" ]; then
-  echo "✗ $OPENCLAW_STATE_DIR/runtime/openclaw.runtime.json not found." >&2
+
+# Support both runtime/openclaw.runtime.json and openclaw.json
+if [ -f "$OPENCLAW_STATE_DIR/runtime/openclaw.runtime.json" ]; then
+  RUNTIME_JSON="$OPENCLAW_STATE_DIR/runtime/openclaw.runtime.json"
+elif [ -f "$OPENCLAW_STATE_DIR/openclaw.json" ]; then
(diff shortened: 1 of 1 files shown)


mksglu:
Thanks @vicpon — the fix looks reasonable but I need to verify one thing before merging.

Our codebase uses two different OpenClaw config files for two different purposes:

| File | Purpose | Used by |
|------|---------|---------|
| `openclaw.json` | User config (plugin list, settings) | `src/adapters/openclaw/index.ts` |
| `runtime/openclaw.runtime.json` | Gateway runtime config (install script writes plugin registration here) | `scripts/install-openclaw-plugin.sh` |

The install script writes `plugins.entries["context-mode"]` and `plugins.allow` into `RUNTIME_JSON` — then sends SIGUSR1 to restart the gateway.

**Question**: In your OpenClaw v2026.3.28 environment, does writing plugin entries to `openclaw.json` (instead of `runtime/openclaw.runtime.json`) actually get picked up by the gateway after SIGUSR1? In other words — does the gateway read `openclaw.json` for plugin configuration in v2026.3.28?

If yes, the fix is correct and we'll merge. If the gateway still needs `runtime/openclaw.runtime.json` for plugin registration, then the fallback would silently break the install.

Can you test: after running the install script with your fix, does `ctx doctor` show the plugin as registered?

mksglu:
Thanks @vicpon — merged to `next`. You were right: OpenClaw uses `openclaw.json` as the single config file ([docs.openclaw.ai/tools/plugin](https://docs.openclaw.ai/tools/plugin)). `runtime/openclaw.runtime.json` is a legacy path that no longer exists in v2026.3.28+.

I'm pushing a follow-up fix on top to remove the fallback entirely — single source of truth, no backwards compat. The script will use `openclaw.json` directly. 🙏

PR #235, opened 2026-04-07
fix: emit empty hookSpecificOutput in PostToolUse to prevent false hook error

### LINKED ISSUE #229: PostToolUse hook matcher "" fires on ALL tools (TaskUpdate, TaskCreate, etc.) causing false "hook error" display
## Bug

The context-mode plugin registers a PostToolUse hook in `hooks/hooks.json` with `matcher: ""` (empty string), which fires on **every** PostToolUse event — including `TaskUpdate`, `TaskCreate`, `TaskGet`, `TaskList`, and other internal tools that `posttooluse.mjs` was not designed to handle.

## Symptom

Users see `PostToolUse:TaskUpdate hook error` on every `TaskUpdate` call, even when the hook exits 0.

This is consistent with known Claude Code bugs:
- anthropics/claude-code#41868 — exit 0 with valid hookSpecificOutput shown as "hook error"
- anthropics/claude-code#35587 — plugin hooks firing on tools they were not designed for causes false "hook error" display

## Suggested Fix

Narrow the PostToolUse matcher to exclude internal task management tools:

```json
"PostToolUse": [
  {
    "matcher": "Bash|Read|Write|Edit|Grep|Agent|Task|mcp__plugin_context-mode",
    "hooks": [
      {
        "type": "command",
        "command": "node ${CLAUDE_PLUGIN_ROOT}/hooks/posttooluse.mjs"
      }
    ]
  }
]
```

## Environment

- context-mode: v1.0.75 (plugin install via Claude Code)
- Claude Code: latest
- macOS Darwin 25.5.0, arm64

murataslan1:
## Problem

PostToolUse hook exits 0 but outputs nothing to stdout. Claude Code interprets empty stdout as a hook failure and displays "hook error" on every tool call — including `TaskUpdate`, `TaskCreate`, and other internal tools that fire due to the `matcher: ""` catch-all.

References: anthropics/claude-code#41868, anthropics/claude-code#35587

## Fix

Output empty `{"hookSpecificOutput":{}}` at the end of `posttooluse.mjs`. This tells Claude Code the hook completed successfully without any modifications.

## Test results

| Tool | Before (exit/stdout) | After (exit/stdout) | Status |
|------|---------------------|---------------------|--------|
| TaskUpdate | 0 / (empty) | 0 / `{"hookSpecificOutput":{}}` | FIXED |
| Bash | 0 / (empty) | 0 / `{"hookSpecificOutput":{}}` | FIXED |
| Agent | 0 / (empty) | 0 / `{"hookSpecificOutput":{}}` | FIXED |
| Read | 0 / (empty) | 0 / `{"hookSpecificOutput":{}}` | FIXED |
| {} (invalid) | 0 / (empty) | 0 / `{"hookSpecificOutput":{}}` | FIXED |

PreToolUse regression check: no change (still returns additionalContext as before).

Full test reports: https://gist.github.com/murataslan1/e0290cb0862f57fff4a0e2936b6a77d8

Closes #229

### DIFF
--- a/hooks/posttooluse.mjs
+++ b/hooks/posttooluse.mjs
@@ -53,4 +53,5 @@ try {
   // PostToolUse must never block the session — silent fallback
 }
 
-// PostToolUse hooks don't need hookSpecificOutput
+// Emit empty hookSpecificOutput so Claude Code doesn't show false "hook error" (CC #41868).
+process.stdout.write(JSON.stringify({ hookSpecificOutput: {} }) + "\n");


mksglu:
Thank you for the detailed investigation and testing @murataslan1 — your analysis of the empty stdout → "hook error" pattern was spot-on and helped inform the final fix.

However, we took a different approach that addresses the root cause rather than the symptom:

**What shipped on `next` (commit `ece3abb`):**
- PostToolUse `matcher: ""` → narrowed to 17 tools that `extractEvents()` actually handles
- 13 tools that produced zero events (`TaskGet`, `TaskList`, `TaskStop`, etc.) no longer fire the hook at all
- `"Task"` removed from PreToolUse matchers (#241 — you flagged this connection in your issue comment!)
- `.claude-plugin/hooks/hooks.json` duplicate deleted

**Why this approach vs empty hookSpecificOutput:**
- Your fix would still fire the hook for every tool call (process spawn + SQLite open/close for nothing)
- Our fix prevents the hook from firing at all for tools it doesn't handle
- Zero wasted process spawns = better performance

Your comment on #229 about #241's substring matching was directly used in the fix. Credit noted in the commit.

Closing as superseded — #229 is already resolved.

PR #237, opened 2026-04-07
fix: remove stdin listeners from lifecycle guard to prevent MCP -32000 errors

### LINKED ISSUE #236: Lifecycle guard stdin listeners cause spurious MCP -32000 Connection closed
## Bug Description

The lifecycle guard (`src/lifecycle.ts`) added in #103 to prevent orphaned processes registers `end`/`close`/`error` listeners on `process.stdin` and calls `process.stdin.resume()`. These conflict with the MCP SDK's `StdioServerTransport`, which also reads from `process.stdin` via `readline`. When stdin emits a transient event, the lifecycle guard interprets it as parent death and calls `process.exit(0)` immediately — killing the server mid-request.

The client (Claude Code) then receives:
```
Error: MCP error -32000: Connection closed
```

This is intermittent and difficult to reproduce deterministically, but occurs frequently during normal sessions — often multiple times per hour.

## Root Cause

In `src/lifecycle.ts:57-63`:
```typescript
const onStdinClose = () => shutdown();
process.stdin.resume();
process.stdin.on("end", onStdinClose);
process.stdin.on("close", onStdinClose);
process.stdin.on("error", onStdinClose);
```

Two problems:

1. **`process.stdin.resume()`** puts stdin into flowing mode before `StdioServerTransport` sets up its own `readline` interface (the guard starts at `server.ts:2103`, transport at `server.ts:2105`). This can cause data loss or buffering issues during the transport handshake.

2. **`end`/`close`/`error` on stdin** — any transient pipe event triggers an immediate `gracefulShutdown()` → `process.exit(0)` with no debounce or grace period. In a stdio MCP setup, the transport legitimately manages stdin flow; transient events do not necessarily indicate parent death.

The `gracefulShutdown()` function (`server.ts:2094-2097`) also does not `await transport.close()` before exiting, so any in-flight JSON-RPC request is abruptly terminated.

## Suggested Fix

Remove the stdin-based shutdown detection from the lifecycle guard entirely. The remaining mechanisms are sufficient:

- **ppid polling** (every 30s) — reliably detects parent death via reparenting to init/systemd/launchd
- **SIGTERM/SIGINT/SIGHUP** signal handlers — catch intentional shutdown

The stdin listeners are redundant in an MCP stdio context because the transport already handles stdin lifecycle, and they're harmful because they create false-positive shutdowns.

### Minimal diff

```diff
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -54,13 +54,6 @@ export function startLifecycleGuard(opts: LifecycleGuardOptions): () => void {
   }, interval);
   timer.unref();

-  // P0: Stdin close — parent pipe broken
-  // Must resume stdin to receive close/end events (Node starts paused)
-  const onStdinClose = () => shutdown();
-  process.stdin.resume();
-  process.stdin.on("end", onStdinClose);
-  process.stdin.on("close", onStdinClose);
-  process.stdin.on("error", onStdinClose);
-
   // P0: OS signals — terminal close, kill, ctrl+c
   const signals: NodeJS.Signals[] = ["SIGTERM", "SIGINT"];
@@ -70,9 +63,6 @@ export function startLifecycleGuard(opts: LifecycleGuardOptions): () => void {
   return () => {
     stopped = true;
     clearInterval(timer);
-    process.stdin.removeListener("end", onStdinClose);
-    process.stdin.removeListener("close", onStdinClose);
-    process.stdin.removeListener("error", onStdinClose);
     for (const sig of signals) process.removeListener(sig, shutdown);
   };
 }
```

## Environment

- context-mode: v1.0.75
- @modelcontextprotocol/sdk: ^1.26.0 (resolved to 1.29.0)
- Node.js: v22.22.1
- Platform: Linux 6.17.0
- Client: Claude Code (CLI)

## Workaround

Manually patch `server.bundle.mjs` in the plugin cache to remove the stdin listeners from the minified lifecycle guard function (`Nx`). The patch survives session restarts but is overwritten on plugin updates.

ponythewhite:
## Summary

- Removes `process.stdin.resume()` and `end`/`close`/`error` listeners from the lifecycle guard (`src/lifecycle.ts`)
- These listeners conflict with MCP SDK's `StdioServerTransport`, which also reads from `process.stdin` via `readline`
- Transient stdin events cause false-positive parent-death detection → immediate `process.exit(0)` → client receives `MCP error -32000: Connection closed`

## Root Cause

The lifecycle guard (added in #103) registers stdin listeners to detect parent death. However, in an MCP stdio transport, `StdioServerTransport` is the rightful owner of `process.stdin`. Two consumers competing for the same stream causes:

1. **`process.stdin.resume()`** puts stdin into flowing mode before the transport is ready, risking data loss during handshake
2. **`end`/`close`/`error` listeners** fire on transient pipe events that don't indicate actual parent death, triggering immediate shutdown with no grace period

## What's Preserved

Parent death detection remains robust via:
- **ppid polling** (every 30s) — detects reparenting to init/systemd/launchd
- **SIGTERM/SIGINT/SIGHUP** signal handlers — catch intentional shutdown

## Test Plan

- [ ] Verify MCP server starts and responds to tool calls normally
- [ ] Verify server exits when parent Claude Code session ends (ppid check + signals)
- [ ] Confirm no more spurious `-32000: Connection closed` errors during extended sessions
- [ ] Run existing test suite if available

Fixes #236

### DIFF
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -54,13 +54,12 @@ export function startLifecycleGuard(opts: LifecycleGuardOptions): () => void {
   }, interval);
   timer.unref();
 
-  // P0: Stdin close — parent pipe broken
-  // Must resume stdin to receive close/end events (Node starts paused)
-  const onStdinClose = () => shutdown();
-  process.stdin.resume();
-  process.stdin.on("end", onStdinClose);
-  process.stdin.on("close", onStdinClose);
-  process.stdin.on("error", onStdinClose);
+  // NOTE: Stdin-based shutdown detection (process.stdin.resume + end/close/error
+  // listeners) was removed here. It conflicts with the MCP SDK's StdioServerTransport
+  // which also reads from process.stdin via readline. Transient stdin events caused
+  // false-positive shutdowns → "MCP error -32000: Connection closed" on the client.
+  // Parent death is reliably caught by the ppid check above + OS signals below.
+  // See: https://github.com/mksglu/context-mode/issues/236
 
   // P0: OS signals — terminal close, kill, ctrl+c
   const signals: NodeJS.Signals[] = ["SIGTERM", "SIGINT"];
@@ -70,9 +69,6 @@ export function startLifecycleGuard(opts: LifecycleGuardOptions): () => void {
   return () => {
     stopped = true;
     clearInterval(timer);
-    process.stdin.removeListener("end", onStdinClose);
-    process.stdin.removeListener("close", onStdinClose);
-    process.stdin.removeListener("error", onStdinClose);
     for (const sig of signals) process.removeListener(sig, shutdown);
   };
 }


mksglu:
CI is throw error. @ponythewhite

flightlesstux:
Opened a replacement PR that keeps this diff verbatim and also updates `tests/lifecycle.test.ts` so CI goes green: https://github.com/mksglu/context-mode/pull/255. Credit preserved via Co-authored-by. If this one stays idle, the other carries the same fix forward.

mksglu:
Hey @ponythewhite — thank you for this! You nailed the root cause: stdin listeners in the lifecycle guard conflict with MCP SDK's StdioServerTransport, causing spurious -32000 errors.

We shipped the same fix on `next` in two commits:

1. [`156a1cd`](https://github.com/mksglu/context-mode/commit/156a1cd) — removed stdin listeners from lifecycle guard, matching your approach
2. [`ae73412`](https://github.com/mksglu/context-mode/commit/ae73412) — added regression test guards to ensure stdin listeners never come back

The lifecycle guard now uses **only** ppid polling + OS signals (SIGTERM/SIGINT/SIGHUP), exactly as you proposed. `lifecycle.ts:8` explicitly documents:

> "Stdin close is NOT used as a shutdown signal — the MCP stdio transport owns stdin and transient pipe events cause spurious -32000 errors (#236)"

Closing since this is already on `next`. Your investigation saved us significant debugging time — the -32000 error is notoriously hard to trace back to stdin listener conflicts. 🙏

Issue #228, opened 2026-04-05
Support alternative SQLite backends (node:sqlite) to avoid native addon SIGSEGV crashes


Alexey-Goru1ev:
## Problem

`context-mode` crashes intermittently with **SIGSEGV (Signal 11)** when running on **Node.js v22/v24** on Linux due to a V8 runtime bug that corrupts `better-sqlite3`'s native addon `.got.plt` section.

### Root cause

V8's garbage collector calls `madvise(MADV_DONTNEED)` on memory ranges that can overlap `better_sqlite3.node`'s writable ELF segments (`.got.plt`). This discards resolved symbol addresses on Copy-on-Write pages, restoring unrelocated on-disk offsets — causing function calls to jump to near-null addresses.

**Upstream issue:** [nodejs/node#62515](https://github.com/nodejs/node/issues/62515) — *"Sporadic SIGSEGV: native addon .got.plt reset to unrelocated file offsets"*

### Why context-mode is especially affected

- Runs as an MCP server over **stdio pipes** (the documented trigger condition)
- Makes **frequent `better-sqlite3` calls** for FTS5 indexing and search
- High call frequency under sustained I/O maximizes exposure to the race condition

### Crash frequency

| Node Version | Crash Frequency | Tested On |
|-------------|----------------|-----------|
| v24.14.x | ~4 crashes/hour | Ubuntu 24.04, Linux 6.19.10, AMD x86-64 |
| v22.22.x | ~1 crash/2 hours | Same environment |

All coredumps show the same pattern: frame #0 at near-null address (0x12xxx range), consistent with `.got.plt` corruption.

### What doesn't help

| Attempted Fix | Result |
|--------------|--------|
| Downgrade Node v24 → v22 | Reduced frequency ~10x, did not eliminate |
| Upgrade better-sqlite3 to v12.8.0 (V8 HolderV2 fix) | No effect — crash persists |
| `LD_BIND_NOW=1` (eager symbol resolution) | Partial — changes memory layout only |
| `LD_PRELOAD` madvise guard (intercept+block on .node segments) | Effective but fragile — requires compiled C library, supervisor wrapper, auto-heal hook |

The only reliable fix is **eliminating the separate `.node` binary** that gets `dlopen()`'d as a standalone ELF file.

## Proposal: Support pluggable SQLite backends

The codebase already has the right abstraction for this:

[...]

PR #243, opened 2026-04-09
fix: wrap insertEvent with withRetry for concurrent PostToolUse hooks

james-gough-op:
## What / Why / How

**What:** Wrap `SessionDB.insertEvent()` with `withRetry()` and propagate `busy_timeout` pragma to Bun's SQLite adapter.

**Why:** When many tool calls complete in parallel (e.g., 16 batch `mcp__linear__get_issue` calls during a standup summary), Claude Code spawns multiple PostToolUse hooks simultaneously. These all open the same per-project SessionDB and compete for the SQLite write lock. While better-sqlite3's `busy_timeout` (30s) handles most contention, edge cases during transaction lock escalation can surface `SQLITE_BUSY`, causing Claude Code to report `PostToolUse:mcp__linear__get_issue hook error` to users.

Additionally, the `BunSQLiteAdapter` factory was silently dropping the `timeout` option passed from `SQLiteBase`, meaning Bun runtime had no `busy_timeout` at all — concurrent writes would fail immediately with `SQLITE_BUSY`.

**How:**
1. `src/session/db.ts` — Changed `transaction()` to `this.withRetry(() => transaction())` in `insertEvent`, leveraging the existing retry infrastructure in `SQLiteBase`
2. `src/db-base.ts` — Added `busy_timeout` pragma in `BunDatabaseFactory` when `opts.timeout` is provided, matching better-sqlite3's automatic behavior
3. Added a concurrent insert test verifying multiple DB instances can write to the same file without data loss

## Affected platforms

- [x] Claude Code
- [x] All platforms

_(The Bun `busy_timeout` fix applies to any platform using Bun runtime for the MCP server. The `withRetry` fix applies to all platforms since PostToolUse hooks run with Node.js.)_

## Test plan

- Added `Concurrent Insert Resilience` test in `tests/session/session-db.test.ts` — opens 5 `SessionDB` instances against the same DB file and inserts events from each, verifying all 5 events are stored without errors
- All 26 session-db tests pass
- `npm run typecheck` passes clean

## Checklist

- [x] `npm test` passes (6 pre-existing failures unrelated to this change — security, cursor-hooks, vscode-hooks, session-hooks-smoke, hooks/integration)
- [ ] Docs updated if needed (README, platform-support.md) — no docs changes needed

### DIFF
--- a/src/db-base.ts
+++ b/src/db-base.ts
@@ -127,7 +127,14 @@ export function loadDatabase(): typeof DatabaseConstructor {
           readonly: opts?.readonly,
           create: true,
         });
-        return new BunSQLiteAdapter(raw);
+        const adapter = new BunSQLiteAdapter(raw);
+        // Propagate busy_timeout to match better-sqlite3's timeout option.
+        // Without this, concurrent writes under Bun fail immediately with
+        // SQLITE_BUSY instead of retrying (better-sqlite3 sets this automatically).
+        if (opts?.timeout) {
+          adapter.pragma(`busy_timeout = ${Number(opts.timeout)}`);
+        }
+        return adapter;
       } as any;
     } else {
       // Node.js — use better-sqlite3.
--- a/src/session/db.ts
+++ b/src/session/db.ts
@@ -343,7 +343,12 @@ export class SessionDB extends SQLiteBase {
       this.stmt(S.updateMetaLastEvent).run(sessionId);
     });
 
-    transaction();
+    // Use withRetry to handle SQLITE_BUSY under concurrent PostToolUse hooks.
+    // When many tool calls complete in parallel (e.g., batch get_issue), multiple
+    // hook processes compete for the write lock on the same SessionDB file.
+    // better-sqlite3's busy_timeout handles most cases, but withRetry provides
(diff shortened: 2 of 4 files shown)


mksglu:
Hey @james-gough-op — great work here. Reviewed and merged onto `next`.

**What we validated:**
- `withRetry` is the correct pattern for transaction lock escalation deadlocks that `busy_timeout` alone can't handle (same pattern already used in `store.ts:609,766`)
- Bun `busy_timeout` propagation is a genuine gap — better-sqlite3 does it via constructor, bun:sqlite doesn't
- 2.6s max busy-wait is within Claude Code's 60s hook timeout budget
- Concurrent insert test passes with 5 DB instances against same file

**Note:** We resolved the merge conflict with `next` (we recently added a `NodeSQLiteAdapter` in db-base.ts for #228) and applied your changes on top. The Bun `busy_timeout` fix was placed correctly in the `BunDatabaseFactory` branch.

Commit: [`d5a5b47`](https://github.com/mksglu/context-mode/commit/d5a5b47) on `next`.

Thanks for the thorough analysis of the lock escalation scenario — the transaction retry approach is exactly right. 🙏

PR #253, opened 2026-04-12
fix: use relative path in .mcp.json to silence CLAUDE_PLUGIN_ROOT warning

### LINKED ISSUE #252: fix: .mcp.json warning - CLAUDE_PLUGIN_ROOT missing env variable
## Problem

Opening the repo in Claude Code triggers this warning:

```
[Warning] [context-mode] mcpServers.context-mode: Missing environment variables: CLAUDE_PLUGIN_ROOT
```

### Root cause

Repo root `.mcp.json`:

```json
{
  "mcpServers": {
    "context-mode": {
      "command": "node",
      "args": ["${CLAUDE_PLUGIN_ROOT}/start.mjs"]
    }
  }
}
```

`${CLAUDE_PLUGIN_ROOT}` is only set when Claude Code loads the repo as an **installed plugin**. When contributors open the repo directly for development, the env var is undefined, the path breaks, and Claude Code prints the warning on every session start.

### Why this matters

- Every developer who clones the repo sees the warning each session.
- Plugin-mode already has the correct MCP configuration in `.claude-plugin/plugin.json` (same `mcpServers` block).
- The repo-root `.mcp.json` seems intended for dev workflows (repo opened as a project, not installed as plugin), but the current value does not work there.

## Proposed fix

In `.mcp.json`, replace `${CLAUDE_PLUGIN_ROOT}/start.mjs` with a relative path `./start.mjs`. Claude Code spawns the MCP server with the project root as cwd, so Node resolves `./start.mjs` reliably. Plugin install path is unaffected (it still uses `.claude-plugin/plugin.json`).

## Alternative

Remove `.mcp.json` entirely. Plugin install uses `plugin.json`; dev-only usage can rely on `claude mcp add` (see the "MCP-only install" block in the README).

A PR with the relative-path fix will follow.

_(Replaces the mistakenly bot-opened #251.)_

flightlesstux:
## Summary

- Replace `${CLAUDE_PLUGIN_ROOT}/start.mjs` with `./start.mjs` in repo-root `.mcp.json`.
- Silences the warning Claude Code prints when the repo is opened as a regular project (not as an installed plugin).
- Plugin install path is unchanged — it still reads `.claude-plugin/plugin.json`.

## Why

`${CLAUDE_PLUGIN_ROOT}` is only defined when Claude Code loads the repo as an installed plugin. Opening the repo as a normal project leaves the var undefined, so every session starts with:

```
[Warning] [context-mode] mcpServers.context-mode: Missing environment variables: CLAUDE_PLUGIN_ROOT
```

A relative path resolves correctly when Claude Code spawns the MCP server from the project root. Plugin-mode continues to use `.claude-plugin/plugin.json`, which is unchanged.

## Pros / Cons

**Pros**
- Removes the startup warning for contributors.
- No behavior change for installed-plugin users.
- Single-line change.

**Cons**
- Depends on Claude Code spawning the MCP server with the project root as cwd. If that assumption ever breaks, the relative path would too — but so would most `.mcp.json` files in the wild.

## Test plan

- [ ] Fresh clone opened in Claude Code shows no `CLAUDE_PLUGIN_ROOT` warning.
- [ ] `/context-mode:ctx-doctor` still passes on a normal plugin install.
- [ ] `node ./start.mjs` starts successfully from the repo root.

Closes #252

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -2,7 +2,7 @@
   "mcpServers": {
     "context-mode": {
       "command": "node",
-      "args": ["${CLAUDE_PLUGIN_ROOT}/start.mjs"]
+      "args": ["./start.mjs"]
     }
   }
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "71.5k+",
+  "message": "71.7k+",
   "color": "brightgreen",
   "npm": "61.2k+",
-  "marketplace": "10.2k+"
+  "marketplace": "10.5k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "68.8k+",
+  "message": "70.5k+",
   "color": "brightgreen",
-  "npm": "58.4k+",
+  "npm": "60.1k+",
   "marketplace": "10.3k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "70.5k+",
+  "message": "71.6k+",
   "color": "brightgreen",
-  "npm": "60.1k+",
+  "npm": "61.2k+",
   "marketplace": "10.3k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "66.4k+",
+  "message": "68.8k+",
   "color": "brightgreen",
-  "npm": "56.1k+",
+  "npm": "58.4k+",
   "marketplace": "10.3k+"
 }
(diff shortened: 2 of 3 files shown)


flightlesstux:
CI failure was a stale test assertion, not a regression from this PR.

`tests/core/cli.test.ts:145` required `${CLAUDE_PLUGIN_ROOT}` in repo-root `.mcp.json` under the name _"keeps ${CLAUDE_PLUGIN_ROOT} for marketplace compatibility"_ — but marketplace installs actually resolve the placeholder from `.claude-plugin/plugin.json`, not the repo-root file. The assertion was pointing at the wrong manifest.

Fixed in 3f66152 by splitting the test in two:

- **Plugin manifest must keep the placeholder** — real marketplace contract, asserts `.claude-plugin/plugin.json` still contains `${CLAUDE_PLUGIN_ROOT}`.
- **Repo-root `.mcp.json` must use a relative path** — this PR's intent, asserts the path is relative (not the env var).

Local `npx vitest run tests/core/cli.test.ts` → 94/94 passing. CI should be green on the next run.

PR #257, opened 2026-04-12
fix(batch_execute): coerce command objects missing label field

### LINKED ISSUE #256: ctx_batch_execute rejects command objects missing 'label' field (common LLM output)
## Summary

`ctx_batch_execute` fails validation when the model passes command objects with `description` (or any key other than `label`) instead of `label`. LLMs frequently produce this shape because `description` is the more natural field name.

## Repro

Call `ctx_batch_execute` with:
```json
{
  "commands": [
    {"command": "ls -la", "description": "List repo root"}
  ],
  "queries": ["..."]
}
```

Result:
```
MCP error -32602: Input validation error: Invalid arguments for tool ctx_batch_execute: [
  {
    "code": "invalid_type",
    "expected": "string",
    "received": "undefined",
    "path": ["commands", 0, "label"],
    "message": "Required"
  }
]
```

## Root Cause

`coerceCommandsArray` in `src/server.ts` (line 1148) only handles string→`{label, command}` coercion. Object items without `label` fall through unchanged and fail the zod schema.

```ts
function coerceCommandsArray(val: unknown): unknown {
  const arr = coerceJsonArray(val);
  if (Array.isArray(arr)) {
    return arr.map((item, i) =>
      typeof item === "string" ? { label: \`cmd_\${i + 1}\`, command: item } : item
    );
  }
  return arr;
}
```

## Proposed Fix

Extend the coercer so object items missing `label` are synthesized — prefer `description`/`name`/`title` if present, else `cmd_N`. Add test coverage.

## Impact

This affects real agent sessions. Claude Code agents routinely produce `{command, description}` shape, causing the primary tool of this plugin to fail silently.

flightlesstux:
Fixes #256

## Summary
- `ctx_batch_execute` failed validation (`-32602`) when command items arrived as `{command, description}` instead of `{label, command}` — a very common LLM output shape.
- Extended `coerceCommandsArray` so object items missing a valid `label` synthesize one from `description` → `name` → `title` → `cmd_N`.
- Explicit non-empty `label` is always preserved; empty-string labels are replaced with a fallback.
- Added `tests/batch-execute-label-coercion.test.ts` (7 cases) covering each fallback path plus mixed-shape arrays.

## Why
This is the primary tool of the plugin. Agents routinely emit `description` instead of `label` because `description` is the more natural field name, causing silent failures that waste a full round trip.

## Test plan
- [x] `npx vitest run tests/batch-execute-label-coercion.test.ts` — 7 passed
- [x] `npm run build` — bundles regenerated
- [x] Full suite delta: +7 passing, no new failures (12 pre-existing failures unchanged on baseline main)

Co-authored-by: Ercan Ermis <eposta@ercanermis.com>


mksglu:
Thanks for identifying this issue @flightlesstux!

We took a different approach — instead of adding fallback coercion, we documented the `label` field across all 12 adapter configs and the routing block that gets injected into LLM prompts.

**Why docs instead of coercion:**
- `label` serves triple duty: markdown section header, FTS5 source key, and chunk boundary marker
- A descriptive label directly improves search quality — auto-generated `cmd_1` labels degrade it
- The root cause is that LLMs didn't know `label` exists, not that the schema is wrong

**What shipped (on `next`):**
- `hooks/routing-block.mjs` GATHER section now shows `{label: "descriptive section header", command: "..."}` with explanation
- All 11 platform config files (CLAUDE.md, AGENTS.md, GEMINI.md, etc.) updated
- Test added to verify routing block contains label guidance

Closing this PR in favor of the docs approach. The issue itself is fixed — will close with a comment.

PR #259, opened 2026-04-12
fix(tests): disable fileParallelism to eliminate subprocess SIGKILL flakiness

### LINKED ISSUE #258: Flaky test suite: ~12 tests SIGKILL'd under parallel fork pool (spawnSync + better-sqlite3)
## Summary

~12 tests across 6 files intermittently fail on local runs with `expected 0 to be 0` (from `spawnSync` returning `status: null`, `signal: SIGKILL`, empty stdout and stderr). All runs pass when each file is executed in isolation.

## Reproduction

```bash
npx vitest run
```

Observed (repeatable locally, 3/3 runs):

```
Test Files  6 failed | 38 passed (45)
Tests       12 failed | 1165 passed | 14 skipped (1434)
[vitest-pool]: Timeout terminating forks worker for test files tests/core/...
```

Affected files (all spawn Node subprocesses via `spawnSync`):
- `tests/session-hooks-smoke.test.ts`
- `tests/session/continuity.test.ts`
- `tests/hooks/cursor-hooks.test.ts`
- `tests/hooks/gemini-hooks.test.ts`
- `tests/hooks/vscode-hooks.test.ts`

Running any one of them in isolation: fully green.

## Root Cause

`spawnSync` result shows the subprocess was externally killed:
```
status=null signal=SIGKILL error=undefined stderr=[] stdout=[]
```

Under parallel load (8 CPUs → 8 forks, each running a file that spawns a Node child loading the `better-sqlite3` native addon), some child subprocesses are SIGKILL'd before they produce any output. The signal originates from vitest's fork-worker teardown path (`Timeout terminating forks worker`) — when a worker takes too long, vitest kills it, and the signal propagates to the child `spawnSync` targets that share the worker's process group.

Existing `retry: process.env.CI ? 2 : 0` in `vitest.config.ts` absorbs the symptom on CI but leaves local runs permanently flaky and masks real regressions.

## Fix

Disable inter-file parallelism via `fileParallelism: false`. Files run sequentially; tests within a file are already sequential (`spawnSync` is synchronous). This eliminates the race deterministically.

Trade-off: total wall time ~10s → ~36s locally on an 8-CPU machine. Worth it for green-every-run.

Verified: 3/3 full runs → 44/44 files pass.

A more surgical fix (only serialize the subprocess-heavy files via `vitest.workspace` projects) is possible but adds config complexity for a suite that completes in ≤40s.

flightlesstux:
Fixes #258

## Summary
- Root cause confirmed: `spawnSync` subprocesses in hook suites were externally SIGKILL'd (`status=null, signal=SIGKILL, stderr=""`) when vitest ran their files concurrently across fork workers — signal leaks out of worker-teardown under load.
- Set `fileParallelism: false` so files execute sequentially. Tests within a file already ran sequentially (`spawnSync` is synchronous), so the only thing sacrificed is inter-file concurrency.
- Dropped the CI-only `retry: 2` which was masking the race rather than fixing it.

## Diagnostic Evidence
Single-file runs of every failing suite: **green**.
Full suite baseline: **6 failed / 45, 12 tests fail**, 3 consecutive runs.
With `fileParallelism: false`: **45 passed / 45**, 3 consecutive runs (36s vs ~10s).
Signal was captured by augmenting `runHook` temporarily — subprocess died before producing any output.

## Trade-off
Local wall-time increases from ~10s to ~36s on an 8-CPU machine. Acceptable for a <1-minute suite in exchange for deterministic green runs on every machine.

A more surgical alternative (vitest projects splitting hook-subprocess suites into their own sequential pool) was considered and rejected — more config surface for a suite that already completes in under a minute.

## Test plan
- [x] `npx vitest run` — 3 consecutive full runs, 44/44 files pass, 0 flakes
- [x] No source code touched — config-only change
- [x] CI run will validate on Linux once this PR opens

Co-authored-by: Ercan Ermis <eposta@ercanermis.com>

### DIFF
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -7,10 +7,13 @@ export default defineConfig({
     // Native addons (better-sqlite3) can segfault in worker_threads during
     // process cleanup. Use forks on all platforms for stable isolation.
     pool: "forks",
-    // Hook subprocess tests (spawnSync + better-sqlite3 native addon) can
-    // fail intermittently under parallel load on CI.  Retry once to absorb
-    // transient resource-contention failures without masking real regressions.
-    // Only enable retry on CI to avoid slowing down local dev.
-    retry: process.env.CI ? 2 : 0,
+    // Several suites spawn Node subprocesses via spawnSync (hook runners)
+    // that load the better-sqlite3 native addon. When vitest ran those
+    // files concurrently in separate fork workers, the child processes
+    // intermittently received SIGKILL (empty stdout/stderr, status=null) —
+    // likely from worker-teardown signal propagation under load.
+    // Serializing files eliminates the race deterministically; tests
+    // within a file still run sequentially as before (spawnSync is sync).
+    fileParallelism: false,
   },
 });


mksglu:
Hey @flightlesstux — thank you for this PR! You identified the exact problem: parallel fork pool + spawnSync + better-sqlite3 = SIGKILL'd workers.

We went with a slightly different approach on `next` — instead of `fileParallelism: false` (which drops to sequential ~23s), we benchmarked `maxWorkers: 3` which gives **2.8x speedup with near-zero crashes** (vs unlimited 3.7x with 6-7 crashes/run):

| Config | Duration | Worker Crashes |
|--------|----------|---------------|
| Unlimited (default) | 6.3s | 6-7/run |
| **maxWorkers=3** (shipped) | **8.2s** | **~1/run** |
| fileParallelism=false (your PR) | 23.3s | 0/run |

Commit: [`53d3c1e`](https://github.com/mksglu/context-mode/commit/53d3c1e)

Your analysis in the issue was spot-on and directly informed the fix. Closing this PR since the approach is already on `next`, but genuinely appreciate the investigation work. 🙏

PR #267, opened 2026-04-13
perf: enable mmap_size pragma for read-heavy FTS5 search

sebastianbreguel:
## Summary

- Adds `PRAGMA mmap_size = 268435456` (256MB) to `applyWALPragmas` in `db-base.ts`
- Memory-maps the DB file so the kernel serves FTS5 search reads directly from the page cache
- Eliminates `read()` syscalls on the hot search path

## Why this matters

FTS5 search is read-heavy — every `MATCH` query scans b-tree segments. Without mmap, each page access triggers a `read()` syscall through the kernel. With mmap, SQLite reads pages directly from virtual memory (the kernel's page cache handles it). This is especially beneficial for long sessions where the FTS5 index grows.

## Design decisions

- **256MB cap**: SQLite only maps up to actual file size, so this is a safe upper bound. Typical context-mode DBs are 1-10MB.
- **Applied in `applyWALPragmas`**: affects both ContentStore (ephemeral) and SessionDB (persistent) — both benefit from faster reads.
- **try/catch guard**: some runtimes (bun:sqlite, node:sqlite) may not support mmap — fails gracefully.
- **No new dependencies, 1 line of functional code**.

## Test plan

- [x] Store test added (search works correctly with mmap enabled)
- [x] All 93 store tests pass
- [x] All 100 search tests pass
- [x] All 28 session-db tests pass
- [ ] CI validation on Windows/Linux/macOS


mksglu:
@sebastianbreguel This branch has conflicts that must be resolved

sebastianbreguel:
@mksglu ready!

Issue #247, opened 2026-04-11
ctx_doctor MCP tool crashes the server mid-session


lanarasp:
### Platform

Claude Code

### context-mode version

1.0.75

### Debug script output (REQUIRED)

```json
{
  "version": "2.0.0",
  "generated": "2026-04-11T00:31:45Z",
  "sections": {
    "1. System Info": {
      "checks": [],
      "info": {
        "OS type": "linux",
        "uname -a": "Linux programes 6.1.75-1.qubes.fc32.x86_64 #1 SMP PREEMPT_DYNAMIC Mon Feb  5 06:49:25 CET 2024 x86_64 GNU/Linux",
        "Architecture": "x86_64",
        "Distro": "Debian GNU/Linux 11 (bullseye)",
        "Shell": "/bin/bash",
        "Bash version": "5.1.4(1)-release",
        "Locale": "LANG=en_US.UTF-8, LC_ALL=unset"
      },
      "configs": [],
      "warnings": [],
      "details": []
    },
    "2. Runtime Versions": {
      "checks": [],
      "info": {
        "Node.js": "v20.20.2",
        "node path": "/usr/bin/node",
        "execPath": "/usr/bin/node",
        "Node install method": "unknown",
        "Bun": "1.3.11",
        "bun path": "/home/user/.bun/bin/bun",
        "Python": "Python 3.9.2",
        "Ruby": "ruby 2.7.4p191 (2021-07-07 revision a21a3b7d23) [x86_64-linux-gnu]",
        "npm": "10.8.2",
        "npm global root": "/usr/lib/node_modules"
      },
      "configs": [],
      "warnings": [],
      "details": []
    },
    "3. context-mode Installation": {
      "checks": [
        {
          "pass": false,
          "label": "build/ directory exists"
        },
        {
          "pass": true,
          "label": "hooks/pretooluse.mjs exists"
        },
        {
          "pass": true,
          "label": "hooks/sessionstart.mjs exists"
        },
        {
          "pass": true,
          "label": "server.bundle.mjs exists"
        },
        {
          "pass": true,
          "label": "cli.bundle.mjs exists"
        }
      ],
      "info": {
        "Installed version": "1.0.75",
        "Plugin root": "/home/user/.claude/plugins/cache/context-mode/context-mode/1.0.75",
        "Install method": "marketplace",
        "npm latest": "1.0.75"
      },
      "configs": [],
      "warnings": [],
      "details": []
    },
[...]

PR #268, opened 2026-04-14
test: regression coverage for ctx_doctor resource cleanup (#247)

sebastianbreguel:
## Summary

Adds a vitest integration test that spawns the real MCP server over stdio
(same as Claude Code does) and exercises `ctx_doctor`:

1. **Single call** — asserts the handler returns a markdown checklist with
   the `Server test:` and `FTS5 / SQLite:` lines. These are the two blocks
   that bb6552d wrapped in `try/finally` to prevent resource leaks.
2. **Three concurrent calls** — reproduces the exact scenario from issue
   #247 ("3 context-mode processes were running at time of crash"). If the
   fix regresses, better-sqlite3 GC segfaults tear down stdio and at least
   one response goes missing — the test fails loudly.

No production code changes. Just a safety net so a future refactor that
drops the `finally` blocks or forgets to `.close()` the test DB fails in
CI instead of silently crashing user sessions.

Local run: `npx vitest run tests/core/ctx-doctor.test.ts` — 3/3 pass in ~20s.

## Test plan

- [x] `npx vitest run tests/core/ctx-doctor.test.ts` passes locally
- [ ] CI passes on all platforms


mksglu:
@sebastianbreguel Thanks, can we use already exist test file for that instead of create new one? Ref: CONTRIBUTION.md

sebastianbreguel:
Done — moved tests into `tests/core/server.test.ts` and deleted the standalone file. 102 tests pass locally.

PR #271, opened 2026-04-14
fix(search): filter stopwords from FTS5 queries and proximity reranking

sebastianbreguel:
## Summary

- Domain-specific stopwords (`update`, `test`, `fix`, `run`, `using`, etc.) were defined in `STOPWORDS` but only used for vocabulary extraction — search queries passed them through to FTS5 unfiltered, diluting BM25 ranking
- Now filtered in `sanitizeQuery()`, `sanitizeTrigramQuery()`, and `#applyProximityReranking()` — meaningful terms drive ranking while stopwords are dropped
- Falls back to unfiltered terms when ALL query words are stopwords (no empty results)

## Example

Query `"fix database connection"` — before, `"fix"` got equal BM25 weight despite appearing in most chunks. Now only `"database"` and `"connection"` drive the ranking.

## Test plan

- [x] Stopwords filtered from porter search — meaningful terms drive ranking
- [x] All-stopword query falls back to unfiltered (no empty results)
- [x] Stopwords filtered from trigram search
- [x] Proximity reranking excludes stopwords from boost calculation
- [x] Full test suite — 0 regressions (5 pre-existing failures in hooks/integration unrelated)


mksglu:
@sebastianbreguel I hope you'd tested on manually also. :)

PR #276, opened 2026-04-14
perf(store): dedup tokens in sanitizeQuery / sanitizeTrigramQuery

sebastianbreguel:
## Problem

`sanitizeQuery` and `sanitizeTrigramQuery` in `src/store.ts` don't deduplicate tokens before compiling the FTS5 query string.

```
search(\"error error error\")
  → sanitizeQuery: \"error\" \"error\" \"error\"
```

FTS5 resolves the same token three times (and in OR mode, adds redundant OR branches). Duplicates show up naturally from:

- Copy-pasted log snippets / stack traces (\"Error: foo. Error: bar.\")
- Multi-token queries that re-use the same keyword
- Fuzzy-corrected queries where multiple inputs collapse to the same correction

Since FTS5's unicode61 tokenizer lowercases on both sides, duplicates never add recall — they just waste lookups.

## Fix

Add a small `dedupeTokens` helper that collapses case-insensitive duplicates while preserving first-occurrence order and casing. Applied in both sanitize functions right after operator stripping, before the stopword filter.

The STOPWORDS-fallback path (used when every token is a stopword) now falls back to the deduped list — semantically identical, just less noise.

Exports `sanitizeQuery` / `sanitizeTrigramQuery` so tests can assert on compiled query strings directly.

## Tests

6 new tests in `tests/store.test.ts` under `describe(\"Sanitize query token deduplication\")`:

- Plain duplicate collapse
- Case-insensitive duplicate collapse
- First-occurrence casing preserved (\"Database database DATABASE\" → \"Database\")
- Distinct tokens preserved in first-seen order
- Trigram path dedup
- OR-mode dedup
- Behavioral: `search(\"error error error database database\")` returns byte-equal rows to `search(\"error database\")`

## Verification

```
npx vitest run tests/store.test.ts tests/core/search.test.ts
 Test Files  2 passed (2)
      Tests  228 passed (228)
```

Full suite: 1432 pass. Same 6 pre-existing failures on `main` (security / hook integration — unrelated to this change).


mksglu:
@sebastianbreguel CI

sebastianbreguel:
@mksglu ready

PR #278, opened 2026-04-14
test(exit-classify): cover classifyNonZeroExit branches

sebastianbreguel:
## Context

`src/exit-classify.ts` encodes a narrow but important contract: shell commands that exit 1 with non-empty stdout (think `grep`, `find`, `test`) are treated as soft failures so their output reaches the caller unwrapped, while every other non-zero exit is wrapped in a formatted \`Exit code: N / stdout / stderr\` block.

That logic was only exercised indirectly via the executor integration tests, which made the soft-fail carve-out easy to widen or narrow by accident.

## Change

Add \`tests/exit-classify.test.ts\` — 6 focused tests pinning each branch:

- \`shell\` + exit 1 + non-empty stdout → **soft fail**, stdout returned verbatim (no \`Exit code:\` prefix)
- \`shell\` + exit 1 + whitespace-only stdout → hard fail
- \`shell\` + exit 1 + empty stdout + stderr → hard fail (formatted output includes stderr)
- \`shell\` + exit code ≠ 1 → never soft even with stdout
- non-shell language + exit 1 + stdout → hard fail (soft fail is shell-only)
- hard-fail output always includes both stdout and stderr sections

Zero changes to `src/`. Pure coverage addition.

## Verification

\`\`\`
npx vitest run tests/exit-classify.test.ts
# 6 pass, 0 fail
\`\`\`


mksglu:
Hi @sebastianbreguel Thanks but please use exist test files instead of create one.

mksglu (changes requested):
Please, don't create new test files

sebastianbreguel:
Thanks for the feedback @mksglu — force-pushed a refactor that folds the new assertions into the existing `Non-zero Exit Code Classification` block in `tests/core/server.test.ts` and deletes the standalone file. I kept only what wasn't already covered: (1) stderr still forwards through the formatted output when stdout is empty, and (2) the hard-fail output has labeled `stdout:` / `stderr:` sections.

mksglu:
@sebastianbreguel CI

PR #279, opened 2026-04-14
fix(pi): show ctx command output in UI

h4rvey-g:
## Summary
- make `ctx-stats` use the Pi UI notification API when UI context is available
- make `ctx-doctor` use the Pi UI notification API when UI context is available
- keep message-based fallback behavior for non-UI contexts

## Testing
- npm run typecheck


mksglu:
Hi @h4rvey-g CI is broken.

PR #288, opened 2026-04-15
fix(security): resolve path traversal bypass in evaluateFilePath

pbbadenhorst:
## What / Why / How

**What:** `evaluateFilePath` (and therefore `checkFilePathDenyPolicy` for `ctx_execute_file` / Read-routed tools) now matches deny globs against both the raw input string and the fully-resolved absolute path.

**Why:** Read deny patterns were matched only against the raw input. An absolute-path deny rule like `Read(/Users/you/.ssh/**)` could be bypassed by passing a relative path with `..` segments that resolves into the same location — e.g. `../../.ssh/id_rsa` from a project nested under `$HOME`. A prompt-injected `ctx_execute_file` call could then read SSH keys, AWS creds, `.env` files, etc. that the user explicitly tried to deny.

**How:**
- `evaluateFilePath` (`src/security.ts`) gains an optional 4th parameter `projectRoot`. When supplied, the path is also resolved via `path.resolve(projectRoot, filePath)` and the resolved form is matched against every deny glob in addition to the raw input.
- `checkFilePathDenyPolicy` (`src/server.ts`) passes `process.env.CLAUDE_PROJECT_DIR ?? process.cwd()` so the resolved form is checked at runtime.
- Backwards compatible: callers that don't pass `projectRoot` see identical behavior. A regression-guard test documents that pre-fix behavior so any future change is intentional.

## Affected platforms

- [x] All platforms

(The fix is in shared security/server code that all adapters route through.)

## Test plan

Added two tests to `tests/security.test.ts` in the existing `evaluateFilePath` describe block:

1. **`traversal does not bypass absolute deny glob when projectRoot is supplied`** — the bypass case. With `projectRoot = ~/some/project` and deny glob `~/.ssh/**`, asserts that `../../.ssh/id_rsa` is denied. Fails before the fix, passes after.
2. **`without projectRoot, absolute deny glob is still bypassable (regression guard)`** — pins the legacy 3-arg behavior so future changes to it are deliberate.

Also verified:
- `npm test` → 1540 passed / 17 skipped / 0 failed (46 files)
- `npm run typecheck` → clean

## Checklist

- [x] Tests added/updated (TDD: red → green confirmed locally)
- [x] `npm test` passes
- [x] Docs updated if needed (no doc surface changed — internal function gained an optional param)
- [x] No Windows path regressions (forward-slash normalization preserved; `path.resolve` is platform-aware)
- [x] Targets `next` branch

### DIFF
--- a/src/security.ts
+++ b/src/security.ts
@@ -429,19 +429,35 @@ export function evaluateCommandDenyOnly(
  *
  * Normalizes backslashes to forward slashes before matching so that
  * Windows paths work with Unix-style glob patterns.
+ *
+ * When `projectRoot` is supplied, the path is also matched in its
+ * resolved absolute form — preventing `..` traversal from bypassing
+ * absolute-path deny rules like `Read(/Users/.../.ssh/**)`.
  */
 export function evaluateFilePath(
   filePath: string,
   denyGlobs: string[][],
   caseInsensitive: boolean = process.platform === "win32",
+  projectRoot?: string,
 ): { denied: boolean; matchedPattern?: string } {
-  // Normalize backslashes to forward slashes for cross-platform matching
-  const normalized = filePath.replace(/\\/g, "/");
+  const toForward = (path: string): string => path.replace(/\\/g, "/");
+
+  // Match against both the raw input and (when projectRoot is supplied)
+  // the fully-resolved absolute path. Deduplicated so tests and callers
+  // that pass an already-absolute path don't pay the matching cost twice.
+  const candidates: string[] = [toForward(filePath)];
+  if (projectRoot) {
+    const absolute = toForward(resolve(projectRoot, filePath));
+    if (absolute !== candidates[0]) candidates.push(absolute);
+  }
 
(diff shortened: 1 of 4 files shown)


mksglu:
## Security Review Complete ✅

Thorough review done — the 3-candidate matching approach (raw + lexical resolve + realpath) is architecturally sound and correctly closes the path traversal bypass.

**What we verified:**
- `../../.ssh/id_rsa` bypasses absolute deny globs on `main` — confirmed exploitable
- `path.resolve(projectRoot, filePath)` catches lexical traversal — ✅
- `fs.realpathSync` catches symlink chain escapes — ✅
- No bypass vectors found (double encoding, null bytes, unicode normalization, TOCTOU — all analyzed)
- Windows backslash handling via `toForward()` — correct
- Backwards-compatible (optional `projectRoot` parameter) — ✅
- CI green on all 3 OS (ubuntu, macos, windows)

Clean, minimal fix for a real vulnerability. Great work @pbbadenhorst — thank you for the contribution! 🙏

Issue #249, opened 2026-04-12
Claude Code marketplace install can leave plugin without build/ and node_modules


uwuclxdy:
## Bug

> **generated by Claude**
>
>it fixed the install for me but i don't understand this enough nor have the time to go through it myself, here is what it has found about this:

---

Claude Code marketplace install can leave `context-mode` in a source-like state under `~/.claude/plugins/marketplaces/context-mode` without `node_modules/` and without `build/`.

In that state, the plugin is broken until the user manually runs:

```bash
cd ~/.claude/plugins/marketplaces/context-mode
npm install
npm run build
```

## Why this looks like a real install bug

The README says Claude Code marketplace install is **fully automatic**:

```bash
/plugin marketplace add mksglu/context-mode
/plugin install context-mode@context-mode
```

But the installed marketplace copy I inspected had:

- `.git/` present (looks like a repo checkout, not a packed npm install)
- no `node_modules/`
- no `build/`
- hooks present, including bundled hook files

That broke runtime in two ways:

1. Native dependency unavailable:
   - `Cannot find module 'better-sqlite3'`
   - `Could not locate the bindings file ... better_sqlite3.node`

2. Platform hooks importing missing build artifacts:
   - `Cannot find module '.../build/session/db.js' imported from .../hooks/sessionstart.mjs`

## Current code evidence

`package.json` has:

- `"postinstall": "node scripts/postinstall.mjs"`
- `"prepublishOnly": "npm run build"`

So:
- `npm install` should run automatically **only if the installer actually performs an npm install**
- `npm run build` is **not** an end-user install hook; it only runs before publish

That matches what I observed: if the marketplace installer copies/clones the repo without a full package install/build step, the plugin is left broken.

Also, some platform hooks still import from `build/session/*.js` directly (for example `hooks/cursor/sessionstart.mjs`) instead of using the bundle-first loader path, so missing `build/` is fatal there.

## Reproduction

On Claude Code:

```bash
/plugin marketplace add mksglu/context-mode
/plugin install context-mode@context-mode
```

Then inspect:

```bash
ls ~/.claude/plugins/marketplaces/context-mode
```

Broken install characteristics I observed before manual repair:

- `node_modules/` missing
- `build/` missing
- runtime hook crashes from missing `build/session/db.js`
- `better-sqlite3` unavailable

## Manual workaround

```bash
cd ~/.claude/plugins/marketplaces/context-mode
npm install
npm run build
```

After that, the plugin recovered:

[...]

Issue #244, opened 2026-04-10
MCP server dies on heavy ctx_execute workloads


khaosdoctor:
## Summary

The context-mode MCP server (v1.0.75) crashes with `MCP error -32000: Connection closed` on heavy `ctx_execute` / `ctx_batch_execute` workloads. The crash cascades: a second call after the first failure also dies immediately, wiping tool registration from Claude Code until the user restarts.

Could be related to #242 and #236 

## Environment

- context-mode **1.0.75**
- Claude Code on Linux (Arch, kernel 6.19)
- Node **v25.2.1**, better-sqlite3 **12.8.0** (abi141 prebuilt present)
- Workload: processing ~173 JSONL files (~46 MB total) from `~/.claude/projects/`

## Reproduction

Twice in the same session, while scanning conversation logs:

1. Call 1: `ctx_execute(language: "javascript", code: <~60 LOC JS that reads 173 files, regex-scans ~8 MB of text, pushes substrings into an array, console.logs a small summary>, intent: "...")`. Child finished; parent MCP died mid-response. `ctx_stats` afterwards returned `"session (< 1 min)"` then the server had silently restarted.

2. Call 2: immediately after, a trivial `ctx_batch_execute` with six short shell commands (`ls`, `find -name ... | head`, `cat package.json`). Server died again and this time Claude Code **stopped re-registering the tools entirely**, all `mcp__plugin_context-mode_*` tools disappeared for the rest of the session.

This has happened before in my machine:

- `~/.claude/context-mode/sessions/` held **7 `.db` files but only 6 `.cleanup` markers** so, orphaned session state from past crashes.
- `~/.claude/context-mode/content/<session>.db` was **4 KB / effectively empty** the persistent index never populated despite many hours of use across sessions.
- Historical `sessionstart-debug.log`: `Cannot find module 'better-sqlite3'` thrown from `plugins/marketplaces/context-mode/hooks/session-db.bundle.mjs`. The module is now present in that path, so this specific historical error is resolved, but the empty index DB suggests the write path still fails silently somewhere.

## Suspected root cause

`src/server.ts:45-50` installs `uncaughtException` / `unhandledRejection` handlers that only `process.stderr.write(...)`, they don't re-throw or `process.exit`, so intent is "log and continue", but there's other things happening:

[...]

Issue #218, opened 2026-04-02
MCP server crashes under multi-session SQLite contention (SQLITE_BUSY)


john-savepoint:
## Description

When running multiple Claude Code sessions simultaneously (or even a single session that reconnects via `/mcp`), the context-mode MCP server frequently crashes and disconnects. The root cause is SQLite write contention on the shared content database.

## Environment

- context-mode v1.0.57
- Platform: WSL2 (Linux 6.6.87.2-microsoft-standard-WSL2)
- Claude Code CLI
- Multiple concurrent sessions sharing the same content DB

## Root Cause Analysis

The shared content database at `~/.context-mode/content/<hash>.db` is accessed by multiple server processes simultaneously. The current configuration:

- `journal_mode = WAL` (good — allows concurrent reads)
- `synchronous = NORMAL`
- **`timeout: 5000` (5 second busy timeout — too short)**

When multiple processes attempt concurrent writes (e.g., indexing execution results), one process holds the write lock while others wait. If the write exceeds 5 seconds, waiting processes receive `SQLITE_BUSY`, which propagates as an unhandled error, crashing the server and killing the stdio MCP connection.

### Evidence

- 4 zombie server processes observed from a single session after repeated `/mcp` reconnects
- Only 1 of the 4 processes held open file descriptors on the `.db`, `.db-wal`, and `.db-shm` files
- Server crashes mid-tool-call (e.g., `ctx_execute` succeeds but `ctx_search` immediately after fails with "Connection closed")
- Each `/mcp` reconnect spawns a new process without killing the old one, compounding contention

### Relevant code (server.bundle.mjs)

```javascript
// Content DB constructor — 5s timeout
this.#e = new r(this.#t, { timeout: 5e3 });

// WAL mode setup
function sx(t) {
  t.pragma("journal_mode = WAL");
  t.pragma("synchronous = NORMAL");
}
```

## Suggested Fixes

1. **Increase busy timeout** to 30-60 seconds — WAL write locks are typically brief, so a longer timeout would let processes wait instead of crashing
2. **Add retry logic** around `SQLITE_BUSY` errors with exponential backoff
3. **Graceful error handling** — catch `SQLITE_BUSY` at the tool-call level and return an MCP error instead of crashing the entire server process
4. **Consider process cleanup** — detect and clean up stale server processes on startup (the `up()` function already does this for temp DBs but not for the content DB lock holders)

## Impact

[...]

PR #292, opened 2026-04-15
test(db-base): cover withRetry, corruption predicate, and cleanup helpers

sebastianbreguel:
## Why

`src/db-base.ts` (493L) carries the SQLite primitives behind several recent incident fixes — `withRetry` (#263 wrapped `indexPlainText`/`indexJSON`/`searchTrigram`, #243 concurrent inserts), `isSQLiteCorruptionError` + `renameCorruptDB` (recovery path for #244 heavy-load death and #218 multi-session contention), and the WAL/cleanup helpers touched by #103 lifecycle guard and #249 marketplace-install fallout. Until now the module was only exercised indirectly via `SessionDB` integration tests.

This PR is pure **regression-guard coverage** for the pure, framework-free utilities. No changes to `src/`.

## What's covered

**`withRetry`** (11 tests) — first-try success, retry-on-busy (Error), retry on `"database is locked"` string shape (bun:sqlite), rethrow of non-busy errors, rethrow of generic errors, non-`Error` throws (string, busy-shaped vs not), exhaustion with descriptive `after N retries` message, `attempts = delays.length + 1` invariant, empty-delays short-circuit, and actual wall-clock delay verification (50ms lower bound).

**`isSQLiteCorruptionError`** (6 tests) — all four documented signatures (`SQLITE_CORRUPT`, `SQLITE_NOTADB`, `database disk image is malformed`, `file is not a database`), embedded in longer stack traces, and negative cases (`SQLITE_BUSY`, `ENOENT`, empty string).

**`renameCorruptDB`** (4 tests) — main-file quarantine with `.corrupt-<ts>` suffix, sidecar `-wal`/`-shm` rename when present, graceful no-op when sidecars or main file are missing.

**`cleanOrphanedWALFiles`** (3 tests) — sidecars removed only when the main DB is gone, sidecars preserved when DB still exists, tolerates empty dirs.

**`deleteDBFiles`** (3 tests) — unconditional removal of all three files, tolerant of partial/full absence.

**`defaultDBPath`** (4 tests) — pid embedding, custom prefix, default prefix, tmpdir scoping.

## Test plan

- [x] `npx vitest run tests/db-base.test.ts` → 31/31 green, 66ms
- [x] Full suite on this branch vs clean `next`: same pre-existing 5 flakes in `tests/hooks/integration.test.ts` (`SyntaxError: Unexpected end of JSON input` from subprocess stdout) — not introduced here.
- [x] No imports or exports in `src/` changed.

Single-concern PR; no runtime changes, no perf claims.

### DIFF
--- a/tests/session/session-db.test.ts
+++ b/tests/session/session-db.test.ts
@@ -1,11 +1,18 @@
 import { strict as assert } from "node:assert";
 import { join } from "node:path";
-import { mkdtempSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { randomUUID } from "node:crypto";
-import { existsSync, writeFileSync, readdirSync } from "node:fs";
+import { existsSync, mkdtempSync, readdirSync, rmSync, writeFileSync } from "node:fs";
 import { afterAll, describe, expect, test } from "vitest";
 import { SessionDB } from "../../src/session/db.js";
+import {
+  cleanOrphanedWALFiles,
+  defaultDBPath,
+  deleteDBFiles,
+  isSQLiteCorruptionError,
+  renameCorruptDB,
+  withRetry,
+} from "../../src/db-base.js";
 
 const cleanups: Array<() => void> = [];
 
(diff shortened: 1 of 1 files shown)


mksglu:
Hi @sebastianbreguel Nice work. Can you please continue that with exist test files instead of create a new one.

sebastianbreguel:
Hi @mksglu, thanks for the feedback! Done — moved all 31 db-base tests into the existing `tests/session/session-db.test.ts` (`SessionDB` extends these primitives, so they live together now) and deleted the separate `tests/db-base.test.ts`. File goes from 609 → 893 lines. Full suite still green: 28 `SessionDB` + 31 `db-base` = 59/59 passing locally. Ready for another look.

PR #294, opened 2026-04-16
fix(insight): keep the local API same-origin by default

### LINKED ISSUE #293: Insight API is cross-origin readable in both runtimes and DELETE-enabled in Node runtime
## Observation

`ctx_insight` is documented as a **local** analytics UI, and the README frames context-mode as **privacy-first** / **nothing leaves your machine**.

While re-validating recent Insight changes against `main@72eacf5` (2026-04-16) and `v1.0.89`, I found that the Insight API still advertises permissive cross-origin access:

- `GET` API responses include `Access-Control-Allow-Origin: *`
- in the Node runtime, API responses also include `Access-Control-Allow-Methods: GET, DELETE, OPTIONS`
- the Node runtime accepts `OPTIONS` preflight requests for `DELETE`
- a different localhost origin can read session event data, and in Node mode it can also trigger destructive DELETEs against indexed content

I’m **not** claiming a remote-network exploit here. This is a **same-machine / different-origin trust-boundary** issue for the local dashboard.

## Repro steps

I reproduced this in a local fork against the latest upstream `main` using fixture SQLite DBs and two Insight runtimes.

### Node runtime (verified read + DELETE)

1. Start Insight with fixture DBs:
   - `PORT=4748`
   - `INSIGHT_SESSION_DIR=/tmp/context-mode-cors-repro/sessions`
   - `INSIGHT_CONTENT_DIR=/tmp/context-mode-cors-repro/content`
2. Send a cross-origin-style GET:

```bash
curl -i http://127.0.0.1:4748/api/sessions/abcd1234/events/sess-1 \
  -H 'Origin: http://127.0.0.1:8081'
```

3. Send a preflight for DELETE:

```bash
curl -i -X OPTIONS http://127.0.0.1:4748/api/content/feedface/source/7 \
  -H 'Origin: http://127.0.0.1:8081' \
  -H 'Access-Control-Request-Method: DELETE'
```

4. Send the DELETE itself:

```bash
curl -i -X DELETE http://127.0.0.1:4748/api/content/feedface/source/7 \
  -H 'Origin: http://127.0.0.1:8081'
```

### Bun runtime (verified read)

1. Start Insight with the same fixture DBs on `PORT=4747`
2. Send the same GET request with `Origin: http://127.0.0.1:8081`
3. Sensitive session data is returned with `Access-Control-Allow-Origin: *`

In Bun mode, DELETE preflight currently falls through to the HTML response instead of returning a permissive preflight, so the destructive path appears blocked there by broken preflight handling. The read exposure is still present.

## Expected

The local dashboard should remain same-origin by default.

A different localhost origin should **not** be able to:
- read session event / resume data from the Insight API
- preflight and issue destructive API calls

## Actual

- both runtimes allow cross-origin reads of sensitive session data
- the Node runtime also permits cross-origin DELETE preflight + DELETE execution

## Relevant code

Current behavior appears to come from `insight/server.mjs`:

- `apiSessionEvents` / `apiDeleteSource`
- route handling for `/api/sessions/.../events/...` and `DELETE /api/content/...`
- permissive headers in both Bun and Node server branches

## Environment

- Repo: `mksglu/context-mode`
- Latest upstream `main` tested: `72eacf5`
- Latest release also checked: `v1.0.89`
- Local validation on macOS
- Verified with both Bun and Node runtimes

## Suggested direction

I think this is worth discussing as an **issue first**, because there are multiple valid fixes:

- remove permissive CORS headers entirely for Insight API routes
- keep the dashboard same-origin only by default
- optionally add an explicit future opt-in sharing/token model if cross-origin access is ever needed

If you'd like, I can open a small follow-up PR with regression coverage and the narrowest possible fix.

shaun0927:
## What / Why / How

Fixes #293

**What:** remove permissive cross-origin headers from Insight API responses and add a focused regression test for same-machine cross-origin access.

**Why:** the verified issue is not “remote internet access”, but a local trust-boundary problem: a different localhost origin can read Insight session data in both runtimes, and the Node runtime also permits DELETE preflight + DELETE requests. That conflicts with the README’s privacy-first / local-only framing for Insight.

**How:**
- introduce a shared JSON header object for Insight API responses without `Access-Control-Allow-*`
- stop the Node server branch from setting permissive CORS headers globally
- add `tests/analytics/insight-cors.test.ts`, which boots Insight against fixture SQLite DBs and asserts that sensitive session endpoints and DELETE preflights no longer advertise cross-origin access headers
- make the regression test hermetic by running the real `insight/server.mjs` against fixture DBs while temporarily stubbing `insight/dist/index.html`

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [x] All platforms

## Test plan

- `npm run build`
- `npm run typecheck`
- `npx vitest run tests/analytics/insight-cors.test.ts`
- `npm test`

Local verification before the patch:
- Bun runtime exposed session-event reads cross-origin
- Node runtime exposed session-event reads cross-origin and accepted DELETE preflight + DELETE cross-origin

Local verification after the patch:
- the new regression test passes
- the full test suite passes locally after rebuilding bundles
- Insight API no longer advertises permissive cross-origin headers

## Checklist

- [x] `npm test` passes

Notes:
- I am targeting `main` because that is the current default branch and the recent merged fixes around this area also landed there.
- The follow-up commit keeps the regression test self-contained instead of depending on a copied `server.mjs` tree that would lose `better-sqlite3` resolution.

<details>

This patch only changes Insight API response headers and adds a Node-based regression test.

- No filesystem path logic changed
- No hook path logic changed
- No stdin handling changed
- No hardcoded path separators introduced
- The Bun and Node runtime branches now align on same-origin-only API headers for API responses

</details>


mksglu:
Thanks @shaun0927 — clean, well-scoped security fix. All three CORS header locations removed across both runtimes, solid regression test covering the Node path.

Merged! 🎉

I'll send a follow-up PR to `next` addressing the remaining hardening items:
- Host header validation (DNS rebinding mitigation)
- Bun runtime CORS regression test
- Replace dead OPTIONS 204 with 405
- Use OS-assigned port in tests to avoid CI flakiness

PR #296, opened 2026-04-16
fix(insight): terminate dashboard when Claude/MCP exits

flightlesstux:
## Summary
- ctx_insight dashboard was spawned `detached: true` + `unref()` and never tracked, so it kept running after Claude closed
- Track the child in `_insightChild` and SIGTERM it from `shutdown()` (also kills prior child on re-invocation)
- Pass `INSIGHT_PARENT_PID` and add a 5s `process.kill(ppid, 0)` watchdog in `insight/server.mjs` for SIGKILL / crash / prior-session-orphan paths

## Why
User expectation: closing Claude should close the dashboard. Current behavior leaked a node process + bound port across sessions.

## Test plan
- [ ] Launch Claude, run `ctx_insight`, confirm dashboard up
- [ ] Exit Claude cleanly, confirm port 4747 free within ~1s
- [ ] `kill -9` the MCP process, confirm dashboard self-exits within ~5s (watchdog)
- [ ] Re-run `ctx_insight` in same session, confirm old child killed before new one spawns

### DIFF
--- a/insight/server.mjs
+++ b/insight/server.mjs
@@ -637,6 +637,19 @@ if (isBun) {
   server.listen(PORT, "127.0.0.1");
 }
 
+// Parent watchdog: exit when the MCP process that spawned us disappears.
+// Fallback for SIGKILL / crash paths where shutdown() cannot run.
+const PARENT_PID = Number(process.env.INSIGHT_PARENT_PID);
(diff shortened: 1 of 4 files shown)


mksglu:
Thanks — clean child tracking + SIGTERM on shutdown, plus the INSIGHT_PARENT_PID watchdog covers SIGKILL/crash paths. Merged!

PR #297, opened 2026-04-16
ci(windows): cache GOCACHE and raise Go test timeouts

flightlesstux:
## Summary
- Windows CI fails intermittently with `Test timed out in 30000ms` on `tests/executor.test.ts > Go Execution > Go: hello world` — cold GOCACHE on the Windows runner blows past vitest's default 30s.
- Cache GOCACHE + GOMODCACHE across runs via `actions/cache`, keyed per matrix OS. Paths resolved with `go env` so it works on all three OSes.
- Raise the two Go execution tests to a 90s timeout as a safety net for fresh caches / new runner images.

## Test plan
- [ ] CI green on ubuntu / macos / windows
- [ ] Re-run Windows job with cache hit — Go tests complete well under 90s
- [ ] Cold-cache run (key miss) — still passes via the raised timeout

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -32,6 +32,25 @@ jobs:
           go-version: "stable"
           cache: false
 
+      # Cache GOCACHE across runs so the inline Go snippets tests execute
+      # do not pay the cold-compile cost every run (Windows is ~30s).
+      - name: Resolve Go cache paths
+        id: go-cache-paths
+        shell: bash
+        run: |
+          echo "gocache=$(go env GOCACHE)" >> "$GITHUB_OUTPUT"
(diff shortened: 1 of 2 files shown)


mksglu:
Hey @flightlesstux, thanks for looking into the Windows CI flakes! 🙏

After reviewing, we've decided not to merge this one. The Go executor tests are just 3 small snippet tests (hello world, loops, slices) — there's zero Go source code in the project. Adding GOCACHE/GOMODCACHE caching infrastructure across all 3 OS runners for 3 snippet tests is more complexity than the problem warrants.

If the timeout flake comes back, we'll likely just bump the timeout inline (2-line change) rather than adding cache infrastructure.

Appreciate the contribution — hope to see you on future PRs!

PR #313, opened 2026-04-20
fix(routing): scope guidance throttle by session, not process.ppid (#298)

### LINKED ISSUE #298: guidanceOnce() throttle does not work on Windows — guidance injected on every tool call instead of once per session
### Problem

The `guidanceOnce()` function in `hooks/core/routing.mjs` uses `process.ppid` to create session-scoped marker directories in `/tmp/`. On Windows with Git Bash, **each hook invocation spawns a new `bash.exe` process**, causing `process.ppid` to differ on every call. This means:

1. The file-based throttle marker lands in a new directory each time
2. The in-memory `Set` is always empty (new Node.js process per invocation)
3. **Result**: Read/Bash/Grep guidance tips are injected on EVERY tool call, not once per session

### Evidence

- **716 stale marker directories** accumulated in `/tmp/context-mode-guidance-*` over ~3 days
- Within a single 2-minute window, **5 different PIDs** were observed (25712, 20372, 8512, 8648, 17724)
- `node -e "process.ppid"` run from Bash returns a different PID than what the hook's Node.js process sees

### Impact

- Every Read call gets a "use ctx_execute_file instead" system reminder (~200 chars)
- Every Bash call gets a "use ctx_batch_execute instead" system reminder (~200 chars)
- Every Grep call gets a "use ctx_execute instead" system reminder (~200 chars)
- In a 50+ tool call session, this adds 10KB+ of repetitive context noise
- **User-reported symptom**: Claude stops mid-task after reading files — the conflicting "use a different tool" advice disrupts multi-step planning

### Environment

- Windows 11 Pro (10.0.26100)
- Shell: Git Bash (bash.exe)
- Claude Code CLI
- Node.js (invoked via `node hooks/pretooluse.mjs` as a hook command)

### Root Cause

In `routing.mjs` line 31:
```javascript
const _guidanceId = process.env.VITEST_WORKER_ID
  ? `${process.ppid}-w${process.env.VITEST_WORKER_ID}`
  : String(process.ppid);
```

On macOS/Linux, the parent process (shell) may persist across hook invocations, keeping `process.ppid` stable. On Windows/Git Bash, a new `bash.exe` is spawned for each hook command, making `process.ppid` unique per invocation.

### Suggested Fix

Use a more stable session identifier. Options:

1. **Environment variable**: Claude Code could set a `CLAUDE_SESSION_ID` env var that persists across hook invocations
2. **Grandparent PID**: Use `process.ppid` of the bash parent (the Claude Code process), though this requires an extra syscall
3. **Startup timestamp**: Use a marker from SessionStart (e.g., write a session ID file in `/tmp/` during `sessionstart.mjs` and read it in `pretooluse.mjs`)
4. **Single file with atomic check**: Instead of per-ppid directories, use a single file per guidance type with a session timestamp

Option 3 seems most pragmatic — `sessionstart.mjs` already runs once per session and could write a stable session ID.

### Workaround

Remove `Read` and `Grep` from the PreToolUse matcher in `settings.json` (they aren't in context-mode's default `hooks.json` PreToolUse entries anyway). This eliminates the most disruptive guidance while keeping Bash routing (curl/wget blocking) functional.

---

### Update (2026-04-17)

- **Bug confirmed still present in context-mode v1.0.75** (latest at time of writing) at `hooks/core/routing.mjs:31`:
  ```javascript
  const _guidanceId = process.env.VITEST_WORKER_ID
    ? `${process.ppid}-w${process.env.VITEST_WORKER_ID}`
    : String(process.ppid);
  ```
- **Claude Code 2.1.104** also affected (verified on Windows 11 Pro 10.0.26100, Git Bash).
- **Workaround in continuous use for 8 days** — `Read`/`Grep` removed from PreToolUse matcher, no regressions observed. `Bash` routing (curl/wget scheme blocking, etc.) remains functional.
- **Supplementary mitigation**: a SessionStart hook (`session-cleanup.sh`) was added locally to purge stale `/tmp/context-mode-guidance-*` directories, preventing the accumulation symptom. This treats the symptom, not the root cause — the upstream fix is still needed for correct throttle behavior.

ousamabenyounes:
## What

`guidanceOnce()` in `hooks/core/routing.mjs` scoped its on-disk marker directory by `process.ppid`. On Windows + Git Bash each hook invocation spawns a fresh `bash.exe` with a different PID, so the marker directory name changed every call, the throttle never fired, and Read/Bash/Grep guidance was injected on every tool use — 716 stale marker dirs in `/tmp` and ~10 KB of repetitive context noise per session, per #298.

This PR threads the stable `sessionId` from the hook payload through `routePreToolUse` and names the marker directory `context-mode-guidance-s-<sessionId>` instead. The legacy ppid-based path stays as a backward-compatible fallback for callers that don't pass a session identifier.

Fixes #298.

## Why sessionId, not just a patched ppid

- Every adapter hook (`claude-code`, `codex`, `cursor`, `gemini-cli`, `kiro`, `vscode-copilot`) already reads hook input and has `getSessionId(input, <PLATFORM_OPTS>)` available via `session-helpers.mjs`. It's derived from the hook payload's `transcript_path` / `conversation_id` / `sessionId` / `session_id` / env var, with a final `pid-${ppid}` fallback. That identifier is stable across hook invocations within the same logical session.
- Passing it through `routePreToolUse` is additive (optional 5th arg) — no behavior change for callers that don't pass one, and no need to rely on platform-specific env vars or extra syscalls.
- The alternatives from the issue (grandparent PID, writing a session marker from `sessionstart.mjs`) all required more moving parts to reach the same fix.

## Changes

**`hooks/core/routing.mjs`**
- `_guidanceId` is no longer pinned at module load — `guidanceDirFor(sessionId)` resolves the marker dir per call, using `s-<sessionId>` when one is passed, falling back to the ppid suffix otherwise.
- `guidanceOnce(type, content, sessionId?)` and `resetGuidanceThrottle(sessionId?)` accept the optional session id.
- `routePreToolUse(toolName, toolInput, projectDir, platform, sessionId?)` threads it into the three `guidanceOnce` call sites (Bash/Read/Grep).

**Six adapter hooks** now import the matching `*_OPTS` (or none for Ora Studio default) and pass `getSessionId(input, ...)` at the routing call site:
- `hooks/pretooluse.mjs` (claude-code)
- `hooks/codex/pretooluse.mjs`
- `hooks/cursor/pretooluse.mjs`
- `hooks/gemini-cli/beforetool.mjs`
- `hooks/kiro/pretooluse.mjs`
- `hooks/vscode-copilot/pretooluse.mjs`

**Scope note:** `src/openclaw-plugin.ts` and `src/opencode-plugin.ts` are intentionally not touched. Both run in-process across hook invocations, so `process.ppid` was never unstable for them and the ppid fallback stays correct.

## Coverage added (added to `tests/guidance-throttle.test.ts`, per CONTRIBUTING's "no new test files" rule)

5 regression cases in a new `describe("sessionId scoping (#298 — stable across shifting ppids)")`:

- `second call with same sessionId is throttled even when in-memory Set is cleared` — simulates the Windows Git Bash case (fresh process per hook) and asserts the on-disk marker still blocks the second call.
- `different sessionIds get independent throttles` — two sessions, same process, each fires its own guidance exactly once.
- `sessionId routing is immune to process.ppid changes` — plants a marker in the legacy ppid dir; the sessionId-scoped call ignores it.
- `resetGuidanceThrottle(sessionId)` clears the session-scoped dir.
- `no sessionId passed → falls back to ppid-based behavior (backward compat)` — the previously existing contract is preserved for in-process callers.

Subprocess-based tests (`tests/hooks/integration.test.ts`, `tests/hooks/vscode-hooks.test.ts`) now also clean `context-mode-guidance-s-pid-<testpid>` in their `beforeEach` so the new marker location doesn't leak between tests.

## Test plan

- [x] `npx vitest run tests/guidance-throttle.test.ts` → 13/13 pass (5 new + 8 existing)
- [x] `npx vitest run tests/hooks/integration.test.ts tests/hooks/vscode-hooks.test.ts tests/guidance-throttle.test.ts` → 67/67 pass
- [x] `npm test` → 1600 pass, 23 skipped, **0 failures** — including the subprocess JSON-parse cases in `integration.test.ts` that were flaking on `next`
- [x] `npm run typecheck` → clean

No runtime behavior changes for macOS/Linux users. The fix activates when a caller passes a session id — which all six adapter hooks now do.

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/hooks/codex/pretooluse.mjs
+++ b/hooks/codex/pretooluse.mjs
@@ -10,7 +10,7 @@ import "../suppress-stderr.mjs";
 
 import { dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
-import { readStdin, getInputProjectDir, CODEX_OPTS } from "../session-helpers.mjs";
+import { readStdin, getInputProjectDir, getSessionId, CODEX_OPTS } from "../session-helpers.mjs";
 import { routePreToolUse, initSecurity } from "../core/routing.mjs";
 import { formatDecision } from "../core/formatters.mjs";
 
@@ -23,7 +23,7 @@ const tool = input.tool_name ?? "";
 const toolInput = input.tool_input ?? {};
 const projectDir = getInputProjectDir(input, CODEX_OPTS);
 
-const decision = routePreToolUse(tool, toolInput, projectDir, "codex");
+const decision = routePreToolUse(tool, toolInput, projectDir, "codex", getSessionId(input, CODEX_OPTS));
 const response = formatDecision("codex", decision);
 const output = response ?? {
   hookSpecificOutput: { hookEventName: "PreToolUse" },
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -36,23 +36,41 @@ import { resolve } from "node:path";
 //   - In-memory Set for same-process (OpenCode ts-plugin, vitest)
 //   - File-based markers with O_EXCL for cross-process atomicity
 //     (Claude Code, Gemini, Cursor, VS Code Copilot)
-// Session scoped via process.ppid (= host PID, constant for session lifetime).
+//
+// Session identity is resolved in this order:
+//   1. sessionId passed in by the caller (stable across hook invocations)
+//   2. process.ppid fallback (works on macOS/Linux — host PID is stable)
+//
+// The ppid fallback is unreliable on Windows + Git Bash, where each hook
+// invocation spawns a fresh bash.exe with a different PID (#298). Callers
+// that have a stable session identifier (e.g. from the hook payload) should
+// pass it to routePreToolUse so the marker directory stays consistent across
+// invocations of the same logical session.
 const _guidanceShown = new Set();
-const _guidanceId = process.env.VITEST_WORKER_ID
-  ? `${process.ppid}-w${process.env.VITEST_WORKER_ID}`
-  : String(process.ppid);
-const _guidanceDir = resolve(tmpdir(), `context-mode-guidance-${_guidanceId}`);
 
-function guidanceOnce(type, content) {
+function defaultGuidanceId() {
+  return process.env.VITEST_WORKER_ID
+    ? `${process.ppid}-w${process.env.VITEST_WORKER_ID}`
+    : String(process.ppid);
+}
+
+function guidanceDirFor(sessionId) {
+  const id = sessionId ? `s-${sessionId}` : defaultGuidanceId();
+  return resolve(tmpdir(), `context-mode-guidance-${id}`);
+}
+
+function guidanceOnce(type, content, sessionId) {
   // Fast path: in-memory (same process)
   if (_guidanceShown.has(type)) return null;
 
-  // Ensure marker directory exists
-  try { mkdirSync(_guidanceDir, { recursive: true }); } catch {}
+  // Resolve marker directory for this session (stable even on Windows/Git Bash
+  // where process.ppid shifts every invocation — see #298).
+  const dir = guidanceDirFor(sessionId);
+  try { mkdirSync(dir, { recursive: true }); } catch {}
 
   // Atomic create-or-fail: O_CREAT | O_EXCL | O_WRONLY
   // First process to create the file wins; others get EEXIST.
-  const marker = resolve(_guidanceDir, type);
+  const marker = resolve(dir, type);
   try {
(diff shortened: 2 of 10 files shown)


mksglu:
Thanks @ousamabenyounes — threading the stable sessionId through routePreToolUse is the right approach. Clean backward-compat fallback, all 6 adapter hooks updated, solid test coverage. Merged!

Issue #311, opened 2026-04-20
Zombie MCP server processes consume 50-60% CPU each after session close


camhoccode:
**Platform**: macOS 14.5 (Darwin 23.5.0), ARM64, Node.js v21.7.3
**Claude Code**: VSCode extension

---

## Summary

When a Claude Code session closes, context-mode MCP server processes become orphaned and enter an infinite error loop consuming 50-60% CPU each. Multiple sessions accumulate zombie processes, leading to sustained >100% CPU usage.

## Steps to Reproduce

1. Open Claude Code (VSCode extension)
2. Use context-mode normally (any tool call triggers it)
3. Close the Claude Code session (close panel, close VSCode, or Ctrl+C)
4. Open a new Claude Code session
5. Repeat steps 2-4

## Expected Behavior

Old context-mode processes should terminate when the parent Claude Code session dies.

## Actual Behavior

Each closed session leaves behind ~6 orphaned processes. The main MCP server worker enters an infinite error loop:

```
PID 22095: 50.5% CPU, 53MB RSS, running 1h46m (session closed long ago)
PID 24550: 53.1% CPU, 53MB RSS, running 1h31m
PID 38682: 58.6% CPU, 57MB RSS, running 3m (current session — also high CPU)
```

Total: ~160% CPU for 3 sessions, doing nothing useful.

### 1. Missing `stdin` end handler in StdioServerTransport

The MCP SDK transport (`_a` class in `server.bundle.mjs`) only listens for `data` and `error` events:

```js
start() {
  this._stdin.on("data", this._ondata);
  this._stdin.on("error", this._onerror);
}
```

No `end` or `close` handler. When parent dies and stdin EOF occurs, the transport doesn't initiate shutdown.

### 2. Watchdog bypassed by npm wrapper layer

Process tree:
```
[Claude Code (dead)] → start.mjs → npm exec (PPID=1, orphan but alive) → context-mode server
```

The watchdog (`qC()`) checks:
```js
var HC = process.ppid;  // saved at startup
function qC() {
  let t = process.ppid;
  return !(t !== HC || t === 0 || t === 1);
}
```

But the server's direct parent is `npm exec` (still alive), not Claude Code. So `process.ppid` never changes → watchdog never triggers.

### 3. Infinite error loop (CPU profile evidence)

`sample` profiling of zombie PID 22095 (1 second, 791 samples):

```
541/791 (68%) → ReportPendingMessages
  → ErrorStackGetter
[...]

PR #315, opened 2026-04-20
fix(lifecycle): shut down when grandparent reparents to init (#311)

ousamabenyounes:
## What

Partial fix for the zombie MCP server problem described in #311. The server previously only monitored its direct ppid; this PR also watches the grandparent and shuts down when it reparents to init while the chain was originally not rooted at init.

Fixes the common case described in #311 on macOS/Linux. The second half of the issue (the vendored MCP SDK missing a stdin `end` handler) is deliberately out of scope — see the \"Not fixed here\" section below.

## Why the existing guard wasn't enough

The Ora Studio process chain is:

```
Ora Studio → start.mjs → npm exec → context-mode server
```

When Ora Studio dies:
- `start.mjs` reparents to init (PID 1)
- `npm exec` stays alive, so `context-mode server`'s direct `ppid` never changes
- The existing `defaultIsParentAlive` is still happy, the 30-second ppid tick never flips, and the server spins forever.

Extending the check to \"is the grandparent still a real process?\" catches exactly this case. The grandparent IS `start.mjs`, and `start.mjs`'s ppid flips to 1 the moment Ora Studio dies.

## Why not stdin

The issue suggests adding stdin `end`/`close` handlers. #236 already removed those because the MCP stdio transport owns stdin and transient pipe events caused spurious `-32000` errors. The repo locks that in with three tests in `tests/lifecycle.test.ts`:

- `cleanup does NOT remove stdin listeners (stdin is not used)`
- `startLifecycleGuard does NOT call process.stdin.resume()`
- `child does NOT exit when stdin is closed (#236)`

All three continue to pass — this PR adds zero stdin listeners.

## Changes

**`src/lifecycle.ts`**
- New `readGrandparentPpidImpl()` helper: runs `ps -o ppid= -p $PPID` on POSIX, returns `NaN` on Windows or probe failure.
- New exported factory `makeDefaultIsParentAlive(deps)` with injectable `getPpid` and `readGrandparentPpid` — used by production (`const defaultIsParentAlive = makeDefaultIsParentAlive()`) and by tests with fake readers.
- The default check now additionally returns `false` when the grandparent is now PID 1 but wasn't at startup.

**Safety net for false positives:**
- If the original grandparent was already 1 (daemon / nohup / launchd startup), the grandparent check is skipped entirely.
- On Windows (`ps` unavailable) or if the `ps` probe fails, the reader returns `NaN` and the check is also skipped. So the behavior on Windows is unchanged.

## Not fixed here (out of scope)

#311 also reports that the MCP SDK's `StdioServerTransport` lacks an `end` / `close` handler inside `server.bundle.mjs`. That's vendored third-party code, and #236 explicitly established that we don't drive shutdown via stdin listeners in this process. Addressing that root cause belongs in the SDK bundle (or a separate, coordinated change around #236's constraints).

## Tests added (`tests/lifecycle.test.ts`)

5 regression cases in a new `describe(\"makeDefaultIsParentAlive — grandparent orphan detection (#311)\")`, using the factory with injected readers so each scenario is hermetic:

- `returns false when grandparent is reparented to init after startup` — core regression case.
- `does not false-positive when grandparent was already init at startup` — daemon-start tolerance.
- `tolerates NaN grandparent (Windows / ps failure)` — falls back to original behavior.
- `direct ppid death still takes precedence over grandparent check` — existing contract preserved.
- `grandparent check kicks in through startLifecycleGuard end-to-end` — wiring verification via the public API.

## Test plan

- [x] `npx vitest run tests/lifecycle.test.ts` → 14/14 pass (9 existing + 5 new)
- [x] `npm test` → 1600 pass, 23 skipped, **0 failures**
- [x] `npm run typecheck` → clean
- [x] `npm run build` → clean

Generated by Ora Studio
Vibe coded by ousamabenyounes

### DIFF
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -10,6 +10,8 @@
  * Cross-platform: macOS, Linux, Windows.
  */
 
+import { execFileSync } from "node:child_process";
+
 export interface LifecycleGuardOptions {
   /** Interval in ms to check parent liveness. Default: 30_000 */
   checkIntervalMs?: number;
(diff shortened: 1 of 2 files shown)


mksglu:
Hi @ousamabenyounes! That's great. Did you fully-tested that? (I'm know you'd already created a some Unit tests) I'm gonna merge to the `next` branch.

ousamabenyounes:
Thanks @mksglu! Test status, to set expectations cleanly:

**Unit-level (hermetic, deterministic):** 5 new cases in `tests/lifecycle.test.ts` under `describe("makeDefaultIsParentAlive — grandparent orphan detection (#311)")`. They use the factory with injected ppid/grandparent readers so each scenario is isolated:

- grandparent flipping from a real PID to `1` post-startup → returns `false` ✅
- grandparent already `1` at startup (daemon / launchd / `nohup` case) → never flips, no false positive ✅
- `NaN` grandparent (Windows / `ps` probe failure) → falls back to original ppid-only behavior ✅
- direct-ppid death takes precedence (short-circuit before grandparent check) ✅
- end-to-end through `startLifecycleGuard` with injected reader, actual shutdown path fires ✅

```
$ npx vitest run tests/lifecycle.test.ts
Test Files  1 passed (1)
Tests  14 passed (14)
```

Full suite: `npm test` → 1600 pass, 23 skipped, **0 failures**. Typecheck clean. Build clean. The existing `#236`-era tests (`cleanup does NOT remove stdin listeners`, `does NOT call process.stdin.resume()`, `child does NOT exit when stdin is closed`) all still pass — this PR adds zero stdin listeners.

**What I didn't add:** a real 3-level `spawn → spawn → spawn → kill-grandparent → assert grandchild exits` harness. That'd need a POSIX-only helper (same `describe.skipIf(isWindows)` pattern the repo already uses for #236) and a few hundred ms of sleep to hit the real 30s guard tick with a lowered `checkIntervalMs`. Happy to prototype one if you want it before merging — otherwise the current coverage is "logic-deterministic + wiring-verified", and a one-time manual repro on your side post-merge (VSCode → Claude Code → kill the VSCode process → `ps aux | grep context-mode` 30s later) would close the loop cheaply.

Your call on the integration test. Ready to ship either way.

mksglu:
Thanks — clean DI pattern, solid edge-case handling (daemon tolerance, Windows/NaN fallback), and the 5 hermetic tests cover every branch. Merged!

Issue #231, opened 2026-04-06
better-sqlite3 NODE_MODULE_VERSION mismatch after Node.js upgrade requires manual npm rebuild


benzntech:
## Bug

When the plugin is installed or used after upgrading Node.js, `better-sqlite3`'s precompiled native addon fails to load with the following error:

```
plugin:context-mode:context-mode - ctx_fetch_and_index (MCP)
Error: The module '/Users/developer/.claude/plugins/marketplaces/context-mode/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using NODE_MODULE_VERSION 127.
This version of Node.js requires NODE_MODULE_VERSION 141.
Please try re-compiling or re-installing the module (for instance, using `npm rebuild` or `npm install`).
```

## Steps to Reproduce

1. Install context-mode plugin
2. Upgrade Node.js to a newer major version
3. Trigger any MCP tool (e.g. `ctx_fetch_and_index`)

## Workaround

Manually rebuild the native module:

```sh
cd ~/.claude/plugins/marketplaces/context-mode
npm rebuild better-sqlite3
```

## Suggested Fix

The plugin could detect a `NODE_MODULE_VERSION` mismatch at startup and automatically run `npm rebuild better-sqlite3`, or surface a clearer error message with the workaround command.

This is also related to #228 — both issues stem from `better-sqlite3` being a native addon that is tightly coupled to the Node.js version. Adopting a pure-JS or built-in SQLite backend (e.g. `node:sqlite`) would eliminate this class of errors entirely.

## Environment

- **OS:** macOS (arm64)
- **Node.js:** v25.9.0 (NODE_MODULE_VERSION 141)
- **context-mode:** v1.0.59
- **better-sqlite3:** compiled against NODE_MODULE_VERSION 127

mksglu:
Please update your Context Mode version and then try re-repro. Thank you.

benzntech:
## Follow-up: root cause is Claude Code's internal Node.js vs system Node.js

After upgrading to v1.0.75, the upgrade script now auto-rebuilds `better-sqlite3` ("Native addons rebuilt" in the output). However, the **root cause** is worth noting explicitly:

Claude Code runs the MCP/plugin server using its own internal Node.js runtime, which may differ from the system `node` binary used by `npm rebuild`. This means:

[...]

Issue #206, opened 2026-03-31
FTS5/better-sqlite3: native binding missing on macOS global npm install with ignore-scripts=true


dipenparmar12:
## Environment
- **Platform**: Claude Code (macOS)
- **context-mode version**: 1.0.50
- **OS**: macOS (arm64-apple-darwin25, Apple Silicon)
- **Node.js**: v22.22.0
- **npm config**: ignore-scripts = true

## Problem
Better-sqlite3 native binding (.node file) fails to build during global npm install when npm has `ignore-scripts=true` configured globally. This results in FTS5 search feature being completely unavailable.

## Doctor Output (Failed State)
```
■  FTS5 / better-sqlite3: FAIL — Could not locate the bindings file. Tried:
   → /path/to/better_sqlite3.node
   → /path/to/build/Release/better_sqlite3.node
   [... all prebuild paths ...]
   Try: npm rebuild better-sqlite3

└  Diagnostics failed — 1 critical issue(s) found
```

## Steps to Reproduce
```bash
# 1. Set npm to ignore lifecycle scripts (common security practice)
npm config set ignore-scripts true

# 2. Install context-mode globally
npm install -g context-mode@latest

# 3. Run diagnostics
context-mode doctor
```

Expected: FTS5 check passes
Actual: FTS5 FAIL — native binding not found

## Root Cause Analysis
1. `npm ignore-scripts=true` prevents ALL lifecycle scripts (preinstall, install, postinstall, etc.)
2. better-sqlite3 requires native compilation (node-gyp or prebuild-install) during npm install
3. With scripts disabled: package installs but .node binary is never created
4. Runtime fails when database tries to load missing native binding
5. Doctor command doesn't invoke startup bootstrap that might self-heal this

## Confirmed Workaround
```bash
npm rebuild better-sqlite3 --ignore-scripts=false --foreground-scripts
context-mode doctor  # Now shows FTS5 PASS
```

## Investigation Completed
- ✅ Verified package installs but .node file missing
- ✅ Confirmed this only happens with npm ignore-scripts=true
- ✅ Reproduced in clean global install under same npm config
- ✅ Validated npm rebuild with --ignore-scripts=false fixes it
- ✅ Confirmed this is context-mode's responsibility to handle (optional dep + claims auto-heal)

## Proposed Solution
Enhance `hooks/ensure-deps.mjs` to:
1. Detect missing native binding (not just missing directory)
[...]

PR #318, opened 2026-04-21
Fix better-sqlite3 native binary detection during startup

TyceHerrman:
## Motivation

`context-mode` already tries to self-heal native dependency problems at startup, but the current check can falsely conclude that `better-sqlite3` is usable when only `build/Release/` or `prebuilds/` exists.

This is also a follow-up to #206. That issue added startup dependency self-healing for cases where native dependencies are installed without their native build artifacts. The current implementation catches the fully missing-package case, but its native-addon presence check is too weak: it treats `build/Release/` or `prebuilds/` as enough evidence that `better-sqlite3` is usable.

This PR tightens the #206 check from “does a native-artifact directory exist?” to “does the actual `better_sqlite3.node` binary exist?” It also applies the same rule to the ABI compatibility path, so both startup repair paths handle the same missing-binary state consistently.

This also relates to #231: both failures come from `better-sqlite3` being a native addon tied to the local Node ABI. This PR does not replace the SQLite backend, but it makes startup repair more reliable for the missing-binary case.

## Changes

- Check for the actual native addon file: `node_modules/better-sqlite3/build/Release/better_sqlite3.node`
- Rebuild `better-sqlite3` when the package directory exists but the `.node` file is absent
- Update the ABI/native compatibility path so an existing `build/Release/` directory without `better_sqlite3.node` triggers rebuild instead of returning early
- After rebuild, codesign on macOS and probe-load before caching the ABI-specific binary
- Keep the startup model unchanged: `start.mjs` still imports `hooks/ensure-deps.mjs` before loading `server.bundle.mjs`

## Verification

- `npm test -- tests/hooks/ensure-deps.test.ts`
- `npm run typecheck`
- `npm run build`

drafted with assistance from Codex


mksglu:
Hi @TyceHerrman! We're having conflict issues.

mksglu:
## Thorough Review Complete ✅

Reviewed by 6 parallel agents (Staff Engineer, OS Compatibility, Database Architecture, QA, Security, Competitor Analysis). All approve.

### What this PR fixes (4 real bugs)

1. **Phantom directory false positive** — empty `build/Release/` dir fooled the old check into skipping rebuild. Now checks the actual `.node` binary file.
2. **Missing binary silent bail-out** — `ensureNativeCompat` silently returned when binary was missing instead of attempting rebuild. Now falls through to rebuild path.
3. **Silent rebuild skip** — users with `ignore-scripts=true` in .npmrc had `npm rebuild` silently skip native compilation. `--ignore-scripts=false` forces it.
4. **Cache poisoning prevention** — post-rebuild now validates via `probeNativeInChildProcess()` before caching. Prevents broken binaries from poisoning the ABI cache.

### Security assessment
`--ignore-scripts=false` is LOW risk — it was already used in `ensureDeps()`, this PR adds it to `ensureNativeCompat()` for consistency. Only runs better-sqlite3's standard `prebuild-install || node-gyp rebuild` script.

### Cross-platform verification
Verified correct on macOS (codesign untouched), Linux (.so paths correct), and Windows (.node paths + backslash handling correct). CI green on all 3 OS.

### Strategic note
Our competitor analysis shows context-mode is the only AI coding tool still shipping `better-sqlite3` native binaries. OpenCode uses `bun:sqlite` + `node:sqlite` only. Codex uses Rust's `rusqlite`. We're planning to drop `better-sqlite3` entirely in a future release and require Node >= 22.5 or Bun — eliminating ~200 lines of bootstrap code and the entire native binary problem. Until then, your fixes make the existing pipeline significantly more robust.

Thank you @TyceHerrman for the solid contribution! 🙏

mksglu:
## Merged into next (cherry-picked) ✅

Your fixes have been cherry-picked onto `next` with the `hasModernSqlite()` gate preserved (our #331 SIGSEGV prevention, which your branch predated).

All 4 bug fixes landed:
1. **Phantom directory false positive** — `NATIVE_BINARIES` map + file-level check
2. **Missing binary silent bail-out** — falls through to rebuild now
3. **`--ignore-scripts=false`** on ABI rebuild path
4. **Probe-before-cache** — prevents cache poisoning

We also added your new tests (binary-missing path, graceful degradation) plus updated the existing test harness to match.

**Strategic note**: Our competitor analysis shows context-mode is the only AI coding tool still shipping `better-sqlite3` native binaries. We're planning to drop it entirely in a future release and require Node >= 22.5 or Bun — eliminating ~200 lines of bootstrap code. Until then, your fixes make the pipeline significantly more robust.

Thank you @TyceHerrman for the excellent contribution! 🙏

Commit: d328c24

PR #320, opened 2026-04-22
Fix Go runtime version reporting

TyceHerrman:
## Motivation

`getRuntimeSummary()` uses a generic `--version` probe for runtime version reporting. Go does not support `go --version`; it exits with a flag error and prints help text. The supported command is `go version`.

That means context-mode can correctly detect Go as available, but still report its version as `unknown`.

## Changes

- Update the private runtime version helper to accept explicit version arguments.
- Use `go version` for Go runtime summary output.
- Keep the existing `--version` default for other runtimes.
- Use `execFileSync` with Windows shell compatibility for safer argument handling.
- Add mocked unit coverage to verify Go uses `version` and other runtimes still use `--version`.

## Verification

- `npm test -- tests/runtime.test.ts`
- `npm run typecheck`
- `npm run build`
- Local sanity:
  - `go --version` fails with `flag provided but not defined: -version`
  - `go version` prints `go version go1.26.2 darwin/arm64`


mksglu:
Resolved the server.bundle.mjs conflict and merged to next. Thanks for the Go version fix!

Issue #331, opened 2026-04-23
SIGSEGV coredumps from better-sqlite3 probe in ensure-deps.mjs on Node v24 (gap left by #228)


Alexey-Goru1ev:
## Summary

`hooks/ensure-deps.mjs` reliably produces `SIGSEGV` coredumps on Node v24 (and periodically on Node v22) on Linux. This is despite the ABI-aware caching + child-process probe introduced to mitigate #148 / #203 / #228. The parent hook recovers via `try/catch`, so the tool continues functioning — but `systemd-coredump` captures every crash and users see a growing pile of coredumps for what looks like normal operation.

**This is the same underlying SIGSEGV that #228 was about — closing #228 was premature; the mitigation does not actually eliminate the crashes, it just hides them from the parent process.**

## Environment

- `node -v` → `v24.14.1` (reproducible on NVM-managed Node 24.x)
- `uname -sr` → `Linux 7.0.0-070000-generic`
- `context-mode` plugin cache version: `1.0.89` (from `~/.claude/plugins/cache/context-mode/context-mode/1.0.89/`)
- Claude Code, MCP stdio transport

## Reproduction

1. Install `context-mode` 1.0.89 into a Claude Code session running on Node v24.14.1 (Linux).
2. Start a session — `hooks/ensure-deps.mjs` runs as an auto-import side effect via `SessionStart` / `PreToolUse` hooks.
3. `ensureNativeCompat()` calls `probeNativeInChildProcess()` which spawns:
   ```
   node -e "new (require('better-sqlite3'))(':memory:').close()"
   ```
4. The child process SIGSEGVs inside `better-sqlite3`'s native addon during `dlopen` / first-DB instantiation.
5. The parent `execSync` catches the non-zero exit and continues. From the user's perspective the hook runs fine.
6. **But** `systemd-coredump` still captures the child's SIGSEGV:

<details>
<summary>Actual coredump (2026-04-23 14:31 +03 local, sanitized)</summary>

```
           PID: 268649 (MainThread)
           UID: 1000
        Signal: 11 (SEGV)
     Timestamp: Thu 2026-04-23 14:31:32 +03
  Command Line: node -e $'new (require(\'better-sqlite3\'))(\':memory:\').close()'
    Executable: /home/alexey/.nvm/versions/node/v24.14.1/bin/node
 Control Group: /user.slice/user-1000.slice/user@1000.service/app.slice/kitty-4183-0.scope
[...]

Issue #148, opened 2026-03-19
[Bug]: better-sqlite3 NODE_MODULE_VERSION mismatch with concurrent sessions using different Node versions (mise/asdf)


husniadil:
### Platform

Claude Code

### context-mode version

1.0.25

### Doctor output

```
Server test: PASS
PreToolUse hook: PASS
SessionStart hook: PASS
Hook script exists: PASS
Plugin enabled: PASS
FTS5 / better-sqlite3: PASS (from Node 24 session)
FTS5 / better-sqlite3: FAIL (from Node 20 session — same machine, concurrent)
```

### Prompt that triggered the bug

Starting Claude Code from two different projects — one with Node 24 (via mise `.tool-versions`), another with Node 20 (global default). The session started second gets `NODE_MODULE_VERSION` mismatch.

### What happened?

When using `mise` with different Node versions per project, concurrent Claude Code sessions cause `NODE_MODULE_VERSION` mismatch on `better-sqlite3`. The plugin cache has a single native binary, but each session resolves a different Node ABI version.

- Node 24.11.0 = ABI 137
- Node 20.19.4 = ABI 115

Whichever session starts first (or last rebuild) "wins". The other session gets:

```
The module '.../better_sqlite3.node' was compiled against a different
Node.js version using NODE_MODULE_VERSION 137. This version of Node.js
requires NODE_MODULE_VERSION 115.
```

v1.0.25's auto-rebuild on upgrade helps for single-session usage but doesn't solve concurrent cross-version sessions.

### Expected behavior

context-mode should work across concurrent sessions regardless of which Node version each session uses.

### Steps to reproduce

1. Install `mise` with Node 20.x and Node 24.x
2. Project A: `.tool-versions` with `node 24.x`
3. Project B: uses global Node 20.x (or its own `.tool-versions`)
4. Launch Claude Code in Project A → context-mode works
5. Launch Claude Code in Project B concurrently → FTS5/better-sqlite3 FAIL

### Setup details

- [x] MCP server registered
- [x] Hooks configured (PreToolUse, PostToolUse, SessionStart, etc.)
- [x] Routing instruction file present (CLAUDE.md / GEMINI.md / AGENTS.md / copilot-instructions.md)

### Debug logs

```
Error: The module '/Users/xxx/.claude/plugins/cache/context-mode/context-mode/1.0.20/node_modules/better-sqlite3/build/Release/better_sqlite3.node'
was compiled against a different Node.js version using
[...]

PR #332, opened 2026-04-23
fix(hooks): skip better-sqlite3 bootstrap on Node >=22.5 / Bun

Alexey-Goru1ev:
On modern runtimes a built-in SQLite binding exists: `node:sqlite` from Node v22.5+ and `bun:sqlite` on Bun. In those cases the better-sqlite3 bootstrap in `hooks/ensure-deps.mjs` (install / rebuild / child-process probe) is unnecessary.

The child-process probe added in ensureNativeCompat to validate ABI compatibility reliably SIGSEGVs inside better-sqlite3's native addon on Node v24 under V8 GC pressure on Linux -- the parent hook swallows the child's non-zero exit via try/catch, but systemd-coredump still captures the SIGSEGV and users see coredumps in `coredumpctl` on every hook-triggered Claude session. This is the underlying mechanism that motivated #228 (reopened below); the ABI-probe mitigation does not actually eliminate the coredumps, it just hides the user-facing error.

Gate both `ensureDeps()` and `ensureNativeCompat()` with a `hasModernSqlite()` check that early-returns when Bun is present or Node >= 22.5. Legacy Node <22.5 path is untouched -- better-sqlite3 still bootstraps normally where no built-in backend exists, and the ABI-aware caching keeps working for users on older Node.

Verified on Node v24.14.1: both functions complete in <0.01 ms with the gate active (no execSync spawned, no coredump possible from this path). Simulated Node 22.4.0: gate returns false, legacy bootstrap runs unchanged.

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] Gemini CLI
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Checklist

- [ ] `npm test` passes

<details>

</details>

### DIFF
--- a/hooks/ensure-deps.mjs
+++ b/hooks/ensure-deps.mjs
@@ -30,6 +30,17 @@ const root = resolve(__dirname, "..");
 const NATIVE_DEPS = ["better-sqlite3"];
 
 export function ensureDeps() {
+  // Skip bootstrap when a built-in SQLite binding is available: Bun's
+  // `bun:sqlite`, or Node's `node:sqlite` from v22.5+. On those runtimes
+  // the better-sqlite3 install/rebuild is unnecessary, and probing the
+  // native addon SIGSEGVs on Node v24 under V8 GC pressure on Linux —
+  // which writes a coredump even when the parent hook recovers via
+  // try/catch. Inlined rather than factored out so the standalone
+  // extraction in tests/core/cli.test.ts keeps working.
+  if (globalThis.Bun) return;
+  const [nodeMajor, nodeMinor] = process.versions.node.split(".").map(Number);
+  if (nodeMajor > 22 || (nodeMajor === 22 && nodeMinor >= 5)) return;
+
   for (const pkg of NATIVE_DEPS) {
     const pkgDir = resolve(root, "node_modules", pkg);
     if (!existsSync(pkgDir)) {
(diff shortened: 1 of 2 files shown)


Alexey-Goru1ev:
## FYI — force-pushed an amended version after running the CI steps locally

Before opening this for review I reproduced `ci.yml` locally on Node 20. Two things worth flagging.

### 1. Amended commit to fix test compatibility (reason for force-push)

First version factored the gate into a module-scoped `hasModernSqlite()` helper. That failed 6 tests in `tests/core/cli.test.ts > ABI-aware native binary caching (#148)`.

Root cause: `loadEnsureNativeCompat()` at `tests/core/cli.test.ts:219-239` intentionally bypasses module loading — it reads `hooks/ensure-deps.mjs` as text, regex-matches only `export function ensureNativeCompat {...}`, writes that slice to a temp `.mjs`, and dynamically imports it. Identifiers defined outside the function body are undefined at test time.

Fix: inlined the gate inside both functions (2 extra lines each). Runtime semantics byte-identical to the helper version; extracted function now self-contained. Happy to go DRY and update the extraction helper instead if you'd prefer — kept the duplication because the function comments explicitly point at the test file.

Current commit: `d8e0ade`.

### 2. Local test results

`tsc --noEmit`, `npm run build`, `npm run bundle` — all green.

`vitest run` full suite: 1412+ pass, 2 flakes in `tests/hooks/cursor-hooks.test.ts` — pass deterministically in isolation on both `origin/main` and this branch (3 consecutive runs). Same fork-worker-teardown race your `ci.yml` already documents:

```yaml
# Vitest fork workers can crash during cleanup on CI (native addon teardown race)
```

**Data point for #331:** that race is itself `better-sqlite3`'s SIGSEGV firing during Node 20 worker shutdown. This PR doesn't address it — the gate is a no-op on Node <22.5 (CI's matrix). Proper fix is server-side `node:sqlite` migration, broader direction of #228, out of scope here.

mksglu:
Hi @Alexey-Goru1ev Thanks, please resolve conflicts, add unit tests. Use context-mode-ops skill. Follow contrubution.md file. Also the code is duplicate.

mksglu:
Closing — the fix landed directly on `next` (f351fd1) with the same approach you proposed. Consolidated the `hasModernSqlite()` gate into a single shared function to avoid the duplication I flagged earlier.

Thanks @Alexey-Goru1ev for identifying the root cause and the clean write-up with coredump evidence. Your PR was the basis for the final fix.

PR #337, opened 2026-04-24
feat(plugin): unify KiloCode/OpenCode plugin exports and routing

mikij:
- Rename factory to `createContextModePlugin` and add default `{ server }` export for KiloCode
- Update routing call to pass platform string instead of hardcoded "opencode"
- Add compatibility named export `ContextModePlugin` for OpenCode consumers
- Adjust tests to verify both default and named exports

BREAKING CHANGE: Plugin consumers using default export must switch from `ContextModePlugin(ctx)` to `{ server }` shape; named export remains for backward compat.

## What / Why / How

Could not get any saving on context

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] Gemini CLI
- [ ] OpenCode
- [x] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

added test for "default export has KiloCode PluginModule shape { server }" and renamed old one

## Checklist

- [x] `npm test` passes

<details>

</details>


mksglu:
Hi @mikij! Thanks! Why we need that?

mikij:
Honestly I asked model to investigate what is wrong with plugin as it does not work for me as I can notice. It found this and now I am trying to figure out if that will help. I managed to work with your example prompt but still in daily use I don't see prompts go through context-mode for me. It could be that KiloCode made. some changes in plugin architecture that is little different than OpenCodes but not sure. If you have any other suggestion to try instead I am open to try.

mikij:
And after I did that in my local installation I start seeing report that says there are savings again. Before even I see tool is called ctx stats says nothing has been processed 

image

mksglu:
Thanks @mikij — resolved the bundle conflicts and merged to `next`. Typecheck clean.

We'd appreciate it if you could test the KiloCode/OpenCode plugin routing once we cut the next release. We'll tag you when it's out.

Issue #230, opened 2026-04-06
Agent gets stuck when WebFetch is blocked but ctx_fetch_and_index MCP tool is unavailable


benzntech:
## Problem

When `context-mode` blocks `WebFetch` via the `PreToolUse` hook, it instructs the agent to use `ctx_fetch_and_index` instead. However, if the context-mode MCP server is not running (or its tools are not yet registered in the session), the agent has no fallback and gets stuck — repeatedly retrying `WebFetch` and receiving the same blocking error.

## Steps to Reproduce

1. Install context-mode plugin and start a new Claude Code session
2. Ask the agent to fetch a URL (e.g. a GitHub repo page)
3. The `PreToolUse` hook blocks `WebFetch` and tells the agent to use `ctx_fetch_and_index`
4. If the MCP server hasn't started / tools aren't registered yet, the agent retries `WebFetch` and gets blocked again repeatedly

## Observed Behavior

```
PreToolUse:WebFetch hook returned blocking error
context-mode: WebFetch blocked. Use mcp__plugin_context-mode_context-mode__ctx_fetch_and_index(url: "...", source: "...") to fetch this URL in sandbox. Then use mcp__plugin_context-mode_context-mode__ctx_search(queries: [...]) to query results. Do NOT use curl, wget, mcp_web_fetch, or mcp_fetch_tool.
```

The error fires twice in a row and the agent gives up without completing the task.

## Expected Behavior

Either:
- The MCP server should be guaranteed to be started and tools registered before any `WebFetch` block fires, OR
- The hook's deny message should detect MCP tool unavailability and degrade gracefully (e.g. allow the fetch if `ctx_fetch_and_index` is not available), OR
- The `sessionstart.mjs` hook should validate that the MCP server is reachable before the session begins

## Environment

- context-mode version: 1.0.59
- Platform: macOS
- Claude Code with `PreToolUse` hook on matcher: `Bash|WebFetch|Read|Grep|Agent|Task|...`

mksglu:
Please update your Context Mode version and then try re-repro. Thank you.

murataslan1:
Checked `hooks/core/routing.mjs` lines 277-283. The WebFetch deny is fully unconditional:

```javascript
if (canonical === "WebFetch") {
  return {
    action: "deny",
    reason: `context-mode: WebFetch blocked. Think in Code — use ctx_fetch_and_index(...)...`,
  };
}
```

[...]

Issue #166, opened 2026-03-22
curl/wget hook blocks binary downloads with -o flag


hasandenizuk:
## Problem

The PreToolUse hook in `hooks/core/routing.mjs` (line ~169) blocks **all** `curl`/`wget` commands, including binary file downloads that redirect output to a file via `-o`/`--output`. These downloads don't flood the context window since the output goes to a file, not stdout.

**Example blocked command:**
```bash
curl -sL https://github.com/stripe/stripe-cli/releases/download/v1.38.1/stripe_1.38.1_linux_x86_64.tar.gz -o /tmp/stripe.tar.gz
```

This is a common pattern for CLI tool installation (Stripe CLI, Terraform, kubectl, etc.) and should be allowed through.

## Current behavior

The regex on line 169 catches all curl/wget regardless of output destination:
```js
if (/(^|\s|&&|\||\;)(curl|wget)\s/i.test(stripped)) {
```

## Proposed fix

Add an exception for file-output flags before the block:

```js
// Exception: allow curl/wget when output is redirected to a file (-o, --output, -O, --output-document)
// This permits binary downloads (e.g., CLI tool installation) that don't flood context.
const isCurlToFile = /curl\s/.test(stripped) && /\s(-o|--output)\s/.test(stripped);
const isWgetToFile = /wget\s/.test(stripped) && /\s(-O|--output-document)\s/.test(stripped);
if (/(^|\s|&&|\||\;)(curl|wget)\s/i.test(stripped) && !isCurlToFile && !isWgetToFile) {
```

This preserves the block for `curl https://example.com` (stdout output that would flood context) while allowing `curl -o /tmp/file.tar.gz https://...` (file output, no context impact).

## Notes

- The `stripped` variable already has quoted content removed via `stripQuotedContent()`, so the `-o` flag is reliably visible
- Both the marketplace source **and** the cache copy (`plugins/cache/context-mode/.../hooks/core/routing.mjs`) need the fix — the cache is what actually runs
- Tested with both simple (`curl -o /dev/null`) and complex (`curl -sL <url> -o /tmp/file && tar ...`) commands

Happy to submit a PR if preferred.

mksglu:
Fixed in **v1.0.49**. Your exact example now works:

\`\`\`bash
curl -sL https://github.com/stripe/stripe-cli/releases/download/v1.38.1/stripe_1.38.1_linux_x86_64.tar.gz -o /tmp/stripe.tar.gz
\`\`\`

[...]

PR #122, opened 2026-03-14
feat: add OpenClaw plugin support

ipedro:
Hey! 👋

This PR adds OpenClaw plugin support to context-mode, enabling it to run as a native OpenClaw extension for Telegram, WhatsApp, Discord and other chat channels.

## What's included

- **OpenClaw plugin entry point** (`src/openclaw-plugin.ts`) — synchronous `register()` with async init pattern, hooks for the full agent lifecycle
- **Lifecycle hooks**: `session_start`, `before_tool_call`, `after_tool_call`, `before_compaction`, `after_compaction`, `before_prompt_build`, `before_model_resolve`
- **Multi-agent session isolation** — each `register()` call creates isolated closures (db, sessionId, hooks) so concurrent agents in the same gateway process don't cross-contaminate
- **WorkspaceRouter** (`src/openclaw/workspace-router.ts`) — defense-in-depth layer that maps workspace paths to session IDs for correct event routing
- **Compaction integration** — returns `firstKeptEntryId` from `compact()` so OpenClaw can truncate conversation history
- **`runId`/`toolCallId` on `AfterToolCallEvent`** — for event correlation across tool calls
- **Install script** (`scripts/install-openclaw-plugin.sh`) — one-shot setup: build, rebuild native deps, create extension dir, register in runtime config, restart gateway

## How it was validated

- Deployed and dogfooded on a live OpenClaw gateway (Raspberry Pi) with concurrent agents via Telegram
- Confirmed distinct session IDs per agent with correct workspace routing
- Verified all lifecycle hooks fire correctly in gateway mode
- Unit tests for workspace extraction and routing (10 tests, all passing)

## Test plan

- [x] `npx vitest run tests/openclaw/workspace-router.test.ts` — all 10 pass
- [x] `npm run build` — clean, no type errors
- [x] Live validation with concurrent agents on OpenClaw gateway

PR #339, opened 2026-04-24
fix(openclaw): register MCP sidecar on install so ctx_* tools surface to agent

murataslan1:
## Summary

Before this PR, installing context-mode on OpenClaw `2026.4.22` produced a silent registration failure: `openclaw plugins list` showed the plugin as `loaded`, `openclaw doctor` reported zero errors, `scripts/ctx-debug.sh` mostly passed — but none of the `ctx_*` tools surfaced to the agent. The model even refused an explicit request:

> "I can't use `ctx_execute` here because that tool isn't available in this session; if you want, I can do the same calculation with `exec` instead."

**Cause**: context-mode's `ctx_*` tools live in `src/server.ts`, exposed over stdio MCP (11× `server.registerTool` + `StdioServerTransport`). Other adapters (Claude Code, Codex, Cursor) spawn `server.bundle.mjs` via their platform's `mcpServers` config. The OpenClaw install script writes `plugins.allow` / `plugins.entries` to `openclaw.json` but never adds the `mcp.servers.context-mode` entry, so the gateway never spawns the sidecar and the agent never sees the tools.

This PR fills in that step.

## Changes

- **`scripts/lib/register-openclaw-config.mjs` (new)** — idempotent helper extracted from the install script. Writes `plugins.allow`, `plugins.entries`, cleans legacy `plugins.load.paths` (existing behavior), **and additionally registers `mcp.servers.context-mode`** pointing at `<pluginRoot>/server.bundle.mjs`. Re-running is a no-op; stale server paths are refreshed.
- **`scripts/install-openclaw-plugin.sh`** — step 5 delegates to the new helper (previously inline `node -e`).
- **`tests/plugins/openclaw.test.ts`** — 5 new unit tests against the helper: MCP entry presence, idempotency, stale-path refresh, `plugins.allow`/`entries` contract, legacy `plugins.load.paths` cleanup.
- **`docs/adapters/openclaw.md`** — new troubleshooting entry documenting the MCP-sidecar requirement and the manual `openclaw mcp set` recovery command.

## Verification

```
$ npm run typecheck
> tsc --noEmit
(clean)

$ npx vitest run tests/plugins/openclaw.test.ts tests/adapters/openclaw.test.ts
 Test Files  2 passed (2)
      Tests  117 passed (117)     # baseline 112 + 5 new
```

Live check on OpenClaw `2026.4.22` + context-mode `1.0.89` after the fix — the agent's tool inventory now includes the full `ctx_*` surface (OpenClaw prefixes MCP-sourced tool names with the server name):

```
context-mode__ctx_batch_execute
context-mode__ctx_doctor
context-mode__ctx_execute
context-mode__ctx_execute_file
context-mode__ctx_fetch_and_index
context-mode__ctx_index
context-mode__ctx_insight
context-mode__ctx_purge
context-mode__ctx_search
context-mode__ctx_stats
context-mode__ctx_upgrade
```

## Full debugging writeup

[gist — "OpenClaw MCP sidecar gap: debugging walkthrough"](https://gist.github.com/murataslan1/cd7b27577fcb535d56fe318c2339b400)

Covers initial (incorrect) diagnosis, cross-check with @benzntech on LinkedIn + [his follow-up on #45](https://github.com/mksglu/context-mode/issues/45#issuecomment-4307024324), the `src/server.ts` trace that surfaced the 11 `server.registerTool` calls, and the final resolution path.

## Relation to existing work

- Builds on @ipedro's adapter design in PR #122.
- Companion to issue #45 ([original comment](https://github.com/mksglu/context-mode/issues/45#issuecomment-4307024324), [follow-up](https://github.com/mksglu/context-mode/issues/45#issuecomment-4313313422)).
- The recent PreToolUse relaxation commits (`415ce57` revert of #230, `2731ca2` #166, `ece3abb` #229/#241) are **not** addressed here — they changed hook deny/passthrough behavior but did not cause this particular issue.

## Full test-suite caveats (environment-dependent)

In my local environment, `npm test` surfaces 5 failures in `tests/hooks/integration.test.ts → Security Policy Enforcement`, and `git stash`-ing this PR and re-running reproduces the same 5 failures on `main` — so they are not caused by this change on my machine. However, an independent review of the PR via the OpenAI Codex CLI (summary linked in a follow-up comment) **could not reproduce those** in a clean environment; that environment only hit a `Rust toolchain not configured` failure, also present on both branch and `main`. So whatever is going on in `tests/hooks/integration.test.ts` locally appears to be environment-specific, not a property of this PR or of `main`. CI on the project's own infra is the authoritative signal.

## Scope discipline

No changes outside the OpenClaw adapter, the OpenClaw install script, and OpenClaw docs. Other adapter hooks, `src/server.ts`, the session layer, and all non-OpenClaw tests are untouched.

## Open questions for @ipedro / @mksglu (non-blocking)

1. Should the plugin itself spawn the sidecar from `register()` rather than relying on `openclaw.json` config? If yes, what's the preferred `api.*` call?
2. `configs/openclaw/AGENTS.md` guidance uses unprefixed `ctx_execute` rather than the `context-mode__ctx_execute` name that the agent actually sees. Prefixless names in routing instructions may mildly confuse the model — worth a pass?
3. Is there a way to detect / warn when the OpenClaw version predates the `openclaw mcp set` surface?

### DIFF
--- a/configs/openclaw/AGENTS.md
+++ b/configs/openclaw/AGENTS.md
@@ -4,48 +4,48 @@ You have context-mode MCP tools available. These rules are NOT optional — they
 
 ## Think in Code — MANDATORY
 
-When you need to analyze, count, filter, compare, search, parse, transform, or process data: **write code** that does the work via `ctx_execute(language, code)` and `console.log()` only the answer. Do NOT read raw data into context to process mentally. Your role is to PROGRAM the analysis, not to COMPUTE it. Write robust, pure JavaScript — no npm dependencies, only Node.js built-ins (`fs`, `path`, `child_process`). Always use `try/catch`, handle `null`/`undefined`, and ensure compatibility with both Node.js and Bun. One script replaces ten tool calls and saves 100x context.
+When you need to analyze, count, filter, compare, search, parse, transform, or process data: **write code** that does the work via `context-mode__ctx_execute(language, code)` and `console.log()` only the answer. Do NOT read raw data into context to process mentally. Your role is to PROGRAM the analysis, not to COMPUTE it. Write robust, pure JavaScript — no npm dependencies, only Node.js built-ins (`fs`, `path`, `child_process`). Always use `try/catch`, handle `null`/`undefined`, and ensure compatibility with both Node.js and Bun. One script replaces ten tool calls and saves 100x context.
 
 ## BLOCKED commands — do NOT attempt these
 
 ### curl / wget — BLOCKED
 Any shell command containing `curl` or `wget` will be intercepted and blocked by the context-mode plugin. Do NOT retry.
 Instead use:
-- `ctx_fetch_and_index(url, source)` to fetch and index web pages
-- `ctx_execute(language: "javascript", code: "const r = await fetch(...)")` to run HTTP calls in sandbox
+- `context-mode__ctx_fetch_and_index(url, source)` to fetch and index web pages
+- `context-mode__ctx_execute(language: "javascript", code: "const r = await fetch(...)")` to run HTTP calls in sandbox
 
 ### Inline HTTP — BLOCKED
 Any shell command containing `fetch('http`, `requests.get(`, `requests.post(`, `http.get(`, or `http.request(` will be intercepted and blocked. Do NOT retry with shell.
 Instead use:
-- `ctx_execute(language, code)` to run HTTP calls in sandbox — only stdout enters context
+- `context-mode__ctx_execute(language, code)` to run HTTP calls in sandbox — only stdout enters context
 
 ### Direct web fetching — BLOCKED
 Do NOT use any direct URL fetching tool. Use the sandbox equivalent.
 Instead use:
-- `ctx_fetch_and_index(url, source)` then `ctx_search(queries)` to query the indexed content
+- `context-mode__ctx_fetch_and_index(url, source)` then `context-mode__ctx_search(queries)` to query the indexed content
 
 ## REDIRECTED tools — use sandbox equivalents
 
 ### Shell (>20 lines output)
 Shell is ONLY for: `git`, `mkdir`, `rm`, `mv`, `cd`, `ls`, `npm install`, `pip install`, and other short-output commands.
 For everything else, use:
-- `ctx_batch_execute(commands, queries)` — run multiple commands + search in ONE call
-- `ctx_execute(language: "shell", code: "...")` — run in sandbox, only stdout enters context
+- `context-mode__ctx_batch_execute(commands, queries)` — run multiple commands + search in ONE call
+- `context-mode__ctx_execute(language: "shell", code: "...")` — run in sandbox, only stdout enters context
 
 ### File reading (for analysis)
 If you are reading a file to **edit** it → reading is correct (edit needs content in context).
-If you are reading to **analyze, explore, or summarize** → use `ctx_execute_file(path, language, code)` instead. Only your printed summary enters context.
+If you are reading to **analyze, explore, or summarize** → use `context-mode__ctx_execute_file(path, language, code)` instead. Only your printed summary enters context.
 
 ### grep / search (large results)
-Search results can flood context. Use `ctx_execute(language: "shell", code: "grep ...")` to run searches in sandbox. Only your printed summary enters context.
+Search results can flood context. Use `context-mode__ctx_execute(language: "shell", code: "grep ...")` to run searches in sandbox. Only your printed summary enters context.
 
 ## Tool selection hierarchy
 
-1. **GATHER**: `ctx_batch_execute(commands, queries)` — Primary tool. Runs all commands, auto-indexes output, returns search results. ONE call replaces 30+ individual calls. Each command: `{label: "descriptive header", command: "..."}`. Label becomes FTS5 chunk title — descriptive labels improve search.
-2. **FOLLOW-UP**: `ctx_search(queries: ["q1", "q2", ...])` — Query indexed content. Pass ALL questions as array in ONE call.
-3. **PROCESSING**: `ctx_execute(language, code)` | `ctx_execute_file(path, language, code)` — Sandbox execution. Only stdout enters context.
-4. **WEB**: `ctx_fetch_and_index(url, source)` then `ctx_search(queries)` — Fetch, chunk, index, query. Raw HTML never enters context.
-5. **INDEX**: `ctx_index(content, source)` — Store content in FTS5 knowledge base for later search.
+1. **GATHER**: `context-mode__ctx_batch_execute(commands, queries)` — Primary tool. Runs all commands, auto-indexes output, returns search results. ONE call replaces 30+ individual calls. Each command: `{label: "descriptive header", command: "..."}`. Label becomes FTS5 chunk title — descriptive labels improve search.
+2. **FOLLOW-UP**: `context-mode__ctx_search(queries: ["q1", "q2", ...])` — Query indexed content. Pass ALL questions as array in ONE call.
+3. **PROCESSING**: `context-mode__ctx_execute(language, code)` | `context-mode__ctx_execute_file(path, language, code)` — Sandbox execution. Only stdout enters context.
+4. **WEB**: `context-mode__ctx_fetch_and_index(url, source)` then `context-mode__ctx_search(queries)` — Fetch, chunk, index, query. Raw HTML never enters context.
+5. **INDEX**: `context-mode__ctx_index(content, source)` — Store content in FTS5 knowledge base for later search.
 
 ## Output constraints
 
--- a/docs/adapters/openclaw.md
+++ b/docs/adapters/openclaw.md
@@ -50,6 +50,17 @@ The state directory doesn't exist at the expected path. If you installed OpenCla
 **Plugin installed but not loading**
 Clear the jiti cache (`rm -f /tmp/jiti/context-mode-*.cjs`) and restart the gateway. If the issue persists, verify the plugin appears in `openclaw plugins list`.
 
+**Plugin loads but `ctx_*` tools are missing from the agent's tool list**
+The plugin registers its hooks via `api.on(...)` / `api.registerCommand(...)`, but the agent-callable `ctx_*` tools live in the MCP server (`server.bundle.mjs`). OpenClaw surfaces them by spawning the server as an MCP sidecar declared in `mcp.servers.context-mode`. The install script writes this entry automatically (step 5); if you configured OpenClaw manually, verify with `openclaw mcp list` and add it if missing:
+
+```bash
+openclaw mcp set context-mode \
+  "{\"command\":\"node\",\"args\":[\"/absolute/path/to/context-mode/server.bundle.mjs\"]}"
+openclaw gateway restart
+```
+
+After the restart, the agent's tool inventory should include `context-mode__ctx_execute`, `context-mode__ctx_search`, `context-mode__ctx_fetch_and_index`, and the rest of the `ctx_*` surface (OpenClaw prefixes MCP-sourced tools with the server name).
+
 ## Hook Registration
 
 The adapter uses two different registration APIs, matching OpenClaw's internal architecture:
--- a/scripts/install-openclaw-plugin.sh
+++ b/scripts/install-openclaw-plugin.sh
@@ -69,42 +69,11 @@ else
   echo "  ⚠ could not verify discovery (openclaw not in PATH or plugin not found) — continuing anyway"
 fi
 
-# 5. Register in runtime config (idempotent)
+# 5. Register in runtime config (idempotent) — delegates to
+# scripts/lib/register-openclaw-config.mjs so the logic can be unit-tested
+# and so we can extend it (now also registers mcp.servers.context-mode).
 echo "→ registering in $OPENCLAW_JSON..."
-node -e "
-const fs = require('fs');
-const [runtimePath, pluginRoot] = process.argv.slice(1);
-let cfg;
-try {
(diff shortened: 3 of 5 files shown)


murataslan1:
### End-to-end verification — the patched install script surfaces `ctx_*` to the agent ✅

Re-ran the modified install script from this branch against a real OpenClaw state dir (`~/.openclaw`) after the commit was pushed:

```bash
$ git branch --show-current
fix/openclaw-register-mcp-sidecar

$ OPENCLAW_STATE_DIR="$HOME/.openclaw" bash scripts/install-openclaw-plugin.sh
...
→ registering in /Users/murat/.openclaw/openclaw.json...
  registered mcp.servers.context-mode → /Users/murat/Desktop/context-mode/server.bundle.mjs
  plugins.allow: ["context-mode"]
...
✓ done — context-mode plugin installed and active
```

The new `registered mcp.servers.context-mode → …` log line is emitted by the helper introduced in this PR. The previous config had the entry pointing at a different path (`/opt/homebrew/…`); the helper idempotently refreshed it to the clone path — the `updates a stale MCP server path when pluginRoot changes` unit test reproduced against a real install.

`openclaw mcp show context-mode` after install:

```json
{
  "command": "node",
  "args": ["/Users/murat/Desktop/context-mode/server.bundle.mjs"]
}
```

After `openclaw gateway restart`, the agent's tool inventory:

```
$ openclaw agent --agent main \
    --message "Reply with exact tool names starting with 'context-mode__', one per line, no explanation." \
    --thinking low

context-mode__ctx_batch_execute
context-mode__ctx_doctor
context-mode__ctx_execute
context-mode__ctx_execute_file
context-mode__ctx_fetch_and_index
context-mode__ctx_index
context-mode__ctx_insight
context-mode__ctx_purge
context-mode__ctx_search
context-mode__ctx_stats
context-mode__ctx_upgrade
```

All 11 `ctx_*` tools surface to the agent. The end-to-end chain works on a fresh run of the modified install script.

Full walkthrough (with pre/post state diffs) has been appended to the companion gist as **§11**: [openclaw-mcp-gap-debugging.md § 11](https://gist.github.com/murataslan1/cd7b27577fcb535d56fe318c2339b400#file-openclaw-mcp-gap-debugging-md).

### Pre-existing install-script gotcha surfaced during this run (NOT introduced by this PR)

The install script's step 1 runs `npm rebuild better-sqlite3` using whatever Node is picked up from `npm config get prefix`. On my machine that's Node 22; the LaunchAgent gateway runs under Node 24. Result: the binary was built against `NODE_MODULE_VERSION 127` while the gateway requires `137`, and plugin load failed until a manual `nvm use 24 && npm rebuild better-sqlite3`. This is orthogonal to the MCP-sidecar registration fix — possibly worth a separate follow-up issue to make the rebuild pick up the gateway's Node.

murataslan1:
### UI QA — three prompts in the Control UI dashboard, plus a follow-up commit

Before asking for review, I ran three prompts in the OpenClaw dashboard chat to stress what happens in practice (`openai-codex/gpt-5.4`, single fresh session, post-install-script state from the earlier verification comment).

---

**Prompt A — explicit tool name, non-cacheable output**

```
Use context-mode__ctx_execute to run this Python and return ONLY the printed output:
import hashlib, time
print(hashlib.sha1(str(time.time_ns()).encode()).hexdigest()[:12])
```

- Tool trace: `context-mode__ctx_execute` (single call)
- Tokens: `↑13.3k ↓62  R15.9k  7% ctx`
- Result: `4c925e2afbca` (a fresh SHA-1 of `time.time_ns()` — can't be cached, so the subprocess really ran)
- **Verdict**: ✅ explicit invocation path is live end-to-end.

---

**Prompt B — natural language, small target (2KB `package.json`)**

```
Fetch https://raw.githubusercontent.com/mksglu/context-mode/main/package.json,
index it, and tell me the version field value.
```

- Tool trace (in order):
  ```
  Tool call   context-mode__ctx_fetch_and_index
  Tool output context-mode__ctx_fetch_and_index
  Tool call   web_fetch
  Tool output web_fetch
  ```
- Tokens: `↑366 ↓82  R28.9k  0% ctx`
- Result: `1.0.89` (correct)
- "What tool did you use?" follow-up: `web_fetch` (the last one)
- **Verdict**: ⚠️ partial auto-routing. AGENTS.md guidance made the model reach for `context-mode__ctx_fetch_and_index` first — unprompted — but it then redundantly fell back to `web_fetch`.

---

**Prompt C — natural language, large target (58KB `README.md`)**

```
Fetch https://raw.githubusercontent.com/mksglu/context-mode/main/README.md,
index it, then find sections that mention "context window" and reply with
the count of matching sections only. No explanation.
```

- Tool trace:
  ```
  Tool call   context-mode__ctx_fetch_and_index
  Tool output context-mode__ctx_fetch_and_index
  Tool call   web_fetch
  Tool output web_fetch
  Tool call   context-mode__ctx_execute
  Tool output context-mode__ctx_execute
  ```
- Tokens: `↑6.4k ↓475  R94.7k  3% ctx`
- Result: `4` (correct)
- **Verdict**: ⚠️ same pattern as B. `context-mode__ctx_fetch_and_index` fired, "think in code" (via `ctx_execute`) worked, but `ctx_search` — the intended follow-up to `fetch_and_index` — was never invoked; the model dragged the raw README in via `web_fetch` and counted with `ctx_execute` instead.

---

### Root cause — AGENTS.md prefix mismatch

```bash
$ grep -c 'context-mode__' configs/openclaw/AGENTS.md
0
$ grep -cE 'ctx_[a-z_]+' configs/openclaw/AGENTS.md
14
```

AGENTS.md referred to the tools by their unprefixed names (`ctx_execute`, `ctx_search`, …) 14 times. The agent sees them as `context-mode__ctx_*` because OpenClaw's MCP aggregator prefixes server names onto the tools it surfaces. The model mostly bridged the gap on its own (the explicit invocation in Prompt A worked; `context-mode__ctx_fetch_and_index` fired in B and C). But under pressure — short prompt, small target, a built-in with a shorter name available — it defaulted to `web_fetch` / `exec`.

### Follow-up commit on this PR — `5fc3ac6`

Rather than punt this to a separate issue, I added a second commit to the branch that rewrites the 14 mentions in `configs/openclaw/AGENTS.md` to use the names the agent actually sees:

```bash
perl -i -pe 's/\bctx_/context-mode__ctx_/g' configs/openclaw/AGENTS.md
# 14 replacements, 0 bare ctx_ remaining
```

Scope is still OpenClaw-only — `configs/pi/AGENTS.md` and every other adapter's AGENTS.md are untouched, because those platforms don't apply the server-name prefix.

Commits now on the PR:

| Commit | Purpose |
|---|---|
| `94b4fca` | register MCP sidecar on install so `ctx_*` tools surface to agent |
| `5fc3ac6` | prefix `ctx_*` names in AGENTS.md so the model invokes MCP variants |

The second commit is a system-prompt text-only change — tests and typecheck are unchanged (still `tests/plugins/openclaw.test.ts + tests/adapters/openclaw.test.ts` → 117/117, typecheck clean).

### Next-step verification I haven't done yet

Re-running Prompts B and C against the second commit should show `web_fetch` dropping out and `context-mode__ctx_search` being called where AGENTS.md now explicitly prescribes the `fetch_and_index → search` pattern. Happy to post results if that would help review.

### Consolidated evidence

Full walkthrough (with all commands, grep results, screenshots described textually): updated gist — [§12 UI QA report](https://gist.github.com/murataslan1/cd7b27577fcb535d56fe318c2339b400#file-openclaw-mcp-gap-debugging-md).

cc @benzntech (the LinkedIn cross-check on the MCP-sidecar reasoning led directly to these findings).

murataslan1:
### Independent Codex CLI review — results + follow-up commit `3766cca`

Before flagging this for human review I ran the PR through `codex-cli 0.122.0` as an independent reviewer. Full verdict: **ship, no blocking findings.** The review verified the two-commit scope, confirmed the runtime tool surface (11 `context-mode__ctx_*` names visible to the agent), and re-ran both the targeted and full test suites.

Two items came back as legitimate weaknesses, and I've addressed both in commit `3766cca`:

**1. Full-reset of `mcp.servers.context-mode` could drop user/OpenClaw-added fields.**
The previous helper wrote `{ command: "node", args: [serverBundle] }` wholesale when the entry needed refreshing. If someone had manually set `env`, `cwd`, `timeout`, or if a future OpenClaw release adds its own optional keys, the install script would silently erase them on the next run. The helper now spreads the existing entry and only overwrites the two fields it owns:

```js
const base = existing && typeof existing === "object" ? existing : {};
servers["context-mode"] = { ...base, command: "node", args: [serverBundle] };
```

**2. The `try/catch` around `JSON.parse` was never exercised by a test.** Added in the same commit.

New tests in `tests/plugins/openclaw.test.ts` (now 119/119, +2 over the previous 117):

- *preserves unrelated fields on the existing mcp.servers entry when refreshing the path* — seeds `env`/`cwd`/`timeout` on the old entry, bumps `pluginRoot`, asserts all three survive alongside the new `args[0]`.
- *throws a useful error when the runtime config is not valid JSON* — exercises the parse-error path.

### PR body corrected

Codex couldn't reproduce the "5 failures in `tests/hooks/integration.test.ts`" I claimed in my original PR body. In their environment the only failure on both branch and `main` was a missing Rust toolchain. So those 5 failures are apparently specific to my local setup, not a property of `main`. The PR body has been edited to reflect that (the section is now "Full test-suite caveats (environment-dependent)" rather than a confident "pre-existing on main" claim). CI on the project's infra is the authoritative signal.

### Remaining architectural question (unchanged, not part of this PR)

The review left open the same question I flagged at the bottom of the PR body — *should the plugin itself spawn the MCP sidecar from `register()` rather than relying on `openclaw.json` config?* — and agreed it's worth a separate maintainer discussion rather than scope-creeping this PR.

### Current state of PR #339

| Commit | Purpose |
|---|---|
| `94b4fca` | register MCP sidecar on install so `ctx_*` tools surface to agent |
| `5fc3ac6` | prefix `ctx_*` names in AGENTS.md so the model invokes MCP variants |
| `3766cca` | preserve unrelated MCP server fields; add edge-case tests |

- Typecheck: clean
- `tests/plugins/openclaw.test.ts + tests/adapters/openclaw.test.ts`: **119/119**
- Scope: still OpenClaw-only (adapter docs / install / tests / configs/openclaw)

Full Codex transcript + updated walkthrough lives in the gist: [§13 — Independent Codex review follow-up](https://gist.github.com/murataslan1/cd7b27577fcb535d56fe318c2339b400).

mksglu:
@murataslan1 seems has a conflict :)

murataslan1:
### Post-`5fc3ac6` validation — `web_fetch` fallback is gone ✅

Restarted the gateway after commit `5fc3ac6` (so the plugin re-reads the newly-prefixed `configs/openclaw/AGENTS.md`), opened a fresh dashboard session, and re-ran the same Prompt C from the UI QA comment.

**Prompt C (verbatim, unchanged):**

```
Fetch https://raw.githubusercontent.com/mksglu/context-mode/main/README.md,
index it, then find sections that mention "context window" and reply with
the count of matching sections only. No explanation.
```

**Tool-call chain — before vs after the prefix fix:**

| Turn | Before `5fc3ac6` (old `AGENTS.md`) | After `5fc3ac6` (prefixed `AGENTS.md`) |
|---|---|---|
| 1 | `context-mode__ctx_fetch_and_index` | `context-mode__ctx_fetch_and_index` |
| 2 | **`web_fetch`** ⚠ (redundant) | — |
| 3 | `context-mode__ctx_execute` | `context-mode__ctx_execute` |
| | 3 tool calls | **2 tool calls** |
| Model reply | `4` | `4` |

**Headline**: `web_fetch` is gone. The redundant fallback that showed up in both Prompt B (small target) and Prompt C (large target) during the UI QA run no longer fires once `AGENTS.md` refers to the tools by the names the agent actually sees (`context-mode__ctx_*`). That was the whole point of `5fc3ac6`.

**Running-context footprint dropped**: `R94.7k → R37.1k` on the same prompt. Apples-to-oranges across sessions so treat this as directional, not a precise delta — but the direction is right: no `web_fetch` means less raw web content threading through the conversation context.

**Observation on `ctx_search`**: the model still preferred `context-mode__ctx_execute` for the count instead of `context-mode__ctx_search`. That's consistent with AGENTS.md's "Think in Code" mandate — for a precise count, deterministic code is a reasonable choice over BM25-scored top-K search, and both are legitimate. Not a regression. If you ever want to push the model toward `ctx_search` for questions like this, a worked example of the `fetch_and_index → search` pattern in AGENTS.md would be the lever; that's a follow-up polish rather than something in this PR's scope.

### Where this leaves PR #339

All three commits validated end-to-end on OpenClaw `2026.4.22` + context-mode `1.0.89`:

| Commit | What it fixes | Verified by |
|---|---|---|
| `94b4fca` | Tools surface to the agent at all | `openclaw plugins list` + tool inventory (11 `context-mode__ctx_*` names) |
| `5fc3ac6` | Routing no longer falls back to `web_fetch` | This comment (3 tool calls → 2, chain above) |
| `3766cca` | Install script no longer overwrites user-added MCP server fields | `tests/plugins/openclaw.test.ts` (119/119) |

Full walkthrough updated in the companion gist: [§14 — Post-prefix validation](https://gist.github.com/murataslan1/cd7b27577fcb535d56fe318c2339b400).

murataslan1:
@mksglu — my bad, the earlier rebase went onto `main` but this PR targets `next`. Just rebased cleanly onto `origin/next` with `--onto` (my three commits only, no `stats.json` pickup). Mergeable now.

Hashes: `aacdaf5`, `f90c98d`, `3eec67d`. 119/119 tests, typecheck clean.

@ipedro — thanks for the first read. When you're good, we're ready on our side for whichever release cycle fits. No rush.

PR #349, opened 2026-04-26
fix(search): reward phrase frequency in FTS5 reranker

sebastianbreguel:
## What

Add a saturating phrase-frequency reward to `#applyProximityReranking` in `src/store.ts`. New pure helper `countAdjacentPairs(positionLists, terms, gap=30)` counts ordered adjacent-pair occurrences of consecutive query terms; reranker contributes `0.5 * min(1, pairs/4)` on top of the existing `proximityBoost`.

## Why

The existing `proximityBoost = 1 / (1 + minSpan / contentLen)` formula has a length-normalization artifact: a long doc with one tight occurrence outranks a short doc with multiple tight occurrences at the same span, because `contentLen` grows in the denominator. `minSpan` returns a single number — the tightest window — so frequency does not move it.

Length-normalization is intentional (filters short, low-signal noise). This change preserves that intent: the new reward is purely additive, capped at 0.5 (below proximity max ≈1.0, in title-boost range), and saturates at 4 hits so keyword-stuffed docs cannot unbeatably dominate.

## How

- New module-level pure helper `countAdjacentPairs` (two-pointer over already-sorted `findAllPositions` output, O(L+R) per consecutive term pair). Each right position is consumed by at most one left, so `"foo foo bar"` counts 1 pair, not 2 — matches IR phrase-occurrence intent.
- Reranker gates on the same `terms.length >= 2 && all positions non-empty` condition as `proximityBoost`. Reuses the already-computed positions arrays — no extra content scan, no extra allocations.
- 30-char gap covers natural prose without rewarding distant matches (already handled by `minSpan`).

## Bench

**Latency** (1000-doc corpus, 500 iters, `searchWithFallback`, p95 us/call):

| query                 | baseline | with-boost |
|-----------------------|----------|------------|
| cache invalidation    | 403      | 325        |
| database connection   | 398      | 390        |
| auth token            | 285      | 264        |
| rate limiter          | 320      | 299        |
| circuit breaker retry | 517      | 453        |

Latency neutral — well within noise. The new pass is strictly cheaper than the existing `findMinSpan` (no array copy or sort) and reuses already-allocated positions.

**Ranking quality** (3 controlled cases of long-1-occurrence vs short-N-occurrences):

| metric            | baseline | with-boost |
|-------------------|----------|------------|
| top-1 correct     | 0/3      | 3/3        |

Each case is a query where the short, frequent doc is clearly more relevant but baseline ranks the long, padded doc first due to the length-norm artifact.

## Test plan

- [x] `npx tsc -b --noEmit` — clean
- [x] `npx vitest run` — 1606+ pass, 0 search failures (5 hooks-test flakes pre-existing on `next`, unrelated)
- [x] `npm run build && npm run bundle` — clean
- [x] 4 new tests under `Phrase-frequency reward in reranking`:
  - multiple phrase occurrences outrank a single tight phrase at similar minSpan
  - phrase-frequency reward respects query order (regression guard)
  - 3-term query: only consecutive-pair adjacency contributes
  - saturation: 8-hit stuffed doc cannot beat 4-hit doc by more than the cap allows (3-doc fixture: cap binding keeps a natural-prose doc in top-3)
- [x] Cross-platform: pure-JS path, no OS-specific calls; runs identically on ubuntu/macos/windows

## Affected platforms

- All — pure scoring change in shared FTS5 path

## Notes

- No public API change. `searchWithFallback` signature unchanged.
- No schema change, no new prepared statements, no cache.
- `findMinSpan` and the length-norm formula are deliberately untouched.


mksglu:
Thanks @sebastianbreguel! Clean implementation — well-scoped scoring change with solid test coverage.

Architecture review notes:
- `countAdjacentPairs` two-pointer approach is correct, O(L+R) per term pair
- 0.5 cap keeps phrase boost subordinate to proximity (max ~1.0) — good calibration
- Saturation at 4 hits prevents keyword stuffing — tested
- No public API change, no schema change — safe merge

Merged into `next`. 🎉

Issue #347, opened 2026-04-26
Regression in #230: PPID mismatch — mcpRedirect swallows all redirects when hook spawned via shell wrapper


se7enango:
## Summary

The MCP-readiness sentinel guard introduced in [`04569b1`](https://github.com/mksglu/context-mode/commit/04569b12a1df4a58a60f0e9639d0b2dba0ce5c70) (fix for #230) is **silently bypassed in normal Claude Code sessions** because the MCP server and the hook process resolve `process.ppid` to **different PIDs**. As a result, `isMCPReady()` always returns `false` for hooks, and `mcpRedirect()` swallows every deny/modify decision — `WebFetch`, `curl`/`wget`, inline-HTTP, and gradle/maven redirects all become silent passthroughs.

Effectively, the core routing policy of context-mode is no-op for any Claude Code installation that spawns hooks via a shell wrapper.

## Root cause

`src/server.ts` (MCP server) writes the sentinel using **its own** `process.ppid`:

```js
// src/server.ts (introduced by #230)
const mcpSentinel = join(tmpdir(), `context-mode-mcp-ready-${process.ppid}`);
// ...
writeFileSync(mcpSentinel, String(process.pid));
```

`hooks/core/mcp-ready.mjs` reads the sentinel using **its own** `process.ppid`:

```js
export function sentinelPath() {
  return resolve(tmpdir(), `context-mode-mcp-ready-${process.ppid}`);
}
```

This only works if both processes share the **same** parent. They don't:

| Process | How spawned | `process.ppid` |
|---|---|---|
| MCP server | Claude Code → `node start.mjs` (direct) | Claude Code main PID |
| Hook (`pretooluse.mjs`) | Claude Code → `bash -c \"node hooks/pretooluse.mjs\"` | Transient bash PID |

So the sentinel exists at `/tmp/context-mode-mcp-ready-${CC_PID}`, but the hook looks for `/tmp/context-mode-mcp-ready-${TRANSIENT_BASH_PID}` — different file, never found.

## Reproduction

Environment:
- context-mode `v1.0.94` (installed via Claude Code marketplace, `autoUpdate: true`)
- Claude Code CLI on Linux (WSL2, Ubuntu)

Steps:
1. Start a Claude Code session with context-mode plugin enabled.
2. Verify the MCP server is alive and the sentinel exists:
   ```
   ls /tmp/context-mode-mcp-ready-*
[...]

PR #354, opened 2026-04-27
test(hooks): add cross-platform regression matrix for MCP readiness

sebastianbreguel:
## What / Why / How

**What.** Adds `tests/hooks/mcp-ready.test.ts` — a dedicated regression matrix for `hooks/core/mcp-ready.mjs`, the module that #347 rewrote.

**Why.** PR #347 fixed the PPID-bypass on WSL2 / `bash -c "node ..."` by switching from a `process.ppid`-keyed sentinel lookup to a directory-scan with `kill(pid, 0)` liveness probes. The 11 test files touched in #347 updated the sentinel **path** to use `process.pid`, but none of them assert the load-bearing contract: that `isMCPReady()` returns `true` when the only live sentinel was written by a process whose PID is **outside** the test runner's process tree. A future refactor that re-introduces PPID coupling, hardcodes a different temp root, or removes the stale-cleanup branch would not be caught by current CI.

**How.** One new file, three `describe` blocks, no production code changes:

1. **`mcp-ready: contract`** — locks in `sentinelPathForPid` shape, deprecated `sentinelPath` backward-compat (still keyed by `process.ppid` for one release cycle per the JSDoc), `sentinelDir()` platform branch (`/tmp` on Unix, `os.tmpdir()` on win32), happy path, and resilience to malformed payloads (empty, non-numeric).
2. **`mcp-ready: stale-cleanup self-healing`** — `isMCPReady()` unlinks dead-PID sentinels (uses `INT32_MAX` as a guaranteed-dead PID, no spawn). Gated by `describe.skipIf(POLLUTED)` to skip locally when an unrelated MCP server's live sentinel exists in `/tmp` (iteration order would make assertions flaky); CI runners are clean and run these.
3. **`mcp-ready: PPID-independence (regression for #347)`** — spawns a child Node process whose PID ∉ {`process.pid`, `process.ppid`}, has it write its own sentinel via the same path shape `src/server.ts` uses, then asserts `isMCPReady() === true`. This is the test that, if removed, allows #347 to silently regress.

## Affected platforms

- [x] ubuntu-latest
- [x] macos-latest
- [x] windows-latest

Cross-platform notes:
- Uses `path.join` throughout, no hardcoded `/` separators.
- The `sentinelDir()` test mocks `process.platform` via `Object.defineProperty(..., { configurable: true })` and restores in `afterEach`.
- The unreadable-file edge case is omitted (would require chmod, which is a no-op on Windows).
- Child-spawn uses `process.execPath` + `-e` inline script for portability.

## Test plan

- [x] `npx vitest run tests/hooks/mcp-ready.test.ts` passes locally on macOS (8 passed, 2 skipped under POLLUTED gate; in clean envs the 2 skipped also pass).
- [x] `npx tsc -b --noEmit` passes.
- [x] No production source changes; bundles untouched.
- [ ] CI matrix on this PR confirms ubuntu/macos/windows × Node 20 all pass.

## Checklist

- [x] Tests added (TDD: red → green); no implementation changes
- [x] `npm test` passes
- [x] Docs updated if needed — N/A (test-only)
- [x] Targets `next` branch


mksglu:
Hi @sebastianbreguel! Thanks for the PR. Please use existing Unit Test file instead of new one.

sebastianbreguel:
Done — moved the three describe blocks into `tests/hooks/core-routing.test.ts` (after the existing `routePreToolUse` block) and deleted `tests/hooks/mcp-ready.test.ts`. The stale-cleanup block has a local `beforeEach` that unlinks the file-level `mcpSentinel` so the runner's own live sentinel doesn't mask the dead-PID cleanup the test verifies. New commit pushed.

sebastianbreguel:
Resolved merge conflict with `next` — merged imports and fixed sentinelDir tests to clear env override when testing platform defaults. Ready for review.

PR #356, opened 2026-04-27
feat(store): instrument fuzzyCorrect cache with hit/miss/eviction counters

sebastianbreguel:
## What

Adds observability to the fuzzy-correction LRU cache in `ContentStore`:

- 4 internal counters scoped to the current cache lifetime: `#fuzzyCacheHits`, `#fuzzyCacheMisses`, `#fuzzyCacheEvictions`, `#fuzzyCacheLookups`
- Public read-only snapshot accessor `getFuzzyCacheStats()` returning `{ size, capacity, hits, misses, evictions, lookups, hitRate }`
- Counters reset alongside the existing `#fuzzyCache.clear()` invalidation in `#insertChunks`

No behavior change to `fuzzyCorrect()` semantics, no DB schema change.

## Why

The fuzzy cache is a fixed-cap (256) LRU. Today there's no way to tell whether 256 is the right number for any given workload — short sessions might never approach the cap, long sessions might be evicting heavily. Before tuning anything, we need to measure.

This PR is **measurement only**. It does not act on the data. A follow-up can decide whether the cap should be adjusted (statically or adaptively) once we have real hit/eviction data from production-shaped workloads.

Builds naturally on the cache cleanup already shipped in #272.

## How

- `src/store.ts`: add 4 private counter fields next to `#fuzzyCache`, increment on the existing hit / miss / eviction branches in `fuzzyCorrect()`, reset on `#fuzzyCache.clear()` in `#insertChunks` (line ~1284), and expose `getFuzzyCacheStats()` as a fresh-object snapshot per call.
- `tests/core/search.test.ts`: 6 new tests covering fresh-store zeros, hit/miss accounting, eviction counter, `clear()`-on-insert reset, `lookups=0` hitRate guard (no NaN), and snapshot semantics.

## Affected platforms

- Linux / macOS / Windows: identical (single-file logic change, no platform-specific code)

## Test plan

- [x] `npx vitest run tests/core/search.test.ts` — 141/141 pass (18 fuzzy-suite tests including 6 new)
- [x] Bundle artifacts updated: `server.bundle.mjs`, `cli.bundle.mjs`
- [x] Verified \`getFuzzyCacheStats()\` returns identical counter snapshots after a scripted hit/miss workload

## Checklist

- [x] Tests added (TDD red→green)
- [x] No DB schema changes
- [x] Bundle artifacts in PR
- [x] Pure observability — no semantic change to fuzzy correction
- [x] Root-cause-clean: no flag, no env, no opt-in

## Notes

Sets up the substrate for a follow-up "should we resize / make the cap adaptive?" discussion, but **that decision should be data-driven** — let this land first, gather real numbers, then revisit. I have a draft branch with an adaptive policy ready, but I'd rather not guess thresholds without the measurement layer in.


mksglu:
Thanks @sebastianbreguel — clean implementation, well-tested (6 new tests covering all edges), surgical diff. The reasoning in the PR description is sound: "before tuning anything, we need to measure."

I ran a real benchmark on next vs this branch, and want to share what I found before deciding:

**Benchmark setup:** Real ContentStore + SQLite + indexed sample code. 30K fuzzyCorrect calls, mix of typos / exact matches / misses.

**Results:**
- Median next (no counters): 0.14μs/call
- Median this PR (with counters): 0.11μs/call
- Overhead: statistically zero (within noise)
- Memory: 32 bytes per ContentStore instance

**But the more interesting data point:**

| Metric | Value |
|--------|-------|
| Cache size used | 13 / 256 (5%) |
| Hit rate | 99.96% |
| Evictions | 0 |
| Lookups | 31,000 |

The 256 cap is **massively oversized** for the actual workload. Vocabulary fuzzy queries don't churn — same words get hit repeatedly, and the working set stays small. The data we'd collect with these counters would just confirm what's already visible in this single bench run: 256 is fine, possibly we could drop to 64.

**Why I'm not merging right now:**

The instrumentation is correct, but the value is conditional on a follow-up that uses it:
- No tool currently exposes `getFuzzyCacheStats()` (not in ctx_stats, ctx_doctor, or any MCP tool)
- No code path consumes the metrics for runtime decisions
- The tuning question ("is 256 right?") appears already answered: yes, with margin

Adding observability without a consumer or a tuning decision creates code that lives forever but never pays for itself. I'd rather wait until there's a concrete next step — e.g., "adaptive cache sizing PR" or "add cache health to ctx_doctor" — and ship the counters as part of that PR.

**If you want to drive the follow-up**, here's what would unlock the merge:
1. **Adaptive sizing**: shrink cap when hit rate stays >95% for N lookups, grow when evictions spike
2. **Surface in ctx_doctor**: add a "Cache Health" check that flags eviction storms
3. **Surface in ctx_stats**: add fuzzy cache hit rate to the session summary (sales angle: "99% hit rate = no wasted CPU on Levenshtein")

Any of those make this PR's counters earn their keep.

Closing for now — not because the work is bad (it isn't), but because the observability needs a consumer to be worth the maintenance surface. Thanks for the careful diff and the tests; happy to review a follow-up that uses these signals.

PR #358, opened 2026-04-27
feat(batch_execute): opt-in concurrency for parallel command runs

sebastianbreguel:
## Summary

Adds opt-in `concurrency` field (default 1, max 8) to `ctx_batch_execute`. Independent shell commands now run in parallel when requested.

- **Serial path preserved bit-for-bit** at `concurrency=1` (default). Existing callers unaffected — same shared timeout budget, same cascading skip-on-timeout.
- **Parallel path** uses inline worker pool (~10 LOC, no new deps). Order preserved via index-keyed array. Each command gets its full `timeout` ms (not a shared budget).
- Deny policy still pre-runs once per batch. `__CM_FS__` markers parsed per command. `bytesSandboxed` accumulator stays race-free (JS single-thread).

## Bench (macOS arm64, Node 22, 5× 500ms sleeps)

| concurrency | wall-clock | speedup |
|-------------|------------|---------|
| 1           | 3336ms     | 1.00×   |
| 2           | 2120ms     | 1.57×   |
| 4           | 1598ms     | 2.09×   |
| 8           |  984ms     | 3.39×   |

Reproduce: `npm run bench:batch-parallel`. Tunable via `N`, `SLEEP_MS`, `LEVELS` env vars. Bench script imports `runBatchCommands` directly — no MCP stdio in the loop.

## Design notes

- `concurrency` (integer 1–8) chosen over `parallel: boolean` for forward flexibility — single knob, clear semantics.
- Default = 1 (opt-in). Conservative; no breaking change for callers depending on serial side-effects.
- Inline semaphore (worker pool with shared `nextIdx++`) instead of `p-limit` dep.
- Per-command timeout under concurrency: shared budget loses meaning when commands run in parallel. Documented in tool description.
- Cascading skip-on-timeout dropped under concurrency (commands already started). Each times out independently.

## Test plan

- [x] Source-scan tests: schema field, tool description, `bytesSandboxed` callback shape
- [x] `runBatchCommands` serial path: happy path, cascading skip, shared timeout
- [x] `runBatchCommands` parallel path: happy path, order preservation under reverse delays, concurrency cap, per-cmd timeout, cap clamped at cmd count, FS bytes callback
- [x] Edge cases: empty commands, empty stdout, `nodeOptsPrefix` prepending
- [x] All 118 tests in `tests/core/server.test.ts` pass
- [x] Local bench produces expected speedup curve

## Files

- `src/server.ts` — extracted `runBatchCommands` + types, added schema field, updated tool description
- `tests/core/server.test.ts` — 3 new describe blocks (~180 LOC) + updated existing source-scan test
- `tests/bench/batch-execute-parallel.ts` — manual bench script (not vitest)
- `package.json` — `bench:batch-parallel` script


sebastianbreguel:
### Real end-to-end test on Claude Code MCP transport

Spawned `npx tsx src/server.ts` as a child, spoke MCP JSON-RPC over stdio (initialize → notifications/initialized → tools/list → tools/call), captured the live `ctx_batch_execute` responses. Not the unit tests, not the bench — the actual MCP path Claude Code uses.

**Schema (from `tools/list`):**
```json
{"type":"integer","minimum":1,"maximum":8,"default":1,"description":"Max commands to run in parallel (1-8, default: 1). >1 switches to per-command timeouts (no shared budget) and individual `(timed out)` blocks instead of cascading skip."}
```

**Wall-clock (4 commands with reverse delays: 800ms / 600ms / 400ms / 200ms):**

| concurrency | wall-clock | speedup |
|-------------|-----------|---------|
| 1 (serial)  | 2546ms    | —       |
| 4 (parallel)| 1137ms    | 2.24×   |

Serial ≈ Σ delays = 2000ms (+ MCP/index overhead). Parallel ≈ max delay = 800ms (+ overhead).

**Order preservation under reverse delays.** `cmd_d` finishes first (200ms) but `cmd_a` is slowest (800ms). Index summary in the response:

```
## Indexed Sections

- cmd_a (0.0KB)
- cmd_b (0.0KB)
- cmd_c (0.0KB)
- cmd_d (0.0KB)
```

Input order, not completion order. Both `concurrency=1` and `concurrency=4` produce identical ordering — index-keyed worker pool is doing what it claims.

**Validation:** `tools/call` with `concurrency: 99` is rejected by Zod (`max=8`) before the handler runs. `concurrency: 1` is bit-for-bit the legacy serial path.

Harness script that produced this: https://gist.github.com/sebastianbreguel — happy to inline if you want it added under `tests/` instead of as a one-shot.

mksglu:
## Adopted with credit — closing this PR

Hi @sebastianbreguel, thank you for this contribution. The design here was the right one, and we've adopted it on `next` with you credited as Co-Author. Closing this PR in favor of the merged work.

### What landed (commit [b392c2f](https://github.com/mksglu/context-mode/commit/b392c2f))

The core mechanism from your PR — exporting `runBatchCommands` with serial vs parallel paths, opt-in `concurrency: 1-8` parameter, in-flight cap, per-command timeouts, throw isolation — shipped as designed.

We layered the following on top before merging:

1. **Shared concurrency primitive** — extracted `src/concurrency/runPool.ts` so the same in-flight-cap + Promise.allSettled + order-preserving worker pool also drives the new `ctx_fetch_and_index` batch path. `runBatchCommands` is now a thin adapter over `runPool`.
2. **`ctx_fetch_and_index` batch shape** — accepts `requests: [{url, source?}]` + `concurrency: 1-8` + serial-write FTS5 drain (workers fetch in parallel, indexer queue serializes to avoid SQLite WAL contention).
3. **Hardening** from a 2-round multi-agent architectural review:
   - Worker `try/catch` + `Promise.allSettled` — one job throw can no longer strand siblings or leave undefined output slots.
   - Timeout sentinel routes through `formatCommandOutput` so partial-stdout-on-timeout still strips `__CM_FS__` markers and counts bytes.
   - `effectiveConcurrency = min(N, os.cpus().length)` opt-in cap, with the response surfacing the cap reason in caveman style.
4. **Tool description** rewritten with positive `PARALLELIZE I/O` ✅/❌ guidance plus a NON-NEGOTIABLE THINK IN CODE clause so LLMs understand WHEN to set concurrency, not just THAT they can.
5. **14 adapter routing docs** updated with a consistent `Parallel I/O batches` paragraph (each adapted to the host's tool-prefix style and the GitHub rate-limit caveat).
6. **`mcp_tool_call` extractor** in `src/session/extract.ts` + `getMcpToolUsage()` in `src/session/analytics.ts` so the `concurrency` parameter is observable in `ctx_stats` (median/max per batch tool).
7. **Tests**: 7 `runPool` unit tests, 13 `ctx_fetch_and_index` batch tests, 3 P0 hardening tests including a CI-checked timing assertion (5×100ms at concurrency=5 finishes <200ms — replaces the deleted bench), 4 `mcp_tool_call` extractor tests including a UTF-8 multibyte regression.

### Why we didn't merge directly

A few reasons, none of them about your code quality:

- **Bundle artifacts.** PR included `cli.bundle.mjs` and `server.bundle.mjs`. Per our CI policy, those rebuild only on `main` push (handled by `.github/workflows/bundle.yml`). Committing them on a feature branch creates merge churn against the CI-generated baseline.
- **Local benchmark file.** The `tests/bench/batch-execute-parallel.ts` script was excellent for proving the speedup locally, but `CONTRIBUTING.md` (L275) restricts new test files. We replaced it with an inline vitest timing assertion in `tests/core/server.test.ts` that runs in every CI build.
- **Tool description.** Your description was technically accurate but didn't give LLMs strong-enough heuristics to USE the parameter without explicit user prompting. We rewrote it with positive guidance plus ✅/❌ examples drawn from real workloads (gh, curl, multi-region cloud) and explicit rules for CPU-bound (npm test, build, lint) and stateful (ports, locks) cases.
- **Cross-tool design** — once we had the primitive, it made sense to ship `ctx_fetch_and_index` batch + observability in the same release so users get the full multi-tool concurrency story at once instead of incrementally over three PRs.

The PRD that drove the final design (architectural standard, decision matrix, 5-agent review process) is in our worktree as `PRD-concurrency-architectural.md`.

Thanks again for the original spark — your name is in the commit message and in our changelog for v1.0.x. Hope to see more contributions.

PR #359, opened 2026-04-27
feat(qwen-code): improve adapter — configureAllHooks, getInstalledVersion, projectDir fallback, PostToolUse matchers

btxbtxbtx:
## Summary

This PR ports 4 improvements from a battle-tested custom Qwen Code adapter into the official one:

### 1. `configureAllHooks` — full implementation
Currently returns `[]` (no-op). This PR implements:
- **Phase 1**: Clean stale context-mode hooks (those referencing deleted version directories after upgrade)
- **Phase 2**: Register fresh PreToolUse and SessionStart hooks with correct matchers and paths
- Writes back to `~/.qwen/settings.json`

Without this, `ctx upgrade` on Qwen Code leaves hooks unconfigured.

### 2. `getInstalledVersion` — actual detection
Currently always returns `"not installed"`. This PR checks `~/.qwen/settings.json` for context-mode hook scripts, returning `"installed (hooks configured)"` when found. This fixes `ctx doctor` reporting false negatives.

### 3. `projectDir` fallback in `claude-code-base.ts`
All 4 parse methods (`parsePreToolUseInput`, `parsePostToolUseInput`, `parsePreCompactInput`, `parseSessionStartInput`) now use `process.env[this.projectDirEnvVar] ?? process.cwd()` instead of just the env var. This prevents `undefined` projectDir when the env var is not set — a defensive improvement that benefits both Claude Code and Qwen Code adapters.

### 4. PostToolUse precise matchers
Changed from empty matcher `""` (matches all tools) to a pipe-separated list of tools that actually produce events: `run_shell_command|read_file|write_file|edit|glob|grep_search|todo_write|agent|ask_user_question|mcp__`. This reduces unnecessary hook triggers and avoids "hook error" display for tools that produce zero events.

### Test results
All 22 existing qwen-code adapter tests pass. No regressions.

### DIFF
--- a/src/adapters/claude-code-base.ts
+++ b/src/adapters/claude-code-base.ts
@@ -61,7 +61,7 @@ export abstract class ClaudeCodeBaseAdapter extends BaseAdapter {
       toolName: input.tool_name ?? "",
       toolInput: input.tool_input ?? {},
       sessionId: this.extractSessionId(input),
-      projectDir: process.env[this.projectDirEnvVar],
+      projectDir: process.env[this.projectDirEnvVar] ?? process.cwd(),
       raw,
     };
   }
@@ -74,7 +74,7 @@ export abstract class ClaudeCodeBaseAdapter extends BaseAdapter {
       toolOutput: input.tool_output,
       isError: input.is_error,
       sessionId: this.extractSessionId(input),
-      projectDir: process.env[this.projectDirEnvVar],
+      projectDir: process.env[this.projectDirEnvVar] ?? process.cwd(),
       raw,
     };
   }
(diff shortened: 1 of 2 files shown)


mksglu:
Correcting my earlier review — after auditing all 14 adapters:

`process.cwd()` fallback is the **established convention** across the codebase. 7 adapters, all hooks (`sessionstart.mjs`, `session-helpers.mjs`), and `server.ts` already use it. `ClaudeCodeBaseAdapter` was the only outlier returning `undefined` instead.

This PR aligns the base class with the existing pattern. Apologies for the incorrect flag. Merged.

PR #290, opened 2026-04-15
fix: respect CLAUDE_CONFIG_DIR in hooks for multi-profile setups

gjaya79:
## Summary

Fixes #289

Hooks hardcoded `~/.claude` paths via `join(homedir(), ".claude", ...)`, ignoring the `CLAUDE_CONFIG_DIR` environment variable. This broke multi-profile setups where users launch Claude Code with different config directories:

```bash
alias cc="CLAUDE_CONFIG_DIR=~/.claude-personal claude"
alias ccw="CLAUDE_CONFIG_DIR=~/.claude-work claude"
```

All path resolution now checks `process.env.CLAUDE_CONFIG_DIR` first, falling back to `~/.claude` when unset. No behavior change for single-profile users.

## Changes

- **`session-helpers.mjs`**: Added `getClaudeConfigDir()` helper. Updated `getSessionDBPath()`, `getSessionEventsPath()`, `getCleanupFlagPath()` to use it for Claude opts.
- **`pretooluse.mjs`**: Self-heal block now resolves `configBase` from env var for `settings.json` and `installed_plugins.json` paths.
- **`sessionstart.mjs`**: CLAUDE.md discovery and debug log path respect the env var.
- **`precompact.mjs`**: Debug log path respects the env var.

## Test plan

- [ ] Set `CLAUDE_CONFIG_DIR=~/.claude-test` and verify session DB is created under `~/.claude-test/context-mode/sessions/`
- [ ] Verify default behavior (no env var set) still uses `~/.claude/`
- [ ] Verify self-heal in pretooluse reads correct `settings.json` and `installed_plugins.json`
- [ ] Verify CLAUDE.md is read from the correct config directory on session start

Issue #289, opened 2026-04-15
Hooks ignore CLAUDE_CONFIG_DIR — hardcode ~/.claude paths


gjaya79:
## Bug Description

context-mode hooks hardcode `~/.claude` paths instead of respecting the `CLAUDE_CONFIG_DIR` environment variable. This breaks setups where users run multiple Claude Code profiles with separate config directories.

## How to Reproduce

1. Set up two Claude Code profiles using `CLAUDE_CONFIG_DIR`:
   ```bash
   alias cc="CLAUDE_CONFIG_DIR=~/.claude-personal claude"
   alias ccw="CLAUDE_CONFIG_DIR=~/.claude-work claude"
   ```
2. Install context-mode in one profile
3. Launch Claude with either alias
4. context-mode hooks always read/write to `~/.claude/` regardless of which profile is active

## Root Cause

Multiple files hardcode the path resolution using `homedir() + ".claude"` instead of checking `process.env.CLAUDE_CONFIG_DIR`:

**`hooks/session-helpers.mjs`** — `CLAUDE_OPTS` hardcodes `configDir: ".claude"`:
```javascript
const CLAUDE_OPTS = {
  configDir: ".claude",
  projectDirEnv: "CLAUDE_PROJECT_DIR",
  sessionIdEnv: "CLAUDE_SESSION_ID",
};
```

Used in `getSessionDBPath()`, `getSessionEventsPath()`, `getCleanupFlagPath()` — all resolve to `join(homedir(), ".claude", ...)`.

**`hooks/pretooluse.mjs`** — self-heal block:
```javascript
const settingsPath = resolve(homedir(), ".claude", "settings.json");
const ipPath = resolve(homedir(), ".claude", "plugins", "installed_plugins.json");
```

**`hooks/sessionstart.mjs`** — similar hardcoded paths.

## Proposed Fix

Read `CLAUDE_CONFIG_DIR` with fallback:

```javascript
import { basename } from "node:path";

const CLAUDE_OPTS = {
  configDir: process.env.CLAUDE_CONFIG_DIR
    ? basename(process.env.CLAUDE_CONFIG_DIR)
    : ".claude",
  projectDirEnv: "CLAUDE_PROJECT_DIR",
  sessionIdEnv: "CLAUDE_SESSION_ID",
};
```

For `pretooluse.mjs` self-heal:
```javascript
const configBase = process.env.CLAUDE_CONFIG_DIR || resolve(homedir(), ".claude");
[...]

PR #367, opened 2026-04-27
feat: unified persistent memory & timeline search

mksglu:
## Summary

- Unified timeline search: `ctx_search(sort: "timeline")` across 3 sources (ContentStore + SessionDB + auto-memory)
- Auto-injection on compaction: `<behavioral_directive>`, `<rules>`, `<active_skills>` (~500 tok, fires once)
- Static reinforcement: 12 adapter configs + routing block (~99 tok, session start)
- 3 bug fixes: role snapshot, skill priority, pi-extension filter
- 12 new event categories (Phase 1 + Phase 2)
- 42 files changed, +3942/-424 lines

## Details

Full PRD: `PRD-knowledge-categories.md` — 8 review rounds, 16+ engineers per round.

### Pillar 1: Unified Timeline Search
- `sort="relevance"` (default) = ContentStore only, BM25 ranked — zero behavior change
- `sort="timeline"` = 3 sources merged chronologically
- Error isolation: try/catch per source, partial results always returned

### Pillar 2: Auto-Injection on Compaction
- `buildAutoInjection()` fires ONLY on `SessionStart(source="compact")`
- 500 token hard cap with priority: role > decisions > skills > intent
- NO mid-session injection — context-mode does not pollute the context it protects

### Bug Fixes
- `snapshot.ts`: missing `case "role"` — role events silently dropped
- `extract.ts`: skill priority P3 → P2
- `pi-extension.ts`: minPriority 2 → 3

### New Categories
Phase 1: `user-prompt`, `compaction`, `rejected-approach`, `session-resume`, `constraint`, `knowledge-reuse`
Phase 2: `agent-finding`, `error-resolution`, `external-ref`, `blocked-on`, `iteration-loop`, `latency`

## Test plan

1905 tests pass, 0 failures. TDD-first — all new code has tests in existing test files.


src/search/unified.ts
```diff
@@ -0,0 +1,141 @@
...
+      const memResults = searchAutoMemory([query], limit, projectDir, configDir);
+      results.push(...memResults);
+    } catch {
+      // Auto-memory search failed — continue
+    }
+  }
+
+  // ── Sort ──
+  if (sort === "timeline") {
+    results.sort((a, b) => (a.timestamp || "").localeCompare(b.timestamp || ""));
```
ousamabenyounes, line 135: **Bug — timeline sort doesn't merge chronologically.** ContentStore (source 1) and auto-memory (source 3) push results without a `timestamp` (the `SearchResult` type doesn't carry one). Repro:
```
store-1 → ts=(undef)
store-2 → ts=(undef)
mem-1   → ts=(undef)
db-old  → 2026-01-01
db-new  → 2026-04-27
```
All three sources collapse to the start in lex order before any DB hit, so `sort="timeline"` returns ContentStore + auto-memory first, then DB chronologically — not actually merged.

**Fix:** surface `created_at` from ContentStore chunks (already in DB) into `SearchResult.timestamp`, and fall back to `Date.now().toISOString()` for current-session items if no per-chunk timestamp is available.

src/search/unified.ts
```diff
@@ -0,0 +1,141 @@
...
+          try {
+            const reuseEvent: SessionEvent = {
+              type: "search_hit_prior",
+              category: "knowledge-reuse",
+              data: `Timeline search found ${dbResults.length} results from prior sessions`,
+              priority: 3,
+              data_hash: "",
+            };
+            sessionDB.ensureSession(sessionId, projectDir || "");
+            sessionDB.insertEvent(sessionId, reuseEvent, "ctx_search");
```
ousamabenyounes: **Read path mutates state.** This is more of an architectural concern than a perf one — better-sqlite3 inserts are ~20 µs each so the cost is real but not painful. Two things still bother me about the shape:

1. **Read-paths writing.** `ctx_search` is otherwise pure; embedding `ensureSession` + `insertEvent` here means "search" now has a write side-effect that callers don't expect. Makes reasoning about the call (and testing it) harder than it needs to be.
2. **Self-feeding pattern.** The `knowledge-reuse` event lands in the search corpus and would itself be returned by the next timeline search. Today's `data` text is fixed so it doesn't actually loop — but any future change that makes `data` query-dependent would. Quietly fragile.

**Simplest fixes**, in increasing complexity:
- Just drop the write here. The same metric can be reconstructed offline from existing logs / event counts if needed.
- Or accumulate a counter in-memory in the server process and write a single summary event when the session ends.

Both avoid mutating state from a read path and remove the self-feeding risk.

src/server.ts
```diff
@@ -1320,7 +1331,36 @@ server.registerTool(
...
+            results = searchAllSources({
+              query: q,
+              limit: effectiveLimit,
+              store,
+              sort,
+              source,
+              contentType,
+              sessionDB: timelineDB,
+              projectDir: getProjectDir(),
+              configDir: ".claude",
```
ousamabenyounes: **Two regressions on this single line.**

1. **`CLAUDE_CONFIG_DIR` ignored.** PR #290 plumbed `CLAUDE_CONFIG_DIR` through hooks; the server-side path now hardcodes `".claude"`, breaking multi-profile setups (#289).
2. **Relative path bug.** `auto-memory.ts:42-56` does `join(configDir, "CLAUDE.md")` and `join(configDir, "memory")`. Passing the literal `".claude"` resolves relative to `process.cwd()` — usually the project dir, not `$HOME`. Net effect: silently looks at `<projectDir>/.claude/...` instead of `~/.claude/...`. Auto-memory never finds user-level files even on the default profile.

**Fix:** use the existing helper:
```ts
import { resolveConfigDir } from "./hooks/session-helpers.mjs";
// …
configDir: resolveConfigDir(),
```

src/server.ts
```diff
@@ -1320,7 +1331,36 @@ server.registerTool(
...
-        const results = store.searchWithFallback(q, effectiveLimit, source, contentType);
+        let results;
+        if (sort === "timeline") {
+          // Timeline mode: unified search across ContentStore + SessionDB + auto-memory
+          let timelineDB: InstanceType<typeof SessionDB> | null = null;
+          try {
+            const sessionsDir = join(homedir(), ".claude", "context-mode", "sessions");
+            const dbFile = join(sessionsDir, `${hashProjectDir()}.db`);
+            if (existsSync(dbFile)) {
+              timelineDB = new SessionDB({ dbPath: dbFile });
```
ousamabenyounes: **DB open/close inside the per-query loop.** With 5 queries this is 5× WAL handshake + 5× prepared-statement cache rebuild. Hoist outside the loop:
```ts
let timelineDB: SessionDB | null = null;
if (sort === "timeline") {
  const dbFile = join(resolveConfigDir(), "context-mode", "sessions", `${hashProjectDir()}.db`);
  if (existsSync(dbFile)) timelineDB = new SessionDB({ dbPath: dbFile });
}
try {
  for (const q of queries) { /* pass timelineDB to searchAllSources */ }
} finally {
  timelineDB?.close();
}
```
Same `homedir() + ".claude"` hardcoding here as line 1356 — `resolveConfigDir()` is the right call.

src/search/auto-memory.ts
```diff
@@ -0,0 +1,115 @@
...
+      }
+    } catch { /* best-effort */ }
+  }
+
+  // Search each candidate file for matching queries
+  for (const candidate of candidates) {
+    if (results.length >= limit) break;
+
+    try {
+      const content = readFileSync(candidate.path, "utf-8");
```
ousamabenyounes, line 88: **No size guard before `readFileSync`.** A user with a 10–50 MB `CLAUDE.md` (real case: an agent accidentally writes logs there) blocks the search call thread + risks OOM on the subsequent `.toLowerCase()` of the buffer.

```ts
import { statSync } from "node:fs";
// …
try {
  if (statSync(candidate.path).size > 1_000_000) continue;   // skip > 1 MB
  const content = readFileSync(candidate.path, "utf-8");
  // …
}
```

src/session/db.ts
```diff
@@ -333,6 +334,16 @@ export class SessionDB extends SQLiteBase {
     p(S.deleteMeta, `DELETE FROM session_meta WHERE session_id = ?`);
     p(S.deleteResume, `DELETE FROM session_resume WHERE session_id = ?`);
 
+    // ── Search ──
+    p(S.searchEvents,
```
ousamabenyounes, line 338: **No FTS5 index on `session_events`.** This is `LIKE '%q%'` over a column with no index → full table scan. Microbench (no-match worst case, in-memory better-sqlite3):
```
  1k events  →   0.17 ms
 10k events  →   1.5  ms
 50k events  →   8    ms
200k events  →  34    ms
```
Fine at current scale, degrades linearly. The rest of the codebase uses FTS5 (ContentStore) — keeping consistency would also help here.

**Suggestion:** add an FTS5 virtual table mirroring `data + category`, keep this LIKE path as fallback for legacy DBs. Not strictly blocking for this PR, but worth flagging before this is the load-bearing path for cross-session search.

src/search/unified.ts
```diff
@@ -0,0 +1,141 @@
...
+        source: r.source,
+        origin: "current-session" as const,
+        rank: r.rank,
+        matchLayer: r.matchLayer,
+        highlighted: r.highlighted,
+        contentType: r.contentType,
+      })),
+    );
+  } catch {
+    // ContentStore search failed — continue with other sources
```
ousamabenyounes: **Silent error swallowing pattern (one of 4 in this file, 2 more in `auto-memory.ts`).** Empty `catch {}` hides corrupted DB, schema mismatches, EACCES — user sees zero results with no signal. At minimum:
```ts
} catch (err) {
  if (process.env.DEBUG?.includes("context-mode")) console.error("unified.searchAllSources:", err);
}
```

src/search/auto-memory.ts
```diff
@@ -0,0 +1,115 @@
...
+      const contentLower = content.toLowerCase();
+
+      for (const query of queries) {
+        if (results.length >= limit) break;
+
+        const queryLower = query.toLowerCase();
+        // Split query into terms, match if any term is found
+        const terms = queryLower.split(/\s+/).filter(t => t.length >= 2);
+        const matched = terms.some(term => contentLower.includes(term));
```
ousamabenyounes: **Substring match too loose.** 2-char threshold + `String.includes()` — query `"if"` matches `specIfic`, `defInItIon`, etc. Also no relevance ranking (`results.slice(0, limit)` cuts arbitrarily).

**Suggestion:** word-boundary match + count-based ranking:
```ts
const escape = (s) => s.replace(/[.*+?^${}()|[\]\\]/g, "\\$&");
const termRegex = new RegExp("\\b(" + terms.map(escape).join("|") + ")\\b", "gi");
const matches = (content.match(termRegex) || []).length;
if (matches > 0) candidates.push({ ..., matches });
// sort by matches desc before slicing
```

src/search/auto-memory.ts
```diff
@@ -0,0 +1,115 @@
...
+        if (matched) {
+          // Extract a relevant section around the first match
+          const firstTermIdx = terms.reduce((best, term) => {
+            const idx = contentLower.indexOf(term);
+            return idx >= 0 && (best < 0 || idx < best) ? idx : best;
+          }, -1);
+
+          const snippetStart = Math.max(0, firstTermIdx - 200);
+          const snippetEnd = Math.min(content.length, firstTermIdx + 500);
+          const snippet = content.slice(snippetStart, snippetEnd).trim();
```
ousamabenyounes: **Snippet boundaries are byte-based, not markdown-aware.** Cuts mid-word and can split a fenced code block, leaving an unbalanced `\`\`\`` marker in the result. Walk back to nearest `\n\n` (or markdown header) for `snippetStart`, walk forward to nearest `\n` for `snippetEnd`, and skip the snippet if a `\`\`\`` open isn't matched by a close inside the window.

ousamabenyounes (changes requested):
Substantial PR — 65 files / +4753 / -581 (PR body undercounts at 42 / +3942). 3 bug fixes are sound, auto-injection wiring is correct (compact-only, 500-tok cap, P1-P4 priority order). Tested locally on this branch: 1898 pass, 0 regressions vs `next` (the 2 `cli.test.ts` ABI failures pre-exist on `next` and are unrelated to this PR).

## Blocking concerns

1. **Timeline sort is broken** — ContentStore + auto-memory results have no `timestamp`, all collapse to the start of the merged list. Reproduced. The PR claim "chronological merge of ContentStore + SessionDB + auto-memory" doesn't hold.
2. **Read path mutates state** — `unified.ts` writes a `knowledge-reuse` event to SessionDB on every successful timeline search. Write amplification + risk of self-feeding pattern.
3. **Regression of #289 / #290** — `configDir: ".claude"` hardcoded in `server.ts`, ignores `CLAUDE_CONFIG_DIR`, and is passed as a **relative path** that `auto-memory.ts` then `join`s relative to `process.cwd()` instead of the user home — so user-level CLAUDE.md / memory/ are never seen.
4. **DB open/close per query** in the `ctx_search` loop — unnecessary WAL handshake amplification on multi-query batches.
5. **No size guard on `readFileSync`** in auto-memory — a multi-MB CLAUDE.md blocks the search call.

Major (should fix but not blocking on its own):

6. No FTS5 index on `session_events.searchEvents` — `LIKE '%q%'` full-table scan, scales linearly. Bench (no-match worst case): 1k=0.17ms, 10k=1.5ms, 50k=8ms, 200k=34ms.
7. Empty `catch {}` blocks (4× in `unified.ts`, 2× in `auto-memory.ts`) silently swallow corrupted-DB / EACCES / schema-mismatch — debugging nightmare.
8. Auto-memory match is naive `String.includes()` with 2-char threshold — `"if"` matches `specIfic`, `defInItIon`, etc. No relevance ranking.
9. Snippet extraction is byte-bounded — cuts mid-word and mid-codeblock.

Nice work overall on the architecture (clear pillar separation, error isolation per source, backward-compatible default `sort="relevance"`). Once the 5 blocking items are addressed I'd be comfortable approving. Inline notes pinpoint each.

Minor: PR body still says "42 files / +3942" and "1905 tests / 0 failures" — please refresh against final HEAD before merge so the description matches the diff.

ousamabenyounes:
Big PR — 65 files / +4753 / -581 currently on the branch (PR body says 42 / +3942, looks like it grew after the description was written). The three bug fixes look right to me, and the auto-injection wiring is exactly what I'd expect — compact-only, 500-tok cap, P1–P4 priority order. Tested locally on this branch: 1898 pass, 0 regressions vs `next` (the 2 `cli.test.ts` ABI failures already exist on `next` and aren't related to this PR).

A few things I'd love to have a closer look at before merge — none are showstoppers and most have small fixes:

## Items I'd want addressed before merge

1. **Timeline sort doesn't quite merge chronologically yet** — `ContentStore` and auto-memory results don't carry a `timestamp`, so in `sort="timeline"` they all collapse to the start of the list before any DB hit. I reproduced it locally. Suggested fix in the inline note.
2. **`unified.ts` writes to SessionDB on every successful timeline search** — more of an architectural concern than a perf one (better-sqlite3 inserts are ~20 µs, the cost itself is fine). What feels off is `ctx_search` quietly mutating state, plus the self-feeding shape: the `knowledge-reuse` event lands in the search corpus and would be returned by the next timeline search. Today's static `data` text means it doesn't actually loop, but any future query-dependent `data` would. Simplest fix is to drop the write entirely (metric can be reconstructed from logs) or write a single summary event at session end.
3. **`configDir: ".claude"` in `server.ts`** — looks like this loses the `CLAUDE_CONFIG_DIR` plumbing that #290 added, and (separately) it ends up being a relative path that `auto-memory.ts` joins against `cwd()` instead of `$HOME`. Easy to miss — `resolveConfigDir()` from `session-helpers.mjs` should cover both.
4. **`SessionDB` open/close inside the per-query loop** — opening once before the loop avoids a WAL handshake per query in batches.
5. **No size guard on `readFileSync` in auto-memory** — wanted to flag it because a few-MB `CLAUDE.md` (an agent that accidentally logs there, etc.) would block the search call. A `statSync(...).size > 1_000_000` skip would be plenty.

## Smaller things, take or leave

6. `session_events.searchEvents` uses `LIKE '%q%'` with no FTS5 index. Fine at current scale (~8 ms at 50 k events in my bench, 34 ms at 200 k), but inconsistent with the rest of the codebase which uses FTS5, so worth keeping in mind as this becomes the load-bearing path for cross-session search.
7. A handful of empty `catch {}` blocks in `unified.ts` and `auto-memory.ts` — these would silently swallow a corrupted DB or EACCES. Even a `process.env.DEBUG?.includes("context-mode") && console.error(...)` would help operators diagnose.
8. Auto-memory match uses `String.includes()` with a 2-char threshold (`"if"` matches `specIfic`, etc.). A `\b…\b` regex + match-count ranking would be tighter.
9. Snippet extraction is byte-based, so it can land mid-word or mid-codeblock — walking back to the nearest `\n\n` would smooth this out.

Really nice work on the overall shape: clear pillar separation, error isolation per source, and the backward-compatible `sort="relevance"` default means existing callers see no change. Once the first 5 are tightened up I'd be glad to flip this to approve. Inline comments pinpoint each.

Minor housekeeping: when you get a chance, the PR body still says "42 files / +3942" and "1905 tests / 0 failures" — would be nice to refresh against final HEAD so the description matches what's actually shipping.

sebastianbreguel:
Hey @mksglu, did a pass on this. Three bug fixes look right, auto-injection wiring looks right too (compact-only, 500 tok cap, P1–P4 is the order I'd expect). Inheriting Ousama's five blockers and adding a few scope and safety items.

You said one PR, fine, but here's what I'd need fixed inside it before approving, plus a few cuts I think would make it survivable in `main`.

### 1. `sort="timeline"` doesn't actually merge chronologically

ContentStore and auto-memory results carry no `timestamp`. In `localeCompare` they all collapse to position 0. Either backfill timestamps from snapshot ingest (`created_at` is already there) or drop the "timeline" wording from the docstring and rename to `sort="merged"`. The docstring as written lies, can't ship.

### 2. Read path mutates state

`unified.ts` writes a `knowledge-reuse` event on every successful timeline search. Two problems. `ctx_search` is supposed to be a read tool, and surprise side effects in read paths are how you get debugging nightmares six months out. Worse, the event lands in the corpus, so the next timeline search returns it. Right now it's masked because `data` is static text. The moment `data` becomes query-dependent, that's a recursion.

Drop the write. The metric is reconstructable from logs, you already log `ctx_search` invocations and result counts. If you really want it persisted, write a single summary event from `posttooluse` at session end, not inside the search call.

```ts
// delete from unified.ts:
if (dbResults.length > 0 && sessionId) {
  try {
    const reuseEvent: SessionEvent = { ... };
    sessionDB.ensureSession(sessionId, projectDir || "");
    sessionDB.insertEvent(sessionId, reuseEvent, "ctx_search");
  } catch { /* ... */ }
}
```

### 3. `configDir: ".claude"` regresses #289 and #290

In `server.ts` around line 1243. Two issues stacked. It drops the `CLAUDE_CONFIG_DIR` env plumbing #290 added, and it's a relative path, so when `auto-memory.ts` does `join(configDir, "CLAUDE.md")` it resolves against `process.cwd()` instead of `$HOME`. The user-level `~/.claude/CLAUDE.md` and `~/.claude/memory/*.md` are never read. The feature literally doesn't work for its stated purpose.

Use `resolveConfigDir()` from `session-helpers.mjs`:

```ts
import { resolveConfigDir } from "./session-helpers.js";
configDir: resolveConfigDir(),
```

### 4. `SessionDB` open and close per-query

The `ctx_search` loop opens and closes the DB on every query, so WAL handshake on every iteration. Fine for one query, painful for batches of five or ten.

```ts
const db = sessionDB ? new SessionDB(...) : null;
try {
  for (const q of queries) {
    const results = searchAllSources({ ..., sessionDB: db });
  }
} finally {
  db?.close();
}
```

### 5. `readFileSync` in auto-memory has no size guard

A multi-MB CLAUDE.md (rogue agent dumping into it, accidental log) blocks the search call, which blocks the LLM turn. One line:

```ts
import { statSync } from "node:fs";
if (statSync(candidate.path).size > 1_000_000) continue;
const content = readFileSync(candidate.path, "utf-8");
```

### 6. macOS CI is red

`test (macos-latest)` is failing. Don't merge red.

---

A few more I'd want fixed in this PR but they're not on their own merge-blocking.

Six empty `catch {}` blocks across `unified.ts` and `auto-memory.ts`. Corrupted DB, EACCES, JSON parse failures all get swallowed silently. Operator gets zero results and no signal why. Gate behind a debug env var across all six sites:

```ts
} catch (e) {
  if (process.env.DEBUG?.includes("context-mode")) {
    console.error("[unified.searchAllSources] ContentStore failed:", e);
  }
}
```

Auto-memory match is too loose:

```ts
const terms = queryLower.split(/\s+/).filter(t => t.length >= 2);
const matched = terms.some(term => contentLower.includes(term));
```

`includes()` with a 2-char floor means `"if"` matches `specIfic`, `defInItIon`, `notIfIcatIon`. First file with any 2-char substring wins. Bump to 3, switch to word boundary, rank by match count:

```ts
const terms = queryLower.split(/\s+/).filter(t => t.length >= 3);
const matchCount = terms.reduce((n, term) => {
  const re = new RegExp(`\\b${escapeRegex(term)}\\b`, "gi");
  return n + (content.match(re)?.length ?? 0);
}, 0);
if (matchCount === 0) continue;
```

Snippet extraction is byte-bounded. The current `slice(firstTermIdx - 200, firstTermIdx + 500)` lands mid-word, mid-codeblock, mid-list-item. Walk to the nearest `\n\n`:

```ts
let start = Math.max(0, firstTermIdx - 200);
let end = Math.min(content.length, firstTermIdx + 500);
const prevBlank = content.lastIndexOf("\n\n", start);
const nextBlank = content.indexOf("\n\n", end);
start = prevBlank >= 0 ? prevBlank + 2 : start;
end = nextBlank >= 0 ? nextBlank : end;
const snippet = content.slice(start, end).trim();
```

`searchEvents` uses `LIKE '%q%'` with no FTS5 index. Linear scan. Ousama benched 1k = 0.17ms, 10k = 1.5ms, 50k = 8ms, 200k = 34ms. Fine today, breaks at scale, and inconsistent with the rest of the codebase which already uses FTS5. Mirror the `store.ts` pattern: FTS5 virtual table over `data + type + category`, populated on insert.

The `ctx_search` description contradicts the PR claim. The PR pitches auto-injection as compact-only, but `server.ts` adds this to the tool description:

> SESSION STATE: If skills, roles, or decisions were set earlier in this conversation, they are still active. Do not discard or contradict them.

Tool descriptions ship on every tool call, that's not "once on compact". Either pull it out and keep it in `auto-injection.mjs` only, or update the PR body to acknowledge there's also a static reinforcement firing every call.

PR body stats are stale: it says 42 files / +3942 / 1905 tests. HEAD is 65 / +4753 / -581 and Ousama got 1898 plus 2 pre-existing fails. Worth `gh pr edit 367 --body` against final HEAD before merge.

---

Now the scope side, since you're keeping it one PR. At minimum I'd cut these three pieces inside it, otherwise they'll drag for a while.

Delete `PRD-knowledge-categories.md`. 557 lines, "8 review rounds × 16+ engineers per round". If it merges it lives in the repo root mismatched against shipped code within a week. PRDs belong in issues or a notebook, not `main`. The code and the tests are the spec.

Cut the 12 new categories down. Honest question: which of these are load-bearing today for a consumer that already exists? My read is `compaction` and `rejected-approach`, since snapshot and extract are already touching them. The other ten feel speculative, and each one is a maintenance commitment across 12 adapter configs. That's 144 surface points to keep coherent. I'd ship those two and open follow-up issues for the other ten, each justified by a real use case.

Same logic for the 12 adapter config touches. You're adding +18–22 lines of routing block to claude-code, codex, cursor, gemini, jetbrains, kilo, kiro, openclaw, opencode, pi, qwen, vscode all in one go. That's prompt bloat carpet-bombed into 12 environments simultaneously. I'd ship `configs/claude-code/CLAUDE.md` only, measure for a week, then roll out the other eleven in a follow-up with data.

Auto-injection itself I'd put behind a flag, default-off. 500 tokens times every compact times every session is a real ongoing cost. `CONTEXT_MODE_AUTO_INJECT=1`, opt-in for two weeks, flip the default once you've got data showing it helps.

---

Stuff I'd leave alone: the three bug fixes are correct, the pillar separation in `unified.ts` with try/catch per source and partial results is the right shape, `sort="relevance"` as default is exactly the right call (existing callers see zero change), the TDD coverage on new code is visible across `tests/core/search.test.ts` and `tests/session/*`, and the P1–P4 priority order in auto-injection (role > decisions > skills > intent) is the right precedence.

If you want a commit sequence that's bisectable inside this PR:

1. three bug fixes (~5 LOC)
2. `configDir` resolution honors `CLAUDE_CONFIG_DIR`
3. `searchAllSources` unified search, relevance default
4. timestamp backfill + drop `knowledge-reuse` write
5. `SessionDB` open once per batch
6. `searchAutoMemory` with size guard + word-boundary match
7. FTS5 index on `session_events`
8. auto-injection behind `CONTEXT_MODE_AUTO_INJECT`
9. routing block in `claude-code` adapter only
10. two categories: `compaction`, `rejected-approach`
11. PR body stat refresh
12. delete PRD

Each one revertable on its own, CI green per commit.

Once the six blockers + four follow-ups are in and the scope cuts above land, I'm good with it.

mksglu:
## All reviewer feedback addressed

| # | Blocker | Status | Commit |
|---|---------|--------|--------|
| 1 | Timeline sort broken (no timestamps) | ✅ Fixed — auto-memory: file mtime, ContentStore: sessionStartTime, SQLite: normalized to ISO | `6739ea9` |
| 2 | Read path mutates state | ✅ Fixed — knowledge-reuse write removed entirely | `134a960` |
| 3 | configDir hardcoded | ✅ Fixed — uses CLAUDE_CONFIG_DIR env var + absolute path | `134a960` |
| 4 | SessionDB per-query open | ✅ Fixed — open once before loop, close in finally | `134a960` |
| 5 | readFileSync size guard | ✅ Fixed — 1MB guard | `134a960` |
| 6 | macOS CI red | ✅ Fixed — sentinel path /tmp not tmpdir() | `5373a94` |

Additional improvements:
- Empty catch blocks → DEBUG=context-mode gated logging
- Auto-memory match: 3-char min + word-boundary regex (was 2-char includes)
- Snippet extraction: walks to nearest \n\n boundary (was mid-word cut)
- Output headers: `[prior-session | 2026-04-27 14:30 | source]`

**Updated stats**: 42 files, +4273/-584, 1902 tests pass / 0 fail.

Re: scope suggestions — the categories and adapter configs are shipping as designed per PRD. The PRD stays in the repo as architectural documentation. Auto-injection has no flag — it fires only on compact (~1-3 times per session, 500 tok each, <0.3% of context).

PR #370, opened 2026-04-28
fix: platform-aware session hooks and auto-memory for Qwen Code

btxbtxbtx:
## Summary

v1.0.100 introduced session continuity (compact/resume) and auto-memory search, but these features were hardcoded to Claude Code conventions — Qwen Code users got no benefit. This PR makes the hook layer and auto-memory platform-aware, merging with #379 (Codex memory roots) to cover all 14 adapters.

### Adapter contract — `getInstructionFiles()`
- Added `getInstructionFiles(): string[]` to `HookAdapter` interface
- Default in `BaseAdapter` returns `["CLAUDE.md"]`
- `QwenCodeAdapter` overrides to return `["QWEN.md"]`
- Future adapters declare their own (e.g., Codex: `["AGENTS.md", "AGENTS.override.md"]`)

### `src/search/auto-memory.ts` — Platform-aware file scanning
- Merged #379's `addCandidate()` + `addMemoryDir()` structure with #370's `memoryFileNames` parameter
- Scans `{memory,memories}/*.md` for both Claude/Qwen (memory) and Codex (memories) conventions
- Deduplicates via `seen` Set

### `src/server.ts` — Platform-aware config dir
- Added `getMemoryConfigDir()` — derives from adapter's `getSettingsPath()` when available
- `ctx_search` timeline uses worktree-suffixed SessionDB (`getWorktreeSuffix()`)

### `src/session/extract.ts` — Platform-aware rule detection
- Regex covers `CLAUDE.md`, `QWEN.md`, `AGENTS.md`, `AGENTS.override.md`, `GEMINI.md`, `MEMORY.md`
- Also detects `.codex/memories/`, `.qwen/memory/`, `.gemini/memory/` dirs

### Hooks — `detectPlatform()` + platform-aware paths
- `hooks/session-helpers.mjs`: Added `detectPlatform()` with `CONTEXT_MODE_PLATFORM` override env var, `QWEN_OPTS`
- `hooks/sessionstart.mjs`: Uses platform env vars, scans `QWEN.md` for Qwen Code
- `hooks/posttooluse.mjs`: Platform-aware `getSessionDBPath()`
- `hooks/precompact.mjs`: Platform-aware `resolveConfigDir()`

### Tests
- Codex auto-memory tests (`AGENTS.md`, `AGENTS.override.md`, `.codex/memories/`)
- Platform-aware config dir test for `ctx_search`
- Worktree-suffixed SessionDB test for `ctx_search` timeline
- QWEN.md rule event extraction test
- All existing tests pass (155 search, 153 session-extract, 104/106 server)

### Removed
- No bundle files in this PR (per review feedback — let maintainer's pipeline regenerate)

## Test results
- search tests: 155/155 pass
- session-extract tests: 153/153 pass
- server tests: 104/106 pass (2 pre-existing Windows shell failures)
- typecheck: pass

### DIFF
--- a/hooks/posttooluse.mjs
+++ b/hooks/posttooluse.mjs
@@ -10,7 +10,7 @@ import "./ensure-deps.mjs";
  * Must be fast (<20ms). No network, no LLM, just SQLite writes.
  */
 
-import { readStdin, parseStdin, getSessionId, getSessionDBPath, getInputProjectDir } from "./session-helpers.mjs";
+import { readStdin, parseStdin, getSessionId, getSessionDBPath, getInputProjectDir, detectPlatform } from "./session-helpers.mjs";
 import { createSessionLoaders, attributeAndInsertEvents } from "./session-loaders.mjs";
 import { dirname, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
@@ -18,20 +18,21 @@ import { readFileSync, unlinkSync } from "node:fs";
 import { tmpdir } from "node:os";
 
 // Resolve absolute path for imports — relative dynamic imports can fail
-// when Claude Code invokes hooks from a different working directory.
+// when the host IDE invokes hooks from a different working directory.
 const HOOK_DIR = dirname(fileURLToPath(import.meta.url));
 const { loadSessionDB, loadExtract, loadProjectAttribution } = createSessionLoaders(HOOK_DIR);
+const platformOpts = detectPlatform();
 
 try {
   const raw = await readStdin();
   const input = parseStdin(raw);
-  const projectDir = getInputProjectDir(input);
+  const projectDir = getInputProjectDir(input, platformOpts);
 
   const { extractEvents } = await loadExtract();
   const { resolveProjectAttributions } = await loadProjectAttribution();
   const { SessionDB } = await loadSessionDB();
 
-  const dbPath = getSessionDBPath();
+  const dbPath = getSessionDBPath(platformOpts);
   const db = new SessionDB({ dbPath });
   const sessionId = getSessionId(input);
 
--- a/hooks/precompact.mjs
+++ b/hooks/precompact.mjs
@@ -9,7 +9,7 @@ import "./ensure-deps.mjs";
  * snapshot (<2KB XML), and stores it for injection after compact.
  */
 
-import { readStdin, parseStdin, getSessionId, getSessionDBPath, resolveConfigDir } from "./session-helpers.mjs";
+import { readStdin, parseStdin, getSessionId, getSessionDBPath, resolveConfigDir, detectPlatform } from "./session-helpers.mjs";
 import { createSessionLoaders } from "./session-loaders.mjs";
 import { appendFileSync } from "node:fs";
 import { join, dirname } from "node:path";
@@ -18,7 +18,8 @@ import { fileURLToPath } from "node:url";
 // Resolve absolute path for imports
 const HOOK_DIR = dirname(fileURLToPath(import.meta.url));
 const { loadSessionDB, loadSnapshot } = createSessionLoaders(HOOK_DIR);
-const DEBUG_LOG = join(resolveConfigDir(), "context-mode", "precompact-debug.log");
+const platformOpts = detectPlatform();
+const DEBUG_LOG = join(resolveConfigDir(platformOpts), "context-mode", "precompact-debug.log");
 
 try {
   const raw = await readStdin();
@@ -27,7 +28,7 @@ try {
   const { buildResumeSnapshot } = await loadSnapshot();
   const { SessionDB } = await loadSessionDB();
 
-  const dbPath = getSessionDBPath();
+  const dbPath = getSessionDBPath(platformOpts);
   const db = new SessionDB({ dbPath });
   const sessionId = getSessionId(input);
 
--- a/hooks/session-helpers.mjs
+++ b/hooks/session-helpers.mjs
@@ -51,6 +51,14 @@ const CLAUDE_OPTS = {
   sessionIdEnv: "CLAUDE_SESSION_ID",
 };
 
+/** Qwen Code platform options. */
+export const QWEN_OPTS = {
+  configDir: ".qwen",
+  configDirEnv: undefined,        // Qwen Code has no config dir env var
+  projectDirEnv: "QWEN_PROJECT_DIR",
+  sessionIdEnv: "QWEN_SESSION_ID",
+};
+
 /** Gemini CLI platform options. */
 export const GEMINI_OPTS = {
   configDir: ".gemini",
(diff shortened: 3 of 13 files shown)


mksglu:
Hi @btxbtxbtx Let's sure 14 Adaptors is works now not just Qwen Code. Thanks.

ousamabenyounes:
Hi @btxbtxbtx, thanks for the patch — sharing a high-level pass before a proper code review.

**Pros**
- `detectPlatform()` + `QWEN_OPTS` is a clean extension of the existing `CLAUDE_OPTS` / `GEMINI_OPTS` / `CODEX_OPTS` shape — consistent with the current convention.
- Adding `memoryFileNames` as a parameter to `searchAutoMemory()` makes the API extensible without a breaking change to the existing signature.
- All four hooks (`sessionstart`, `precompact`, `posttooluse`, `session-helpers`) are wired through the same detection path — symmetric, easy to reason about.

**Concerns / warnings**

1. ⚠️ **No new tests added** for ~270 lines of runtime change across 8 files. The "qwen-code adapter: 22/22" you mention seems to be the existing suite. I couldn't find a regression test that fails without this patch. At minimum, something like:

   ```ts
   // tests/hooks/detect-platform.test.ts
   import { describe, it, expect, beforeEach, afterEach } from "vitest";
   import { detectPlatform, CLAUDE_OPTS, QWEN_OPTS } from "../../hooks/session-helpers.mjs";

   describe("detectPlatform", () => {
     const original = { ...process.env };
     beforeEach(() => { for (const k of ["QWEN_PROJECT_DIR","QWEN_SESSION_ID","GEMINI_PROJECT_DIR","CODEX_HOME"]) delete process.env[k]; });
     afterEach(() => { process.env = { ...original }; });

     it("returns QWEN_OPTS when QWEN_PROJECT_DIR is set", () => {
       process.env.QWEN_PROJECT_DIR = "/tmp/qwen";
       expect(detectPlatform()).toBe(QWEN_OPTS);
     });

     it("falls back to CLAUDE_OPTS when no platform env is set", () => {
       expect(detectPlatform()).toBe(CLAUDE_OPTS);
     });
   });
   ```

   plus a `searchAutoMemory()` test asserting that a `QWEN.md` in `projectDir` is returned when `memoryFileNames` includes it. Without these, the patch is invisible to the test suite — a future refactor can silently break Qwen support and CI stays green.

2. ⚠️ **Bundles committed alongside source.** `cli.bundle.mjs` (`+107/-107`) and `server.bundle.mjs` (`+90/-90`) double the diff and create rebase friction. Suggest one of:
   - Split into two commits: `(a) source changes`, `(b) regenerated bundles`. Easier to review and revert.
   - Or drop the bundles entirely and let the maintainer's release pipeline regenerate them. Worth pinging @mksglu on which workflow they prefer — many repos `.gitignore` build artifacts.

3. **`memoryFileNames` default `["CLAUDE.md", "QWEN.md"]` covers 2 of 14 adapters.** @mksglu's comment explicitly asks for parity across all of them. Right now Codex / Gemini / JetBrains / VSCode / Cursor / Kiro / etc. still won't see their memory files via this path. Suggest moving the file list onto the adapter contract so each one declares its own:

   ```ts
   // adapters/contract.ts
   interface AdapterContract {
     name: string;
     getSettingsPath(): string;
     getInstructionFiles(): string[];   // ["CLAUDE.md"] | ["QWEN.md"] | ["AGENTS.override.md", "AGENTS.md"]
     getMemoryDirs?(): string[];        // [".codex/memories"]
   }

   // src/search/auto-memory.ts
   export function searchAutoMemory(
     queries: string[],
     limit: number = 5,
     projectDir?: string,
     configDir?: string,
     memoryFileNames: string[] = _detectedAdapter?.getInstructionFiles() ?? ["CLAUDE.md"],
   ): AutoMemoryResult[] { … }
   ```

   That way adding the 15th adapter is "implement two methods", not "edit a constant".

4. **`detectPlatform()` resolution order is brittle.** Multiple platform envs can be set in the same shell (e.g. when launching Qwen from a Cursor terminal — both `CURSOR_CWD` and `QWEN_PROJECT_DIR` are present). The current order:

   ```js
   if (process.env.QWEN_PROJECT_DIR || process.env.QWEN_SESSION_ID) return QWEN_OPTS;
   if (process.env.GEMINI_PROJECT_DIR) return GEMINI_OPTS;
   if (process.env.VSCODE_CWD) return VSCODE_OPTS;
   if (process.env.CURSOR_CWD) return CURSOR_OPTS;
   …
   ```

   silently picks Qwen even if the actual host is Cursor. Two suggestions:
   - Prefer the **most specific** signal (session-id over project-dir over cwd-leaks-from-parent-process).
   - Add a `CONTEXT_MODE_PLATFORM` override env var so users can pin the platform when auto-detection misfires.
   - Add a test for the collision case so the precedence is documented and locked in.

5. **Overlap with #379.** Both PRs touch `src/search/auto-memory.ts` and `src/server.ts` with conflicting refactors of the candidate-collection block. Whichever lands second will need a non-trivial rebase — worth coordinating with @MarcusNeufeldt before either is merged. See the side-by-side comment I'll post on both PRs.

I'll do a proper code-level review next pass. Happy to help with the test scaffolding if useful.

btxbtxbtx:
Updated PR addressing all review feedback:

**@mksglu** — now covers all 14 adapters via `getInstructionFiles()` on the adapter contract. Each adapter declares its own instruction files (default: `["CLAUDE.md"]`). Qwen Code returns `["QWEN.md"]`. Future adapters (Codex, Gemini, etc.) can override as needed.

**@ousamabenyounes** — addressed all 5 concerns:

1. ✅ **Tests added** — Codex auto-memory tests, platform-aware config dir test, worktree-suffixed SessionDB test, QWEN.md rule event test. 155/155 search, 153/153 session-extract pass.

2. ✅ **Bundle files removed** — no `cli.bundle.mjs` or `server.bundle.mjs` in this PR. Let maintainer's pipeline regenerate.

3. ✅ **`memoryFileNames` → adapter contract** — `getInstructionFiles()` on `HookAdapter` interface. Default in `BaseAdapter`. `QwenCodeAdapter` overrides. Each adapter owns its file list.

4. ✅ **`detectPlatform()` priority** — `CONTEXT_MODE_PLATFORM` env var override (highest priority). Env var detection uses most-specific signal first (session-id > project-dir > cwd).

5. ✅ **#379 overlap resolved** — merged both PRs. `auto-memory.ts` uses #379's `addCandidate()`/`addMemoryDir()` structure + #370's `memoryFileNames` parameter. `server.ts` uses #379's `getMemoryConfigDir()` + worktree suffix. `session/extract.ts` uses #379's regex. All tests from both PRs pass together.

mksglu:
Thank you for this PR — and especially for the abstraction you proposed.

What you got exactly right:

1. **`getInstructionFiles()` on the adapter interface** — this is the right abstraction. You correctly identified that platform detection needed to be centralized rather than scattered through `if (configDir === ...)` checks.
2. **Qwen Code memory + session hook paths** — accurate against the Qwen-Code conventions.

**Why we're not merging this directly:** your PR surfaced the architectural problem clearly, but in this implementation:
- `getInstructionFiles()` was defined on the interface but only Qwen-Code actually overrode it — the other 12 adapters fell through to the default
- The interface method was never wired up downstream (`sessionstart.mjs` still had `if (configDir === ".qwen")` instead of calling `adapter.getInstructionFiles()`), so the abstraction was dead code in practice
- v1.0.100's "Unified Persistent Memory" had the same hardcoding pattern in 4 other places (`searchAutoMemory`, `extract.ts`, prior-session loading, `ctx_search` timeline) — solving it only at session-start would still leave 13 platforms broken everywhere else

**The fix:** one architectural PR that takes your `getInstructionFiles()` design and:
- Adds it (plus `getConfigDir()` and `getMemoryDir()`) to all 14 adapters with proper overrides
- Wires every memory/instruction call site through the adapter — no string comparisons left
- Removes every hardcoded `~/.claude/` and `if (configDir === ...)` branch

**What we adopted from your work:**
- Your `getInstructionFiles()` interface design (this is the spine of the new architecture)
- Your Qwen-Code paths (moved into `QwenCodeAdapter.getInstructionFiles()` / `getMemoryDir()`)
- The general pattern of adapter-owned file conventions

You'll be `Co-Authored-By` on the architectural commit and credited in the v1.0.101 release notes — your interface design is what made the whole fix possible. When the PR lands this week, please review the Qwen-Code adapter override and tell us if we got the conventions right.

Closing in favor of the unified fix. Thank you for thinking architecturally about this.

PR #376, opened 2026-04-28
feat(adaptation): enhance platform detection and cache path handling …

mikij:
…for OpenCode/KiloCode

## What / Why / How

- I want to add new argument to CLI for upgrade so it will make OpenCode/KiloCode plugin development/testing much easier allowing it to update itself from local git branch. CLI should support local upgrade (`context-mode upgrade local`) which copies current git branch instead of cloning latest GH published version.
- Plugin needs to support new cache folder location. OpenCode silently changed (not documented) where is plugin stored inside of .cache folder (before .cache/opencode/node_modules/context-mode, now .cache/opencode/packages/context-mode@latest/node_modules/context-mode), so doctor or upgrade will not properly work for new versions of OpenCode/KiloCode. KiloCode also has a bug that uses OpenCode folder location instead of own.  So `context-mode upgrade/doctor' will not work until that is [resolved](https://github.com/Kilo-Org/kilocode/issues/9503) for KiloCode. 

This fails in doctor:
```
■  Hook script exists: FAIL — not found at /Users/mickey/.cache/kilo/packages/context-mode@latest/node_modules/context-mode/hooks/pretooluse.mjs
```

Addition:

1. OpenCode now has SessionStart (via experimental.chat.messages.transform) — no longer "not yet available"
2. Both OpenCode and KiloCode now have high continuity (all 4 hooks working)
3. The AGENTS.md copy step is now optional rather than required for routing awareness

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [x] OpenCode
- [x] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Checklist

- [ ] `npm test` passes

<details>

</details>


mikij:
@mksglu this is test run result from main
image

mikij:
image
 from next

mikij:
Opened [PR](https://github.com/Kilo-Org/kilocode/pull/9740) today. It is approved so hopefully it will be merged soon.

mikij:
It was merged a several minutes ago. But it will be part of some 7.2.34+ version of KiloCode (latest is 7.2.33 and it was released just before my commit was merged)

mksglu:
Mickey — real work in here. Thank you. Let me give you a piece-by-piece take so you can see exactly what we're keeping and why.

**What you got right (adopting into the architectural PR):**

1. **OpenCode/KiloCode cache path fix** — the `packages/context-mode@latest/` cache layout change is a real, current bug. Users on OpenCode and KiloCode are hitting it. We're adopting your fix verbatim.
2. **OpenCode SessionStart via `experimental.chat.messages.transform`** — this is the correct hook surface for OpenCode and we didn't have it before. We're adopting this hook integration.

**What we're not adopting (and why — happy to discuss):**

3. **`upgrade local` subcommand** — redundant with the existing symlink workflow. `CONTRIBUTING.md` already documents `npm link` for local development, and adding a CLI subcommand for it adds surface area without solving a new problem. Open to revisiting if you've hit a case where the symlink workflow falls short.
4. **`tool-naming.ts` + `routing-block.ts` extraction** — premature abstraction. Both modules have a single consumer right now and no isolated tests. Extracting them now locks in an interface before we know the second use case. Happy to extract these the moment a second consumer appears.
5. **Bundle artifacts in the PR** — these should be CI-generated on release. We've intentionally kept hand-built bundles out of the source tree to avoid diff noise on every commit.

**Why we're not merging this directly:** your PR ran into the same root cause as #379 and #370 — v1.0.100's "Unified Persistent Memory" hardcoded `~/.claude/` across `searchAutoMemory`, `extract.ts`, prior-session loading, and `ctx_search` timeline. Rather than merge 3 platform-specific PRs (Codex, Qwen, OpenCode/KiloCode), we're doing one architectural fix: `HookAdapter` gets `getConfigDir()`, `getInstructionFiles()`, `getMemoryDir()`, and every adapter (all 14) overrides with its own conventions. Your OpenCode pieces drop into that structure cleanly.

You'll be `Co-Authored-By` on the architectural commit and called out in the v1.0.101 release notes. When the PR lands this week, I'd love your review — especially on the OpenCode adapter to confirm we got the cache path and SessionStart wiring right.

Closing in favor of the unified fix. Thanks for digging into this.

mikij:
@mksglu Have a question about this function:
```typescript
function getPlatform(): AdapterPlatformType {
  const [ kiloEnv ] = PLATFORM_ENV_VARS.filter(p => p[0]==="kilo");
  const [platform, vars] = kiloEnv;
  for (const _var of vars){
    if (process.env[_var]) {
      return platform;
    }
  }
  
  return "opencode";
}
```
I think this is more robust way to check for KiloCode platform and as you did not mention this, I want to ask did you miss it or also did not want to include for some reason?

mksglu:
> @mksglu Have a question about this function:
> 
> ```ts
> function getPlatform(): AdapterPlatformType {
>   const [ kiloEnv ] = PLATFORM_ENV_VARS.filter(p => p[0]==="kilo");
>   const [platform, vars] = kiloEnv;
>   for (const _var of vars){
>     if (process.env[_var]) {
>       return platform;
>     }
>   }
>   
>   return "opencode";
> }
> ```
> 
> I think this is more robust way to check for KiloCode platform and as you did not mention this, I want to ask did you miss it or also did not want to include for some reason?

checking

mksglu:
@mikij — you were right. Landed the full fix on \`next\` as commit [4e75d11](https://github.com/mksglu/context-mode/commit/4e75d11) with you credited as Co-Author.

## What ended up in the fix

Your point about \`getPlatform()\` (`src/opencode-plugin.ts:80-82`) was the surface. Pulling the thread surfaced a deeper problem: the canonical `PLATFORM_ENV_VARS` list itself had drift — half the entries were unverified placeholders, several platforms were entirely missing, and the plugin paradigm's fallback to \`opencode\` was blind (didn't actively check `OPENCODE` env vars either).

Audited every entry against the platform's own runtime source code. Highlights:

**Removed / corrected (no source-line evidence):**
- `kilo`: dropped bare `KILO` — Kilo-Org/kilocode `packages/opencode/src/index.ts:140` only sets `KILO_PID`, never bare `KILO`.
- `jetbrains-copilot`: dropped `IDEA_HOME`, `JETBRAINS_CLIENT_ID` — neither verifiable in any JetBrains source repo.
- `qwen-code`: dropped `QWEN_SESSION_ID` — 0 hits in QwenLM/qwen-code.
- `openclaw`: removed entirely from the env-var tier — runtime never sets `OPENCLAW_HOME`/`OPENCLAW_CLI`. Detection still works via the existing `~/.openclaw/` config-dir tier.

**Added (with verified source citations):**
- `antigravity`: `ANTIGRAVITY_CLI_ALIAS` — verified in google-gemini/gemini-cli `packages/core/src/ide/detect-ide.ts` (Google's own canonical IDE detection map).
- `zed`: `ZED_SESSION_ID`, `ZED_TERM` — verified in zed-industries/zed `crates/terminal/src/terminal.rs::insert_zed_terminal_env()`.
- `pi`: `PI_PROJECT_DIR` — confirmed by our own consumers at `src/pi-extension.ts:154` and `src/server.ts:153`.

**Reordered fork pairs so collision detection works:**
- `kilo` before `opencode` (Kilo sets `OPENCODE=1` because it's an OpenCode fork).
- `cursor` + `antigravity` before `vscode-copilot` (both inherit `VSCODE_PID`).

**`getPlatform()` itself** rewritten exactly along your line — iterates `PLATFORM_ENV_VARS` instead of hardcoding any env var. Symmetric: actively checks BOTH platforms' env vars instead of blind fallback to `opencode`. Filtered to `kilo`+`opencode` so a stray `CLAUDE_PROJECT_DIR` can't leak into the plugin's platform decision.

```typescript
function getPlatform(): AdapterPlatformType {
  for (const [platform, vars] of PLATFORM_ENV_VARS) {
    if (platform !== "kilo" && platform !== "opencode") continue;
    if (vars.some((v) => process.env[v])) {
      return platform as AdapterPlatformType;
    }
  }
  return "opencode";
}
```

## Tests
- `tests/adapters/detect.test.ts`: removed 5 broken assertions (for unverified env vars); added 4 for new platforms + a `KILO_PID + OPENCODE both set → kilo wins` fork-collision assertion.
- `tests/adapters/detect-config-dir.test.ts`: priority chain rewritten as fork-collision tests (`KILO/OPENCODE`, `CURSOR/VSCODE`, `ANTIGRAVITY/VSCODE`, `CURSOR/CODEX`).

451/451 adapter + plugin tests pass on `next`. Typecheck clean.

Thanks again for surfacing this — both the immediate bug and the broader audit it triggered. Happy to credit you for any follow-up contributions.

PR #379, opened 2026-04-28
fix(codex): use platform-aware memory roots

MarcusNeufeldt:
## Summary

Fixes Codex support in the new unified memory/timeline search path.

The memory search feature was still using Claude-oriented defaults in two important places:

- `ctx_search(sort="timeline")` used a Claude config root for auto-memory lookup.
- Auto-memory only searched `CLAUDE.md` and `memory/*.md`, so Codex instruction and memory files were missed.

This patch makes the path handling platform-aware and adds Codex instruction/memory coverage.

## What changed

- Derive timeline auto-memory config root from the detected adapter settings path.
- Respect `CODEX_HOME` for Codex config-root lookup.
- Search Codex instruction files:
  - `AGENTS.override.md`
  - `AGENTS.md`
- Search Codex memory files:
  - `.codex/memories/*.md`
- Preserve existing Claude-style discovery:
  - `CLAUDE.md`
  - `memory/*.md`
- Use the same worktree-suffixed SessionDB filename convention in `ctx_search(sort="timeline")` that `ctx_stats` and `ctx_purge` already use.
- Treat Codex `AGENTS.md`, `AGENTS.override.md`, and `.codex/memories/*.md` reads as durable rule events.

## Why it matters

Without this, Codex users could get current-session timeline results and Claude auto-memory results, but miss their own Codex memory and instruction files. Worktree sessions could also silently miss prior-session events because `ctx_search` looked for the unsuffixed SessionDB.

## Non-goals

- No Qwen behavior changes.
- No broad rewrite of Claude `.claude/` rule extraction behavior.
- No changes to non-timeline relevance search.

## Verification

```bash
bun run test tests/core/search.test.ts tests/session/session-extract.test.ts
```

Result: `307 passed`

```bash
bun x vitest run tests/core/server.test.ts -t ctx_search
```

Result: `2 passed`

```bash
bun run typecheck
```

Result: passed

```bash
git diff --check
```

Result: clean

## Local caveat

The full `tests/core/server.test.ts` file still has unrelated Python runtime failures on this Windows machine (`python3 ENOENT` / Python output cap path). The new `ctx_search` server tests pass directly.

### DIFF
--- a/configs/codex/AGENTS.md
+++ b/configs/codex/AGENTS.md
@@ -1,6 +1,6 @@
 # context-mode — MANDATORY routing rules
 
-context-mode MCP tools and Codex hooks are available. Rules protect context window from flooding. Hooks capture tool/session events; these instructions keep the model routing correctly. Follow strictly.
+context-mode MCP tools available. Rules protect context window from flooding. One unrouted command dumps 56 KB into context. Codex CLI has NO hooks — these instructions are ONLY enforcement. Follow strictly.
 
 ## Think in Code — MANDATORY
 
@@ -61,9 +61,10 @@ Session history is persistent and searchable. On resume, search BEFORE asking th
 |------|---------|
 | What were we working on? | `ctx_search(queries: ["summary"], source: "compaction", sort: "timeline")` |
 | What did we decide? | `ctx_search(queries: ["decision"], source: "decision", sort: "timeline")` |
-| What did the user ask? | `ctx_search(queries: ["user prompt request"], source: "user-prompt", sort: "timeline")` |
 | What NOT to repeat? | `ctx_search(queries: ["rejected"], source: "rejected-approach")` |
 | What constraints exist? | `ctx_search(queries: ["constraint"], source: "constraint")` |
+
+Note: user-prompt history not available.
 
 DO NOT ask "what were we working on?" — SEARCH FIRST.
 If search returns 0 results, proceed as a fresh session.
--- a/configs/codex/AGENTS.md
+++ b/configs/codex/AGENTS.md
@@ -1,6 +1,6 @@
 # context-mode — MANDATORY routing rules
 
-context-mode MCP tools available. Rules protect context window from flooding. One unrouted command dumps 56 KB into context. Codex CLI has NO hooks — these instructions are ONLY enforcement. Follow strictly.
+context-mode MCP tools and Codex hooks are available. Rules protect context window from flooding. Hooks capture tool/session events; these instructions keep the model routing correctly. Follow strictly.
 
 ## Think in Code — MANDATORY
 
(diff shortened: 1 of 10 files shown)


mksglu:
Let's do that each 14 Adaptors not only Codex @MarcusNeufeldt

ousamabenyounes (approved):
Clean PR. `getMemoryConfigDir()` with adapter detection is the right shape, `searchAutoMemory` picks up size guard / word-boundary regex / snippet walk-back / DEBUG stderr in one pass, and `mtime.toISOString()` on auto-memory results helps the timeline merge. Hoisting `timelineDB` outside the loop is a nice perf detail. Worktree-suffixed DB matches what `ctx_stats` / `ctx_purge` already use.

Tested locally: 1904 pass, 0 regressions; typecheck clean. 7 new tests cover the new paths.

Follow-up (out of scope here): `SearchResult` from ContentStore still has no `timestamp`, so current-session items collapse to the start in `sort="timeline"` even with this PR landed. Surfacing `created_at` from chunks would close the loop end-to-end.

Nit: `getMemoryConfigDir()` early fallback hits `process.env.CODEX_HOME` even on Claude if it's set globally — rare, adapter detection takes over after init.

LGTM.

mksglu:
Marcus — thank you for this. Genuinely.

Two things you nailed that we'd missed:

1. **`ctx_search` timeline mode hardcoding `~/.claude/projects/`** — that line in `server.ts` was breaking timeline search for every non-Claude user. You found it.
2. **Worktree SessionDB filename mismatch at `server.ts:1333`** — missing `getWorktreeSuffix()` call. Subtle, real bug. We've adopted your fix verbatim into the architectural PR.
3. The Codex AGENTS.md / `.codex/memories` paths you added are correct and match the codex-rs convention.

**Why we're not merging this directly:** your PR exposed that v1.0.100's "Unified Persistent Memory" was implemented Claude-centric across the whole stack — `searchAutoMemory()`, `extract.ts`, prior-session loading, `ctx_search` timeline — all hardcoded `~/.claude/`. Your Codex fix is correct, but the same root cause exists for 11 other adapters (Cursor, Windsurf, Gemini, Qwen, OpenCode, KiloCode, Crush, Roo, Cline, Amp, Aider). Merging 3 platform-specific PRs would leave the architecture fragmented.

**The fix:** one PR that adds `getConfigDir()`, `getInstructionFiles()`, `getMemoryDir()` to the `HookAdapter` interface. All 14 adapters override with their own conventions. Every memory/timeline call dispatches via adapter — no more hardcoding, ever.

**What we adopted from your work:**
- The `ctx_search` timeline dispatch fix (your insight)
- Worktree filename fix at `server.ts:1333` (your fix, unchanged)
- Codex AGENTS.md + `.codex/memories` paths (your paths, moved into the new `CodexAdapter.getMemoryDir()`)

You'll be `Co-Authored-By` on the architectural commit and called out in the v1.0.101 release notes. When the PR lands (this week), I'd love your review — especially on the Codex adapter override to make sure we got the conventions right.

Closing this PR in favor of the unified fix. Thank you for the careful diagnosis.

PR #380, opened 2026-04-28
fix(hooks): passthrough on `ask` in headless mode (CLAUDE_CODE_HEADLESS)

movax01h:
## Problem

`claude --print` (headless mode) has no TTY to surface a permission prompt. When the routing layer in `hooks/core/routing.mjs` returns `{ action: "ask" }`, the existing `hooks/formatters/claude-code.mjs` formatter emits `permissionDecision: "ask"` and the CLI hangs forever waiting on a user verdict that can never arrive.

This makes context-mode unusable for any CI/automation/agent workflow that drives Claude Code via `--print` whenever a security pattern matches an `ask` rule (sudo, env-file paths, etc.). We hit this on a self-hosted Claude agent fleet — every agent that triggered an ask pattern stalled until manually killed. Today's only workaround is disabling all PreToolUse hooks via `enabledPlugins.*: false` in per-instance `settings.json`, which throws away the entire context-mode benefit.

## Fix

Mirror the precedent that already exists for Gemini CLI (`hooks/formatters/gemini-cli.mjs:34-36`): when a documented opt-in env var is set, return `null` (passthrough) on `ask` instead of blocking.

- New env var: `CLAUDE_CODE_HEADLESS=1`
- One-line guard in the `ask` case of `formatDecision`
- All other actions (`deny`, `modify`, `context`) unchanged
- Interactive sessions are unaffected — without the env var, behavior is byte-identical to before

Launchers running headless agents export the env var explicitly, so the change is opt-in and cannot regress existing users.

## Tests

Added 4 cases to `tests/hooks/formatters.test.ts`, in a nested `describe("when CLAUDE_CODE_HEADLESS=1 ...")` block under the existing `claude-code formatter` suite, with `beforeEach`/`afterEach` to set & restore the env var:

- `ask` + env=1 → `null` (the actual fix)
- `deny` + env=1 → unchanged
- `modify` + env=1 → unchanged
- `context` + env=1 → unchanged

Existing tests (env unset) continue to pass with no edits.

```
 Test Files  1 passed (1)
      Tests  21 passed (21)
```

## Test-suite hygiene note (FYI, not part of this PR)

While running the full suite, I noticed that `tests/hooks/integration.test.ts` spawns the hook scripts as subprocesses inheriting the parent's environment. If `CLAUDE_CODE_HEADLESS=1` happens to be set in the developer's shell (e.g. carried over from a previous run or a wrapper), spawned hooks now correctly skip the `ask` case — but tests asserting `permissionDecision: "ask"` on those subprocess outputs will fail. Not a regression introduced by this PR (those tests were already coupled to the parent shell's env), but a maintainer might want to explicitly clear the var when spawning, e.g. `env: { ...process.env, CLAUDE_CODE_HEADLESS: undefined }`.

## Compatibility

- No new deps
- No public API changes
- Semver-impact: none (opt-in env var, no behavior change without it)

### DIFF
--- a/hooks/formatters/claude-code.mjs
+++ b/hooks/formatters/claude-code.mjs
@@ -14,6 +14,12 @@
  * @param {object | null} decision - Normalized decision from routePreToolUse
  * @returns {object | null} Claude Code hook response, or null for passthrough
  */
+
+// In `claude --print` (headless) the CLI has no TTY to surface an "ask" prompt,
+// so the agent stalls forever. Launchers running headless set CLAUDE_CODE_HEADLESS=1;
+// when set, we mirror gemini-cli.mjs and passthrough on ask instead of blocking.
+const isHeadless = () => process.env.CLAUDE_CODE_HEADLESS === "1";
+
 export function formatDecision(decision) {
   if (!decision) return null;
 
@@ -28,6 +34,7 @@ export function formatDecision(decision) {
       };
 
     case "ask":
+      if (isHeadless()) return null;
       return {
         hookSpecificOutput: {
           hookEventName: "PreToolUse",
--- a/hooks/formatters/claude-code.mjs
+++ b/hooks/formatters/claude-code.mjs
@@ -18,6 +18,17 @@
 // In `claude --print` (headless) the CLI has no TTY to surface an "ask" prompt,
 // so the agent stalls forever. Launchers running headless set CLAUDE_CODE_HEADLESS=1;
 // when set, we mirror gemini-cli.mjs and passthrough on ask instead of blocking.
+//
+// We also passthrough on `deny` and `modify` in headless mode — context-mode's
+// purpose is to nudge the agent toward `ctx_*` tools to save context. In a TTY
+// session that nudge is useful: the agent reconsiders or asks the user. In
+// headless `--print` mode the agent has no UI to reconsider; a `deny` aborts
+// the run with no path forward, and `modify` silently swaps a curl for an
+// `echo "use ctx_execute"` line that produces zero stdout — which is exactly
+// what blocked the homelab daily explorer agent (Loki/Prom curl turned into
+// empty echoes, all data-fetch steps reported BLOCKED). In headless launchers,
+// the operator has already accepted the context-budget tradeoff by running
+// raw tools; let those tools through.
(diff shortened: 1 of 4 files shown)


movax01h:
Pushed an additional commit (`3dc45d4`) extending this PR's scope to also passthrough `case "deny"` and `case "modify"` when `CLAUDE_CODE_HEADLESS=1`. Same pattern as the existing `case "ask"` branch.

Real-world rationale: a Proxmox homelab running this fork on `claude --print` headless agents discovered today that v1.0.103's `routing.mjs` returns `action: "modify"` for "dangerous" curl/wget — silently rewriting the command to an `echo "use ctx_execute"`. In TTY sessions that nudge is a useful discovery surface. In headless mode the agent has no UI to reconsider; the rewritten echo runs, produces zero stdout, and downstream pipelines see silent failure. Concretely it neutered a daily Loki/Prometheus scan agent — every `curl` to the internal monitoring endpoints turned into an empty echo, all data-fetch steps reported `BLOCKED`, agent had nothing to report.

Tests: 21/21 pass via `pnpm exec vitest run tests/hooks/formatters.test.ts`. Two existing tests in the headless `describe` block that asserted "still formats deny/modify normally" are inverted to assert null passthrough, matching the existing `ask` test pattern. The `context` decision is left as-is (additionalContext is informational, doesn't block the tool).

Happy to split into a separate PR if you'd prefer to land the original `ask` fix first and review modify+deny in isolation.

caraseli02 (commented):
## Night PR Review (automated)

### Summary
Makes the Claude Code hook formatter passthrough on `ask`, `deny`, and `modify` decisions when `CLAUDE_CODE_HEADLESS=1` is set. Prevents headless agents from hanging on prompts or getting silently broken by tool modification.

### Findings
🟢 **Important fix for headless mode.** The rationale is solid — headless `--print` mode has no TTY, so `ask` hangs forever, `deny` aborts with no path forward, and `modify` silently breaks tool calls (the curl→echo example is perfect).

🟢 **Context decisions still work** — informational nudges pass through, only blocking decisions are suppressed.

🟢 **Comprehensive tests** — covers all 4 decision types (ask, deny, modify, context) in headless mode with proper env var cleanup.

### Verdict: APPROVE ✅
Well-reasoned fix for a real production pain point.

PR #385, opened 2026-04-30
fix(executor): Windows shell execution silent output and .sh popup

### LINKED ISSUE #384: [Bug]: Shell execution silent/no output on Windows + .sh file association popup
### Platform

OpenCode (also affects all platforms on Windows)

### context-mode version

1.0.103 (latest)

### Debug script output

N/A — bug is in core executor, not hooks or plugins. Verified via direct code inspection and manual testing.

### Exact prompt that triggered the bug

Any prompt that triggers `ctx_execute(language: "shell", ...)` on Windows, e.g.:
```
Run a shell command: echo "test"
```

### Full error output

Two distinct symptoms:

1. **Silent failure — no output returned:**
   ```
   ctx_execute(language: "shell", code: "echo test")
   ⎿  (no output)
   ```
   JavaScript execution works fine in the same environment:
   ```
   ctx_execute(language: "javascript", code: "console.log('test')")
   ⎿  test
   ```

2. **Git Bash window pops up on top of workspace:**
   When shell execution is triggered, a Git Bash console window appears on top of the IDE/workspace because `.sh` files are associated with `bash.exe` on the system.

### Steps to reproduce

1. Run context-mode on Windows (native, not WSL)
2. Ensure Git Bash is installed and `.sh` files are associated with it
3. Trigger any `ctx_execute` with `language: "shell"`
4. Observe: (a) no output returned, and/or (b) Git Bash window pops up

### What I tried to fix it

- Set `SHELL` environment variable in MCP config to `C:/Git/bin/bash.exe` — shell is found but execution still fails silently
- Verified `child_process.execSync` works from JavaScript sandbox — confirms the shell binary is functional
- Inspected `src/executor.ts` and `src/runtime.ts` source code

### Root Cause Analysis

Three issues in the Windows shell execution path:

1. **`#writeScript` uses `.sh` extension** — On Windows, `.sh` files with a Git Bash association trigger a visible console window when accessed.

2. **`#spawn` lacks `windowsHide: true`** — Without this option, spawned console applications on Windows may create visible windows.

3. **`buildCommand` passes script as direct argument** — `bash.exe script.sh` causes MSYS2 path-mangling and file-association side effects. The existing comment in `#spawn` even says:
   > "On Windows with Git Bash, pass the script as `bash -c "source /posix/path"` rather than `bash /path/to/script.sh`."
   
   But the code does NOT implement this — it passes the file path directly.

### Proposed Fix

See PR: three minimal changes to `src/executor.ts` and `src/runtime.ts`:

1. Write shell scripts without extension on Windows (`script` instead of `script.sh`)
2. Add `windowsHide: true` to spawn options on Windows
3. Use `bash -c "source 'filePath'"` for shell commands on Windows

### Operating System

Windows (PowerShell)

### JS Runtime

Node.js v24.3.0

auric:
## Summary

Fixes two Windows shell execution issues in `ctx_execute(language: "shell", ...)`:

1. **Silent/no output** — `bash.exe script.sh` caused MSYS2 path-mangling and stdin issues, returning empty output even when `SHELL` was set.
2. **`.sh` file association popup** — Windows file associations for `.sh` files caused a Git Bash console window to appear on top of the workspace.

Also adds support for the `SHELL` environment variable to override the auto-detected shell on all platforms.

### `src/executor.ts`
- **Shell-specific extensions on Windows** — temp scripts use the right extension for each shell:
  - Bash/sh: no extension (prevents file association popups)
  - PowerShell: `.ps1` (required for `-File` to work)
  - CMD: `.cmd`
- **Add `windowsHide: true` to spawn options** — hides console windows for spawned processes on Windows.

### `src/runtime.ts`
- **Use `bash -c "source 'path'"` for bash/sh on Windows** — avoids MSYS2 path-mangling and file-association side effects.
- **Use `-File` for PowerShell on Windows** — required for `.ps1` scripts to execute correctly.
- **Respect `SHELL` env var** — if `SHELL` is set and the executable exists, it overrides auto-detection on all platforms.

### `tests/executor.test.ts`
- Updated `buildCommand` shell test to expect 3-element array only when shell is bash-like on Windows.
- Added tests for `SHELL` env var override.

### `README.md`
- Added "Windows Shell Execution" section documenting safety measures and the `SHELL` override.

## Testing

Verified on Windows:
- **Git Bash**: `ctx_execute(language: "shell", code: "echo test")` returns `"test\n"` with no popup.
- **PowerShell**: `ctx_execute(language: "shell", code: "Write-Host test")` returns `"test\n"` with no popup.
- **SHELL override**: Setting `SHELL=C:/Git/bin/bash.exe` forces bash instead of auto-detected PowerShell.

Fixes #384

### DIFF
--- a/README.md
+++ b/README.md
@@ -1082,6 +1082,18 @@ Without hooks, one unrouted `curl` or Playwright snapshot can dump 56 KB into co
 
 See [`docs/platform-support.md`](docs/platform-support.md) for the full capability comparison.
 
+### Windows Shell Execution
+
+On Windows, `ctx_execute(language: "shell", ...)` uses Git Bash (if available) with three safety measures:
+
+1. **No `.sh` extension** — temp scripts are written without an extension to prevent Windows file associations from opening a visible Git Bash window on top of your workspace.
+2. **`bash -c "source 'path'"`** — scripts are sourced via `-c` instead of passed as a direct argument. This avoids MSYS2 path-mangling issues and file-association side effects.
+3. **`windowsHide: true`** — the spawned process runs with a hidden console window, preventing any popup from appearing.
+
+If Git Bash is not installed, context-mode falls back to `cmd.exe` or PowerShell.
+
+**Override the shell:** Set the `SHELL` environment variable (e.g., in your MCP server config) to any executable path. If the file exists, context-mode uses it instead of auto-detecting. This works on all operating systems.
+
 ## Utility Commands
 
 **Inside any AI session** — just type the command. The LLM calls the MCP tool automatically:
--- a/src/executor.ts
+++ b/src/executor.ts
@@ -168,9 +168,22 @@ export class PolyglotExecutor {
       code = `Path.wildcard(Path.join(${escaped}, "*/ebin"))\n|> Enum.each(&Code.prepend_path/1)\n\n${code}`;
     }
 
-    const fp = join(tmpDir, `script.${extMap[language]}`);
+    // On Windows, choose shell-specific extension to prevent file association popups
+    // while still allowing the shell to execute the script
+    let ext = extMap[language];
+    if (language === "shell" && isWin) {
+      const shellName = this.#runtimes.shell.toLowerCase();
+      if (shellName.includes("powershell") || shellName.includes("pwsh")) {
+        ext = "ps1";
+      } else if (shellName.includes("cmd")) {
+        ext = "cmd";
+      } else {
+        ext = ""; // bash/sh: no extension avoids file association
+      }
+    }
+    const fp = ext ? join(tmpDir, `script.${ext}`) : join(tmpDir, "script");
     if (language === "shell") {
-      writeFileSync(fp, code, { encoding: "utf-8", mode: 0o700 });
+      writeFileSync(fp, code, { encoding: "utf-8", mode: isWin ? undefined : 0o700 });
     } else {
       writeFileSync(fp, code, "utf-8");
     }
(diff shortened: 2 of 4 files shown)


mksglu:
Thanks for the report and the PR. Shipped a minimal version focused only on the two root causes in commit 9d1f44fba3dda3d1b19da86a60cdcda0ce1cdc14 on `next`:

1. `windowsHide: true` on spawn — hides the console window so stdout flows through the pipe instead of getting intercepted by a visible terminal.
2. Drop `.sh` extension from the temp file on Windows — prevents the file association popup at the source.

Did NOT include:
- `bash -c source` wrapper — changes shell semantics (subshell vs sourced env, exit code propagation), out of scope for a popup/silent-output bug
- `SHELL` env override — separate feature, deserves its own discussion
- README docs — these are implementation details

Two surgical helpers (`buildSpawnOptions`, `buildScriptFilename`) exposed for unit testing without spawn/fs mocks. Both Windows-gated; Linux/macOS unchanged.

The `SHELL` override is a good idea — please open a separate issue for it and I'll prioritize.

Closing this PR; thanks for surfacing the bug.

mksglu:
Update: shipped the additional pieces in commit 21c44d6.

Re-evaluated and incorporated the two pieces I initially deferred:
- `bash -c "source 'path'"` for MSYS2 path mangling
- `SHELL` env var override

Both shipped with TDD coverage. Credit to your PR for catching the MSYS2 angle and the override use case — your analysis was right that those needed to be there for full coverage. The fix as shipped tracks your PR closely on these two points; the only divergence is no README change (kept implementation details out of user-facing docs).

PR #390, opened 2026-05-01
Localize Pi session diagnostics

jerryfan:
The Pi adapter has a small but high-friction user-facing surface: /ctx-stats, /ctx-doctor, and the routing block message that explains why inline curl/wget/fetch was stopped.

Those messages are exactly where users debug whether context-mode is capturing events, whether resume snapshots exist, and why a command was blocked to protect the context window. This PR makes that surface optionally localizable while keeping the current English strings as fallbacks.

- No new dependency
- No behavior change without an i18n provider
- English remains the default/fallback
- Locales: ja, zh-CN, es
- Pi adapter only; no CLI/OpenClaw/Claude hook changes
- Does not include generated bundle churn

Validation:
- npm install --ignore-scripts
- npm run typecheck
- npm run build
- npm pack --dry-run

### DIFF
--- a/src/pi-extension.ts
+++ b/src/pi-extension.ts
@@ -21,6 +21,7 @@ import { extractEvents, extractUserEvents } from "./session/extract.js";
 import type { HookInput } from "./session/extract.js";
 import { buildResumeSnapshot } from "./session/snapshot.js";
 import type { SessionEvent } from "./types.js";
+import { initI18n, t } from "./pi-i18n.js";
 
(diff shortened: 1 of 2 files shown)


jerryfan:
Retracting this PR. I used locale choices that were too broad and not sufficiently supported by explicit repo/product signals. That was not a good fit for a first-touch contribution. I will come back only with a smaller, better-scoped PR using the project’s actual signals and/or a safer Western-default locale set.

Issue #369, opened 2026-04-28
fix: Windows bash shell — auto-deployed hook fails with bare `node` command (cjs/loader:1450)


ShmidtS:
## Problem

On Windows with `defaultShell: "bash"` in Claude Code settings, the auto-deployed `context-mode-cache-heal.mjs` SessionStart hook fails with:

```
Failed with non-blocking status code: node:internal/modules/cjs/loader:1450
```

## Root Cause

The hook is registered in `settings.json` with a bare `node` command:

```json
"command": "node C:\Users\shmid\.claude\hooks\context-mode-cache-heal.mjs"
```

On Windows, when Claude Code uses bash as the default shell, the bare `node` command may not resolve correctly via PATH. Git Bash (sh.exe) has different PATH resolution than CMD/PowerShell — the Node.js directory (`C:\Program Files\nodejs\`) is typically added to PATH by CMD profile but may not be available in the bash subprocess spawned by Claude Code for hooks.

## Fix

Use an absolute path to `node.exe` instead of bare `node`:

```json
"command": "\"C:\Program Files\nodejs\node.exe\" \"C:\Users\shmid\.claude\hooks\context-mode-cache-heal.mjs\""
```

## Suggested Change

In the code that auto-deploys `context-mode-cache-heal.mjs` and registers the hook in `settings.json`, replace:

```js
`node ${hookPath}`
```

with:

```js
`${process.execPath} ${hookPath}`
```

`process.execPath` returns the absolute path to the Node.js executable currently running (e.g. `C:\Program Files\nodejs\node.exe`), which is guaranteed to exist and be executable. This works cross-platform and avoids PATH-resolution issues regardless of the user's shell configuration.

## Environment

- Windows 11 Pro
- Claude Code 2.1.121
- `defaultShell: "bash"` (Git Bash)
- Node.js v25.4.0 installed at `C:\Program Files\nodejs\`
- Context-mode 1.0.100

## Workaround

Manually edit `~/.claude/settings.json` SessionStart hook command to use the absolute `node.exe` path.

mksglu:
Released in **v1.0.101**! 🎉

The fix uses `process.execPath` (absolute, quoted, forward-slashed) instead of bare `node` for all hook commands. This prevents PATH resolution failures on Windows + Git Bash.

A new `buildNodeCommand()` utility in `src/adapters/types.ts` handles cross-platform path quoting — all 8 hook-building adapters have been updated.

**Please update and test:**
```bash
[...]

PR #391, opened 2026-05-01
Localize Pi doctor status copy

jerryfan:
Reposting this as a much smaller follow-up after retracting the earlier over-broad PR.

The Pi `/ctx-stats` and `/ctx-doctor` commands are diagnostics surfaces: users read them when checking session capture, compaction, DB state, and resume state. I kept this PR limited to those Pi-only diagnostic labels instead of touching routing docs, CLI output, adapter copy, or broader status text.

What changed:
- added a tiny optional i18n bridge in `src/i18n.ts`
- registered `es`, `fr`, and `pt-BR` strings
- wrapped only Pi `/ctx-stats` and `/ctx-doctor` diagnostic labels/error states

What did not change:
- no required dependency
- no behavior change without a Pi i18n provider
- English fallback strings stay in the call sites
- no README, CLI, MCP server, hooks, or non-Pi adapter behavior changed
- easy to revert by deleting `src/i18n.ts` and the small wrappers in `src/pi-extension.ts`

Validation:
- targeted `tsc` for `src/pi-extension.ts` and `src/i18n.ts` passed after installing Pi types locally
- `npm run build` passed
- `npm test` ran; failures are from missing `better-sqlite3` native bindings on this Windows checkout, while non-SQLite tests continued to pass

If useful later, I’m happy to handle broader localization in small follow-up PRs using whatever locale set you prefer.

### DIFF
--- a/src/i18n.ts
+++ b/src/i18n.ts
@@ -1,0 +1,87 @@
+import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
+
+type Locale = "en" | "es" | "fr" | "pt-BR";
+type Params = Record<string, string | number>;
+
+const translations: Record<Exclude<Locale, "en">, Record<string, string>> = {
+  es: {
+    "ctx.stats.title": "## estadísticas de context-mode (Pi)",
+    "ctx.stats.session": "- Sesión: `{session}...`",
+    "ctx.stats.events": "- Eventos capturados: {count}",
+    "ctx.stats.compactions": "- Compactaciones: {count}",
+    "ctx.stats.breakdown": "- Desglose de eventos:",
+    "ctx.stats.age": "- Edad de la sesión: {minutes}m",
+    "ctx.stats.unavailable": "estadísticas de context-mode no disponibles (error de la BD de sesión)",
+    "ctx.noActiveSession": "context-mode: no hay sesión activa",
+    "ctx.doctor.dbPath": "- Ruta de BD: `{path}`",
+    "ctx.doctor.dbExists": "- La BD existe: {exists}",
+    "ctx.doctor.sessionId": "- ID de sesión: `{session}`",
+    "ctx.doctor.events": "- Eventos: {count}",
+    "ctx.doctor.compactions": "- Compactaciones: {count}",
+    "ctx.doctor.resume": "- Instantánea de reanudación: {status}",
+    "ctx.doctor.dbError": "- Error de consulta de BD",
+  },
+  fr: {
+    "ctx.stats.title": "## statistiques context-mode (Pi)",
+    "ctx.stats.session": "- Session : `{session}...`",
+    "ctx.stats.events": "- Événements capturés : {count}",
+    "ctx.stats.compactions": "- Compactages : {count}",
+    "ctx.stats.breakdown": "- Répartition des événements :",
+    "ctx.stats.age": "- Âge de la session : {minutes}m",
+    "ctx.stats.unavailable": "statistiques context-mode indisponibles (erreur de BD de session)",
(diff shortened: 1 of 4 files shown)


mksglu:
Thanks again @jerryfan for the thoughtful follow-up here, and for the discipline you showed in retracting #390 and coming back with a tighter, well-scoped patch. The PR description is exemplary: clear blast radius, English fallbacks preserved, easy revert path, and honest validation notes about the Windows `better-sqlite3` situation. That's a great contributor instinct and we appreciate it.

After evaluating against the current codebase and roadmap, we're going to pass on this one. Reasoning below, with file references so you can see where we're coming from:

**1. No localization infrastructure exists anywhere else in the project.**
A repo-wide search for `i18n`, `locale`, `translate` returns zero hits across `src/` (the only match is the unrelated `src/search/unified.ts`). Adding a translation layer for ~14 strings in one diagnostic surface — while the CLI (`src/cli.ts`), the MCP server (`src/server.ts`), the hooks (`hooks/*.bundle.mjs`), the routing-block messages, the OpenClaw plugin, and every other adapter (`src/adapters/{claude-code,codex,cursor,opencode}`) stay English-only — creates an inconsistent UX where `/ctx-stats` speaks Spanish but the very next routed-tool error message speaks English. That's worse than uniformly English.

**2. The Pi i18n bridge is speculative.**
`src/i18n.ts` calls `pi.events?.emit?.("pi-core/i18n/registerBundle", ...)` and `pi.events?.emit?.("pi-core/i18n/requestApi", ...)` with optional chaining all the way down. There's no public Pi API today that guarantees these events exist or that any host wires them up. The optional chaining means in practice this is a no-op for every real user, so we'd be carrying ~90 lines of dead translation tables for a hypothetical future contract.

**3. The user surface is technical-by-design.**
`buildStatsText` (`src/pi-extension.ts:92`) and the `ctx-doctor` handler (`src/pi-extension.ts:375`) print DB paths, session UUID prefixes, event counts, compaction counts, and resume-snapshot status. These are debugger outputs for power users tracing capture/compaction/resume behavior — the same audience that reads `git status` and stack traces. Translating "DB path" to "Ruta de BD" while leaving the path itself, the SQL errors, and the surrounding hook logs in English doesn't meaningfully improve accessibility.

**4. Roadmap mismatch.**
Current focus (see `36fefe2 feat(insight): session analytics — 90 metrics, 37 insights, 4 composite scores` and `e7dbf58 feat: unified persistent memory & timeline search`) is on the Insight analytics dashboard and the unified memory/timeline pillars. When localization lands, we want it to be a project-wide decision driven by real usage signals (which locales, which surfaces matter most, do we want ICU MessageFormat, do hooks need to be locale-aware, etc.) — not bolted onto one adapter's diagnostic copy.

**5. Out-of-scope changes in the diff.**
`stats.json` (the README badge counts) and the `tests/adapters/claude-code.test.ts` Windows-path tweak are unrelated to the localization goal. The test fix in particular is interesting and may be valid on its own — but it should travel as its own PR with a reproducer, not bundled with an i18n change.

**What we did take from your earlier review work:** the conversation around #390/#391 reinforced for us that the Pi adapter's diagnostic strings are the *most* user-facing surface in that adapter, which is good signal for when we eventually do scope a real localization story.

Closing this out, but please don't read it as a "no" to you — it's a "no" to localization-of-one-adapter as an architectural shape. If you want a great next contribution, two things would land easily:

- The Windows forward-slash test fix from this PR, isolated and with a one-line note linking #369/#372.
- Anything that helps the Insight dashboard (`insight/`) or the unified memory/timeline work — there's a lot of surface there and contributions are very welcome.

Thanks again for the careful, low-blast-radius approach. It's exactly how we like to see PRs scoped, and we hope you'll send more.

PR #419, opened 2026-05-04
feat(session): allow omitting data_hash and widen AttributionSource

ShmidtS:
## Summary
- \"insertEvent\" no longer requires callers to pass \"data_hash\" — it is computed internally from \"event.data\" via SHA256. This simplifies all call sites and test fixtures.
- Widen \"AttributionSource\" union to include \"env\" and \"test\" so test fixtures are valid without casting.
- Remove deprecated \"forceExit\" from vitest config (dropped in v4).

## Files changed
- src/session/db.ts
- src/session/project-attribution.ts
- vitest.config.ts


vitest.config.ts
```diff
@@ -24,6 +24,5 @@ export default defineConfig({
     // Without this, Ubuntu CI consistently fails with "Worker exited unexpectedly"
     // even though all tests pass.
     teardownTimeout: isCI ? 15_000 : 5_000,
-    forceExit: isCI,
```
mksglu, line 27: Why we changed this?

mksglu (changes requested):

vitest.config.ts
ShmidtS: This option has been removed in vitest v3/v4. If you leave it, vitest will display a warning:

Unknown option: forceExit

PR #420, opened 2026-05-04
test(types): fix TypeScript errors across test suite

ShmidtS:
## Summary
- Add tests/tsconfig.json so test files have a proper TS project (root tsconfig excludes tests).
- Add .d.mts declaration stubs for hook .mjs modules — silences TS7016 under NodeNext resolution.
- Fix test imports: add .js extensions required by NodeNext, remove obsolete @ts-expect-error.
- Windows compatibility: use \"junction\" type for symlinkSync in tests.
- Insight: remove unused React imports.
- Add necessary type casts and global declarations (Bun) for strict mode compliance.

## Files changed
- tests/tsconfig.json (new)
- hooks/**/*.d.mts (23 new stubs)
- tests/**/*.test.ts (+ benchmark.ts, fixtures)
- insight/src/components/ui/scroll-area.tsx
- insight/src/routes/knowledge_...tsx


mksglu:
Hey @ShmidtS, thank you for the careful work here. I want to be upfront and honest about why this is not the right time to merge.

### 1. Bundles are regenerated by CI

This PR includes changes to `cli.bundle.mjs` (+124/-123) and `server.bundle.mjs` (+72/-72). These files are auto-regenerated by the `bundle.yml` workflow on every push to `main`. Manual edits to these two files get silently overwritten on the next main merge and create distracting review noise in the meantime. Bundle commits should only ever come from CI.

### 2. The test runner does not need these fixes today

`npm test` runs vitest, which uses esbuild for transpilation and does not enforce TypeScript checking by default. The suite is green right now. The "TypeScript errors across test suite" this PR fixes are only visible if you run `tsc --noEmit` against `tests/`, which is not in our CI pipeline. So this PR is fixing a problem that is not currently breaking CI or local development for the maintainer.

### 3. The 23 `.d.mts` stubs add long-term maintenance surface

The declaration stubs for hook `.mjs` modules are a workaround for TS7016 under NodeNext resolution. They will need to stay in sync with every future hook signature change. The cleaner long-term fix is either (a) authoring the hooks as `.ts` and emitting `.d.ts` during build, or (b) enabling `allowJs` with `checkJs: false`. I do not want to take on the stub-maintenance cost right now while features are still moving fast.

### Real value pieces

If parts of this PR fix real bugs (the Windows `junction` handling for `symlinkSync`, the unused React imports in `insight/`), please open small focused PRs for those alone. I will review and merge those quickly.

### What happens next

This is not a no forever. When we move tests to a stricter type pipeline (likely paired with adding `tsc --noEmit` to CI), I will come back to this PR as a reference for the test fixes. Until then I want to keep the test surface minimal so the project can ship features without juggling parallel type infrastructure.

Thank you for caring about the code quality. Genuinely. We will look at this again later.

Closing for now.

Mert

PR #431, opened 2026-05-05
fix(upgrade): copy package files in inline fallback

### LINKED ISSUE #429: ctx_upgrade in-place flow drops bin/ directory — context-mode statusline fails with 'statusline script missing'
# `ctx_upgrade` in-place flow drops `bin/` directory — `context-mode statusline` then fails

## TL;DR

Running the in-place upgrade (`context-mode upgrade`, a.k.a. `ctx_upgrade` MCP tool) lands `cli.bundle.mjs`, `package.json`, etc. in the new versioned plugin-cache directory but **omits the `bin/` directory entirely**. The `statusline` subcommand requires `bin/statusline.mjs` (`cli.bundle.mjs` does `import(<thisDir>/bin/statusline.mjs)`), so after upgrade it exits with `statusline script missing: …/<version>/bin/statusline.mjs`. As a side-effect, `/opt/homebrew/lib/node_modules/context-mode` is left as a broken symlink pointing at the *old* (now-deleted) version directory.

## Reproduction

1. Have a working `context-mode` installed (e.g. via `npm install -g context-mode` at any version).
2. Run the in-place upgrade — either via the MCP tool `ctx_upgrade`, or by invoking `cli.bundle.mjs upgrade` directly.
3. Observe the upgrade reports success (e.g. `Plugin reinstalled from GitHub! — v1.0.107`).
4. After the session restarts:
   ```
   $ context-mode statusline < /dev/null
   statusline script missing: /Users/<me>/.claude/plugins/cache/context-mode/context-mode/1.0.107/bin/statusline.mjs
   exit 1
   ```

## Evidence

After running `ctx_upgrade` from v1.0.72 → v1.0.107, the plugin cache looks like this:

```
~/.claude/plugins/cache/context-mode/context-mode/
├── 1.0.103/    # left over from a previous upgrade
└── 1.0.107/    # current
```

Inside `1.0.107/`:

```
$ ls ~/.claude/plugins/cache/context-mode/context-mode/1.0.107/
BENCHMARK.md  CONTRIBUTING.md  README.md  bun.lock         configs   hooks         llms.txt       package-lock.json  scripts            skills  start.mjs   tests        vitest.config.ts
CLAUDE.md      LICENSE          build      cli.bundle.mjs   docs      llms-full.txt node_modules   package.json       server.bundle.mjs  src     stats.json   tsconfig.json
```

— **no `bin/` entry**. Same on `1.0.103/`. Yet `package.json#files` lists it explicitly:

```js
$ node -e "console.log(require('…/1.0.107/package.json').files.includes('bin'))"
true
```

The hard-coded resolution inside `cli.bundle.mjs` (`statusline` subcommand, function `YD()` in the minified bundle):

```js
function YD(){
  let t = Le(cs(), "bin", "statusline.mjs");
  yr(t) || (process.stderr.write(`statusline script missing: ${t}\n`), process.exit(1));
  import(q$(t).href).catch(e => { process.stderr.write(`statusline failed: ${e?.message ?? e}\n`); … });
}
```

`cs()` resolves to the bundle's own directory (`…/1.0.107/`). Without `bin/statusline.mjs` next to it, the subcommand cannot proceed.

This is **the same class of bug fixed in v1.0.104** (per the [release notes](https://github.com/mksglu/context-mode/releases/tag/v1.0.104)):

> `bin/` directory now packed in npm tarball (was missing pre-1.0.104, breaking `context-mode statusline` for fresh installs)

That fix landed for the **published npm tarball**. The `upgrade` subcommand evidently doesn't share that file-list — it pulls from GitHub and copies a different (smaller) set of paths into the plugin cache.

### Side effect: stale npm-global symlink

The upgrade also leaves the npm-global symlink pointing at the old (now-deleted) version dir:

```
$ ls -la /opt/homebrew/lib/node_modules/context-mode
lrwxr-xr-x 1 user staff 87 Apr 14 00:32 /opt/homebrew/lib/node_modules/context-mode
    -> ../../../../Users/<me>/.claude/plugins/cache/context-mode/context-mode/1.0.72  ← gone

$ ls -la /opt/homebrew/bin/context-mode
lrwxr-xr-x 1 user staff 47 Apr 14 00:32 /opt/homebrew/bin/context-mode
    -> ../lib/node_modules/context-mode/cli.bundle.mjs  ← chases the broken symlink above
```

`command -v context-mode` returns a path (the symlink itself exists), so callers think the tool is installed — but every actual invocation ENOENTs.

## Expected behaviour

After `context-mode upgrade` to version *v*:
1. `~/.claude/plugins/cache/context-mode/context-mode/v/` should contain **every path** listed in `package.json#files` for that version — including `bin/`.
2. `/opt/homebrew/lib/node_modules/context-mode` should resolve to the new version directory (or to a real install of the new version).
3. `context-mode statusline` should succeed.

## Workaround

`npm install -g context-mode` (which uses the published tarball, not the `upgrade` flow) lays down a complete, working install at `/opt/homebrew/lib/node_modules/context-mode/` (real directory, with `bin/statusline.mjs` present). After that the system-PATH `context-mode` works again.

```
$ npm install -g context-mode
$ ls /opt/homebrew/lib/node_modules/context-mode/bin/
statusline.mjs
$ echo '{"session_id":"…"}' | context-mode statusline
context-mode  ●  $30.59 saved across sessions  ·  preserved across compact, restart & upgrade
```

The downside: any future `ctx_upgrade` will likely re-orphan this install.

## Suggested fixes (any one is sufficient)

1. **Pack `bin/` in the upgrade payload.** The `upgrade` subcommand should iterate `package.json#files` and ensure each entry is copied into the new versioned plugin-cache directory — instead of relying on a hard-coded sub-list. (Most durable; fixes the underlying mismatch between `npm install` and `upgrade` results.)
2. **Re-link the npm-global symlink after upgrade.** Even with `bin/` missing, a symlink update would mean PATH resolution succeeds on the new version dir if it's complete — but #1 is still needed so that dir is actually complete.
3. **Defensive fallback in `cli.bundle.mjs#YD`.** When `bin/statusline.mjs` is missing, fall back to a bundled-inline implementation (the bundler could embed it via dynamic import). Preserves working `statusline` even on partially-installed targets.

I'd push for option 1 — the asymmetry between `npm install` and `context-mode upgrade` is the actual root cause, and several other paths (`build/`, `node_modules/` hooks, …) would benefit from a single source-of-truth file list.

## Environment

| | |
|---|---|
| Platform | macOS 26.4.1 (arm64) |
| Node | v25.9.0 |
| npm | 11.12.1 |
| Affected versions | v1.0.103 and v1.0.107 (both observed missing `bin/`) |
| Pre-upgrade version | v1.0.72 (working — npm tarball install) |
| Working version after `npm install -g` | v1.0.111 |
| Trigger | `ctx_upgrade` MCP tool, also `cli.bundle.mjs upgrade` |

## Repro script

```bash
# 1. Snapshot
ls ~/.claude/plugins/cache/context-mode/context-mode/<your-current-version>/bin/ 2>&1
# 2. Trigger
context-mode upgrade        # or invoke ctx_upgrade via your MCP client
# 3. Observe
ls ~/.claude/plugins/cache/context-mode/context-mode/<new-version>/bin/ 2>&1
node -e "console.log(require('…/<new-version>/package.json').files.includes('bin'))"
echo '{"session_id":"x"}' | context-mode statusline
```

Happy to test fixes if you want a specific commit/branch verified.

NgoQuocViet2001:
## Summary
- make the `ctx_upgrade` inline fallback read the cloned `package.json#files` list instead of copying a hard-coded subset
- keep the `.mcp.json` placeholder write in the inline fallback so upgraded plugin caches retain `${CLAUDE_PLUGIN_ROOT}`
- add regression coverage for keeping the statusline `bin` payload in upgrade copies

Fixes #429

## Test plan
- [x] `pnpm exec vitest run tests/core/cli.test.ts -t "statusline bin|inline fallback copies package files"`
- [x] `pnpm run typecheck`
- [x] `git diff --check`
- [ ] `pnpm exec vitest run tests/core/cli.test.ts` — local Windows/pnpm run reaches 114 passing tests, then SQLite adapter tests fail because `better-sqlite3` native bindings are unavailable after pnpm ignored build scripts (`better-sqlite3`, `esbuild`).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2675,14 +2675,11 @@ server.registerTool(
       // Inline fallback: neither CLI file exists (e.g. marketplace installs).
       // Generate a self-contained node -e script that performs the upgrade.
       const repoUrl = "https://github.com/mksglu/context-mode.git";
-      const copyDirs = ["build", "hooks", "skills", "scripts", ".claude-plugin"];
-      const copyFiles = ["start.mjs", "server.bundle.mjs", "cli.bundle.mjs", "package.json"];
-
       // Write inline script to a temp .mjs file — avoids quote-escaping issues
       // across cmd.exe, PowerShell, and bash (node -e '...' breaks on Windows).
       const scriptLines = [
         `import{execFileSync}from"node:child_process";`,
-        `import{cpSync,rmSync,existsSync,mkdtempSync}from"node:fs";`,
+        `import{cpSync,rmSync,existsSync,mkdtempSync,readFileSync,writeFileSync}from"node:fs";`,
         `import{join}from"node:path";`,
         `import{tmpdir}from"node:os";`,
         `const P=${JSON.stringify(pluginRoot)};`,
@@ -2694,15 +2691,11 @@ server.registerTool(
         `execFileSync(process.platform==="win32"?"npm.cmd":"npm",["install"],{cwd:T,stdio:"inherit",shell:process.platform==="win32"});`,
         `execFileSync(process.platform==="win32"?"npm.cmd":"npm",["run","build"],{cwd:T,stdio:"inherit",shell:process.platform==="win32"});`,
         `console.log("- [x] Built from source");`,
-        ...copyDirs.map(
-          (d) =>
-            `if(existsSync(join(T,${JSON.stringify(d)})))cpSync(join(T,${JSON.stringify(d)}),join(P,${JSON.stringify(d)}),{recursive:true,force:true});`,
-        ),
-        ...copyFiles.map(
-          (f) =>
-            `if(existsSync(join(T,${JSON.stringify(f)})))cpSync(join(T,${JSON.stringify(f)}),join(P,${JSON.stringify(f)}),{force:true});`,
-        ),
-        `console.log("- [x] Copied build artifacts");`,
+        `const pkg=JSON.parse(readFileSync(join(T,"package.json"),"utf8"));`,
+        `const items=[...(Array.isArray(pkg.files)?pkg.files:[]),"src","package.json"];`,
+        `for(const item of items){const from=join(T,item);const to=join(P,item);if(existsSync(from)){rmSync(to,{recursive:true,force:true});cpSync(from,to,{recursive:true,force:true});}}`,
+        `writeFileSync(join(P,".mcp.json"),JSON.stringify({mcpServers:{"context-mode":{command:"node",args:["\${CLAUDE_PLUGIN_ROOT}/start.mjs"]}}},null,2)+"\\n");`,
+        `console.log("- [x] Copied package files");`,
         `execFileSync(process.platform==="win32"?"npm.cmd":"npm",["install","--production"],{cwd:P,stdio:"inherit",shell:process.platform==="win32"});`,
         `console.log("- [x] Installed production dependencies");`,
         `console.log("## context-mode upgrade complete");`,
--- a/tests/core/cli.test.ts
+++ b/tests/core/cli.test.ts
@@ -27,6 +27,11 @@ describe("cli.bundle.mjs — marketplace install support", () => {
     expect(pkg.files).toContain("cli.bundle.mjs");
   });
 
+  it("package.json files field includes statusline bin", () => {
+    const pkg = JSON.parse(readFileSync(resolve(ROOT, "package.json"), "utf-8"));
+    expect(pkg.files).toContain("bin");
+  });
+
   it("package.json bundle script builds cli.bundle.mjs", () => {
     const pkg = JSON.parse(readFileSync(resolve(ROOT, "package.json"), "utf-8"));
     expect(pkg.scripts.bundle).toContain("cli.bundle.mjs");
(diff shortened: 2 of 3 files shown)


EPPCOM-Solutions (approved):
**APPROVED — SHIP**

Fix is correct and self-maintaining. The dynamic `package.json#files` read eliminates the whole class of static-list drift that caused #429. `bin/` will now always be included as long as it's in `package.json#files`.

- ✅ Root cause correctly identified (hardcoded copy list missing `bin/`)
- ✅ Fix uses right abstraction — reads `package.json#files` dynamically
- ✅ Windows path handling safe (`path.join` on single-segment POSIX strings)
- ✅ Tests guard the regression (source-text assertions, established pattern)
- ⚠️ Minor fragility: `SERVER_SOURCE.indexOf("cmd =", inlineStart + 500)` boundary in server.test.ts — not a blocker

Ready to merge.

mksglu:
@sebastianbreguel @NgoQuocViet2001 WDYT?

https://github.com/mksglu/context-mode/issues/429#issuecomment-4389097692

NgoQuocViet2001:
Thanks for looping me in.

I agree that the normal CLI `upgrade` path on `main` already reads `clonedPkg.files` dynamically, so a healed v1.0.111 install should copy `bin/` and keep `context-mode statusline` working.

I still think this PR is worth merging because it covers the separate MCP `ctx_upgrade` inline fallback in `src/server.ts`, used when neither `cli.bundle.mjs` nor `build/cli.js` exists. That fallback still had its own hard-coded copy list, so it could miss `bin/` and drift from `package.json#files` again.

So my take: #431 complements the existing CLI fix by making the fallback path self-maintaining too. No extra change needed from my side, though I can tighten the test slice boundary if you prefer before merge.

sebastianbreguel:
Right call swapping to `package.json#files` — kills the drift class behind #429, not just `bin/`. Few notes before merge.

- `src/server.ts:2694`: `[...pkg.files, "src", "package.json"]`. `package.json` makes sense for the post-copy `npm install --production`. `src/` though — `build/` is already in `pkg.files`, runtime is covered. Is `src/` load-bearing or leftover? If not, cutting it shrinks the inline payload.
- `src/server.ts:2696-2697`: copy clones `.mcp.json` (`args: ["./start.mjs"]`), then `writeFileSync` overwrites with `${CLAUDE_PLUGIN_ROOT}`. Works but order-sensitive. Cleaner: commit the `${CLAUDE_PLUGIN_ROOT}` form into the repo's `.mcp.json` so the post-write goes away, or skip `.mcp.json` in the copy filter.
- `src/server.ts:2693`: payload jumps from ~6 → ~24 entries. Worth a quick check on real upgrade size.

Stance: approve-with-nits. The `.mcp.json` reorder is the only thing I'd push on, rest are observations.

cc @mksglu

PR #436, opened 2026-05-05
perf(db): cache hot-path prepared statements in store + analytics

sebastianbreguel:
## What

Cache prepared SQLite statements on two hot paths that were re-running
`db.prepare()` against the same SQL on every call.

## Why

- `ContentStore.#refreshStaleSources()` runs on every `search()` call
  when stale-detection is active, but was calling `this.#db.prepare(SQL)`
  inline each time.
- `AnalyticsEngine.queryAll()` re-prepares 8 statements per dashboard
  refresh, plus one more in `getMcpToolUsage()`.

`db.prepare()` parses + plans SQL each call. Reusing the
`PreparedStatement` skips that work — same SQL, same plan, no behavior
change.

## How

- **`src/store.ts`** — added `#stmtRefreshStaleSources` field, prepared
  once in the constructor alongside the other 30+ cached statements,
  swapped the call site.
- **`src/session/analytics.ts`** — added 9 private optional fields to
  `AnalyticsEngine`, lazy-initialized via `??=` at first call. Lazy
  (not constructor) because the engine may be constructed against a DB
  before migrations have run; deferring the `prepare()` until the
  first method call avoids that ordering issue.
- **`tests/benchmark.ts`** — extended the existing search-path bench
  section with `benchStaleRefreshPrepare()`, mirroring the SQL shape of
  `#refreshStaleSources` against raw `better-sqlite3`.

## Test plan

- [x] `npx vitest run` — 2344 passed, 17 skipped (no behavior change)
- [x] `npm run build` — bundle clean (`cli.bundle.mjs`, `server.bundle.mjs`)
- [x] `npm run benchmark` — added stale-refresh stmt-cache section

### Bench numbers (macOS arm64, 50 sources, 2000 iters)

```
stale-refresh prepare cache (raw sqlite, 50 sources)
  ad-hoc db.prepare() per call : 16.01 µs/call
  cached PreparedStatement     : 12.98 µs/call
  speedup                      : 1.23×
  saved per search()           : 3.03 µs
```

Modest per call, but compounds over a session: ~3 µs saved per search
with stale detection, ~9 × 2 µs saved per dashboard refresh. Linux/Win
numbers will land via CI.

## Affected platforms

- ✅ macOS (verified locally)
- ⏳ Linux (CI)
- ⏳ Windows (CI)

## Checklist

- [x] Tests pass on macOS
- [x] Bundle artifacts updated (`cli.bundle.mjs`, `server.bundle.mjs`)
- [x] No public API change
- [x] No new files; bench added inline to existing `tests/benchmark.ts`


sebastianbreguel:
Closing this — end-to-end stress shows the patch doesn't earn its place.

Ran a real `ContentStore.search()` loop with 100 file-backed sources × 2000 searches (stale-detection active, so `#refreshStaleSources` fires on every call):

| state | mean wall-clock | µs/search |
|---|---|---|
| pre (main @ v1.0.111) | 117.3 ms | 58.66 |
| post (this PR) | 116.4 ms | 58.22 |
| delta | −0.9 ms | **−0.44 µs (~0.4%)** |

Within run-to-run noise. The isolated micro-bench showed ~3 µs saved per `db.prepare()` call, but `#refreshStaleSources` does a `statSync()` per file-backed source on every search — filesystem stat dominates the work and dwarfs the prepare overhead.

Honest read: cleaner code, no regression, but no measurable end-to-end win. Not worth the diff.

Moving on to opps with bigger blast radius (WAL checkpoint pragma, Windows cold-start audit). Apologies for the noise.

PR #449, opened 2026-05-06
feat: implement search throttling and related utilities

hardikrathod777:
This PR addresses three items from the architecture backlog:

1. **Refactor `src/server.ts` into tool modules** — partially complete, foundation established
2. **Optimize persistence path (pooled/lazy DB instance)** — complete
3. **Harden source filtering** — complete

---

### `src/tools/` — New domain modules

Three new modules extracted from `src/server.ts`:

| Module | What it contains | Why extracted |
|---|---|---|
| `src/tools/indexing-search.ts` | `createSearchThrottle()`, throttle constants, `coerceJsonArray()`, `coerceCommandsArray()` | Search throttle logic was inline in server.ts, untestable without booting MCP server |
| `src/tools/meta-tools.ts` | `createMinimalDbAdapter()` | Minimal DB factory was duplicated twice in server.ts |
| `src/tools/fetch-tools.ts` | `FETCH_TTL_MS`, `FETCH_PREVIEW_LIMIT` | Constants were scattered inline |

`src/server.ts` is now 85 lines shorter and imports from these modules instead of defining inline.

---

### `src/session/persist-tool-calls.ts` — Connection pooling

**Before:** Every `persistToolCallCounter()` call (fired on every MCP tool response) opened a new `SessionDB` connection and left it open.

**After:** Connections are pooled in a `Map<sessionDbPath, SessionDB>`. One connection per session, reused across all tool calls. `closePersistentToolCallDBs()` is called on graceful shutdown to checkpoint WAL files and release locks.

```
Before: 100 tool calls → 100 DB open/close cycles
After:  100 tool calls → 1 DB connection, reused
```

---

### `src/store.ts` — Source filter hardening

`#sourceFilterParam()` now escapes `%`, `_`, and `\` in LIKE patterns before passing to SQLite. Without this, a source filter like `"logs%"` would match any source starting with `"logs"` — including sources the caller didn't intend to match.

Added `sourceMatchMode: "exact"` which bypasses LIKE entirely and uses `sources.label = ?` for callers that know the exact label.

---

### `src/tools/indexing-search.ts`

The `coerceJsonArray` / `coerceCommandsArray` helpers work around a Claude Code client bug where array parameters are sent as JSON strings (`"[\"a\",\"b\"]"` instead of `["a","b"]`). These were previously inline in server.ts with a comment referencing the upstream issue. Now they're in a named module, easier to find and update when the client bug is fixed.

### `src/session/persist-tool-calls.ts`

The pool uses `sessionDbPath` as the key (derived from project hash + worktree suffix). This means:
- Different projects get different connections ✅
- Same project across restarts reuses the same path → same connection ✅
- `closePersistentToolCallDBs()` is idempotent (safe to call multiple times) ✅

Error handling: if a DB operation fails, the cached connection is evicted and the error is swallowed. Counter failures must never propagate to the MCP tool call.

### `src/store.ts`

The LIKE escape uses `\\` as the escape character (standard SQL). The prepared statements already had `ESCAPE '\\'` in the SQL — this PR adds the corresponding parameter escaping that was missing.

---

## Test results

```
Test Files  3 failed | 71 passed (74)
     Tests  41 failed | 2273 passed | 30 skipped (2344)
```

The 41 failures are pre-existing Windows environment issues (no Python in PATH, bash-specific shell tests, `which` command not available). They fail identically on the previous commit. All new tests pass.

New test files:
- `tests/session/server-tool-call-wiring.test.ts` — counter persistence + restore cycle
- `tests/store.test.ts` — source filter escaping + exact-match mode

---

## Checklist

- [x] No breaking changes
- [x] TypeScript compiles clean (`npm run typecheck`)
- [x] All existing tests pass (excluding pre-existing Windows failures)
- [x] New tests added for new behavior
- [x] `src/server.ts` is shorter, not longer
- [x] Graceful shutdown closes all DB connections


mksglu:
Thank you but atm no need we that.

Issue #443, opened 2026-05-06
Reconsidering fail-open behavior in deny-policy checks


simtel12:
## Context

The three security policy check functions — `checkDenyPolicy`, `checkNonShellDenyPolicy`, and `checkFilePathDenyPolicy` — currently catch all exceptions and return `null` (allow). The comments make clear this is a deliberate architectural choice:

```typescript
} catch {
  // Security check failed — allow through (fail-open for server,
  // hooks are the primary enforcement layer)
}
```

This issue isn't arguing that decision was wrong — it's raising a scenario where the reasoning may not hold, and asking whether fail-closed might be worth reconsidering.

## The scenario worth thinking through

The stated rationale is that hooks are the primary enforcement layer, so the MCP server can afford to fail open. That works well when hooks are correctly configured and running. But consider what happens when a user has set up deny rules in `settings.json` and that file becomes transiently unreadable — a concurrent write, a failed edit, a permissions change. In that moment, the user *believes* their deny rules are active. The hooks may or may not catch it. The MCP server silently lets execution through anyway.

The failure mode worth highlighting: the AI itself editing `settings.json` and introducing malformed JSON mid-session. That's not a contrived scenario — it's exactly the kind of thing these tools enable. A JSON parse error in `settings.json` would silently disable all three policy checks simultaneously, with no indication to the user that their configured protections are no longer active.

## Why fail-closed may be preferable here

The cost of failing closed is: an execution is blocked unexpectedly, the user sees an error message explaining why, and they fix their settings. Annoying, but recoverable.

The cost of failing open is: a user who configured deny rules to protect sensitive paths or commands gets no protection in the exact moment something went wrong with their configuration — the moment they arguably need it most.

Security defaults tend to be most valuable precisely at the boundary cases. A user who hasn't configured any deny rules is unaffected by either behavior. It's only users who *have* configured protections who are exposed by fail-open — and those are the users who had reason to want the protection in the first place.

## Suggested change

```typescript
} catch (err) {
  return trackResponse(toolName, {
    content: [{
      type: "text" as const,
      text: `Execution blocked: security policy could not be evaluated (${err instanceof Error ? err.message : String(err)}). Check that settings.json is valid.`,
    }],
    isError: true,
  });
}
```

The error message surfaces enough information to diagnose the underlying problem. Small change per function, no impact on the common path.

## The question

[...]

PR #451, opened 2026-05-06
fix(security): apply Read deny-policy to ctx_index path argument (#442)

### LINKED ISSUE #442: Security: ctx_index reads arbitrary file paths without deny-policy enforcement
## Summary

The `ctx_index` tool accepts a `path` parameter and reads the file at that path into the FTS5 full-text search database. Unlike `ctx_execute_file`, which calls `checkFilePathDenyPolicy` before accessing any file, `ctx_index` performs no path validation at all. Any file readable by the MCP server process can be indexed, and once indexed its content is queryable and returned in chunks via `ctx_search`.

## Affected Code

`src/server.ts` — `ctx_execute_file` (lines 1312–1314) correctly guards file access:

```typescript
// ctx_execute_file — has the check:
const pathDenied = checkFilePathDenyPolicy(path, "execute_file");
if (pathDenied) return pathDenied;
```

`ctx_index` (lines 1465–1476) has no equivalent check:

```typescript
// ctx_index — NO path check:
const resolvedPath = path ? resolveProjectPath(path) : undefined;
if (resolvedPath) {
  trackIndexed(fs.readFileSync(resolvedPath).byteLength);
}
const result = store.index({ content, path: resolvedPath, source: source ?? resolvedPath });
```

## Why This Is a Problem

This creates a data exfiltration path through the search index. An AI agent operating under a compromised prompt (prompt injection) or a malicious MCP client can:

1. Call `ctx_index({ path: "/Users/alice/.ssh/id_rsa" })` — the file is read and stored in the FTS5 database.
2. Call `ctx_search({ queries: ["-----BEGIN"] })` — the key material is returned in search result chunks, reaching the LLM's context window.

The same technique applies to any sensitive file: `.env` files, `~/.aws/credentials`, browser cookie databases, local keystores. The `checkFilePathDenyPolicy` function already exists and handles exactly this case — it checks the path against the configured `Read` deny patterns in `settings.json`. `ctx_index` simply never calls it.

This vulnerability is distinct from the general "sandbox has no OS-level isolation" concern. Users who have specifically configured `Read` deny patterns to protect sensitive paths — which is the documented mitigation for that concern — receive no protection via `ctx_index`.

## Suggested Fix

Add the same deny-policy check that `ctx_execute_file` uses, immediately before the `resolveProjectPath` call:

```typescript
async ({ content, path, source }) => {
  // Add this block:
  if (path) {
    const pathDenied = checkFilePathDenyPolicy(path, "ctx_index");
    if (pathDenied) return pathDenied;
  }

  // existing code continues unchanged...
  const resolvedPath = path ? resolveProjectPath(path) : undefined;
```

This is a three-line addition that mirrors the existing `ctx_execute_file` pattern exactly.

sebastianbreguel:
Closes #442.

## Problem

`ctx_index` accepts a `path` parameter and reads it into the FTS5 store, but unlike `ctx_execute_file` it does not call `checkFilePathDenyPolicy` first. Any file readable by the MCP server process could be indexed and then surfaced through `ctx_search`, exposing content like `~/.ssh/id_rsa`, `.env`, or `~/.aws/credentials` to the model. The documented `Read` deny-pattern mitigation simply did not apply on this path.

## Fix

Add the same `checkFilePathDenyPolicy` gate that `ctx_execute_file` uses, immediately after the `(!content && !path)` guard and before `resolveProjectPath`. Six lines, mirrored from the existing pattern at `src/server.ts:1312-1314`. The new check is skipped when only `content` is provided, so the inline-content branch is unaffected.

## Verification

Real-MCP integration test (`tests/core/server.test.ts`, "ctx_index: Read deny-policy enforcement (#442)") spawns the server via stdio JSON-RPC against a temp project containing `.claude/settings.json` with a `Read(./secret.env)` deny pattern, then exercises three scenarios end-to-end:

1. **Denied path**: `ctx_index({ path: "secret.env" })` returns the deny-policy error tool result (`isError: true`, body contains `blocked by security policy` and `Read deny pattern`). No "Indexed N section" success message. A follow-up `ctx_search` for the file's unique marker returns the empty-store signal — proving no chunks reached FTS5.
2. **Allowed path**: `ctx_index({ path: "public-doc.md" })` still indexes successfully and the marker is searchable.
3. **Inline content**: `ctx_index({ content: ... })` (no path) bypasses the new gate, as intended.

Full `tests/core/server.test.ts` suite: 179/179 passing. Build clean.

## Files

- `src/server.ts` — 6 LOC: `checkFilePathDenyPolicy(path, "ctx_index")` gate before `resolveProjectPath`.
- `tests/core/server.test.ts` — new `describe` block with 3 real-MCP integration tests.


mksglu:
Let's resolve conflict. Thank you! @sebastianbreguel

sebastianbreguel:
rdy

ousamabenyounes:
+1, post-merge review from my side.

Pulled `next`, ran the full suite locally (2397 pass / 25 skipped, only flake is the pre-existing pool-teardown on `tests/hooks/kiro-hooks.test.ts`, unrelated). Checked the gate placement in `ctx_index` — `if (path)` wrapper before `resolveProjectPath` does the right thing, inline-content branch stays untouched. Real-MCP tests cover the cases I'd want (rel + abs deny, glob `*.env`, `sub/../secret.env` traversal hitting `**/secret.env`, plus the inline-content bypass with a deny rule that *would* match the inline source if the gate were naive — nice).

Bonus fix worth calling out: `toForward(glob)` in `evaluateFilePath`. Without it, a Windows absolute deny like `Read(C:\Users\...\secret.env)` carried literal backslashes through `JSON.parse` and never matched the forward-slash candidate. Three unit tests pin it. Goes beyond #442 scope but it's the right place to land it.

Agree with the deferred list — `#443` fail-open and the deny→read TOCTOU both need helper-signature changes that would balloon this PR. SSRF guard already covers the parallel `ctx_fetch_and_index` vector (http/https only), so no parallel exfil hole.

Thanks for the careful work.

PR #452, opened 2026-05-06
fix(security): replace ctx_insight execSync interpolation with spawnSync + tighten port schema (#441)

### LINKED ISSUE #441: Security: execSync string interpolation enables shell injection in ctx_insight
## Summary

`ctx_insight` constructs shell commands by interpolating caller-controlled values directly into strings passed to `execSync`. Two separate variables are affected: `url` (derived from the `port` parameter) and `port` itself. The pattern in `cli.ts` already uses `execFile` with argument arrays and includes a comment explicitly calling out this risk — `server.ts` never received the same treatment.

## Affected Code

`src/server.ts` — four `execSync` call sites:

```
Line 2961: execSync(`for /f "tokens=5" %a in ('netstat -ano ^| findstr :${port}') do taskkill /F /PID %a`, ...)
Line 2963: execSync(`lsof -ti:${port} | xargs kill 2>/dev/null`, ...)
Line 2974: execSync(`open "${url}"`, ...)
Line 2976: execSync(`xdg-open "${url}" 2>/dev/null || sensible-browser "${url}" 2>/dev/null`, ...)
```

Lines 3038–3040 repeat the same `open`/`xdg-open` pattern.

## Why This Is a Problem

`execSync` with a string argument spawns `sh -c <string>` on Unix. String interpolation of any caller-supplied value into that string is shell injection. The Linux `xdg-open` variant is the most directly exploitable: the full string (including any shell metacharacters in `url`) lands verbatim in the shell.

`port` is declared as `z.coerce.number().optional()` — coercion via `Number()` strips most metacharacters, which substantially reduces the risk for `port` directly. However:

1. The `url` variable built from `port` is a string, not a number, and is used in double-quoted interpolation. A double-quote character in `url` would break out of the quoted context.
2. Defense-in-depth is lost: any future refactor that makes `url` or `host` caller-configurable is immediately exploitable with no other change required.
3. The Windows `findstr :${port}` and `xargs kill` variants carry the same structural issue.

## Suggested Fix

Replace all four `execSync` string-interpolation sites with `execFile` and argument arrays — the same approach already used in `cli.ts`:

```typescript
import { execFile } from "node:child_process";

// Open browser
if (platform === "darwin")       execFile("open", [url]);
else if (platform === "win32")   execFile("cmd", ["/c", "start", "", url]);
else                             execFile("xdg-open", [url]);
```

Also add integer and range validation to the `port` parameter:

```typescript
port: z.coerce.number().int().min(1).max(65535).optional()
```

sebastianbreguel:
Closes #441.

## Problem

`ctx_insight` built shell commands by interpolating `port` and `url` into strings passed to `execSync` — the same `sh -c <string>` pattern that `cli.ts` already standardized away from. Six call sites in `src/server.ts` were affected:

- **Port-kill (2 sites):**
  - macOS / Linux: `` execSync(`lsof -ti:${port} | xargs kill 2>/dev/null`) ``
  - Windows: `` execSync(`for /f "tokens=5" %a in ('netstat -ano ^| findstr :${port}') do taskkill /F /PID %a`) ``
- **Browser-open (4 sites, repeated for "already running" and "fresh start" branches):**
  - `` execSync(`open "${url}"`) ``
  - `` execSync(`start "" "${url}"`) ``
  - `` execSync(`xdg-open "${url}" 2>/dev/null || sensible-browser "${url}" 2>/dev/null`) ``

Direct exploitation was blunted: `port` is `z.coerce.number()` (NaN-on-injection) and `url` is constructed from `port` + a fixed `http://localhost:` prefix. But defense-in-depth was lost — any future refactor that exposed `host` or made `url` caller-configurable was immediately injectable.

## Fix

Two `spawnSync`-based helpers replace the shell calls. Both use argv arrays, no shell:

```ts
function openBrowserBestEffort(url: string): void { ... }
function killProcessOnPort(port: number): void { ... }
```

The Linux `xdg-open || sensible-browser` fallback is replicated explicitly via `spawnSync` exit-status check, so the previous best-effort semantics are preserved. The Windows port-kill walks `netstat -ano` output in JS to extract PIDs, then invokes `taskkill /F /PID <pid>` per PID — no `for /f` shell loop.

The port schema is tightened from:

```ts
port: z.coerce.number().optional()
```

to:

```ts
port: z.coerce.number().int().min(1).max(65535).optional()
```

This catches `0`, negative values, `65536+`, and non-integer values at the schema layer before the handler runs.

## Verification

Real-MCP integration tests in `tests/core/server.test.ts`:

1. **Source-grep guards** (`ctx_insight: execFile migration source guard (#441)`):
   - The `ctx_insight` handler block contains zero `` execSync(`…${…}…`) `` template-string sites.
   - No shell-quoted `open`, `start`, `xdg-open`, `sensible-browser`, `lsof`, or `for /f` invocations in the handler.
   - The port schema declaration carries `.int()`, `.min(1)`, and `.max(65535)`.
   - The two helpers exist by name.
2. **Schema rejection** (`ctx_insight: port schema rejects invalid values (#441)`): spawns the server via stdio JSON-RPC and exercises five invalid port values (`0`, `-1`, `65536`, `3.14`, `"not-a-port"`) against fresh server processes. Each is rejected before the handler runs (either as a JSON-RPC `error` envelope or a tool result with `isError: true`), and no "Dashboard running" success text leaks.

Full `tests/core/server.test.ts` suite: 187/187 passing. Build clean.

## Out of scope

Two `execSync` call sites remain in the broader file but are deliberately out of scope for this issue:

- `src/server.ts:2659/2661` — port-kill at the **fixed** port `4747` during cleanup; no caller input.
- `src/server.ts:2966/2986` — static `npm install` and `npx vite build`; no caller input.

The reporter explicitly scoped the issue to the six sites that interpolate caller-supplied values.

## Files

- `src/server.ts` — `import { spawnSync }`; two new helper functions; six `execSync` sites collapsed to two helper calls; port schema tightened.
- `tests/core/server.test.ts` — two new `describe` blocks: source-grep guards (6 tests) + schema rejection real-MCP tests (5 tests).


mksglu:
@sebastianbreguel Did you test that?

sebastianbreguel:
Yeah. New file `tests/core/process-utils.test.ts` (22 tests, mocks spawnSync):

- argv arrays only on darwin/win32/linux, opts never has `shell:true`. Feeds `http://x; rm -rf /` and asserts it lands as one argv entry — the actual security property, not the source-grep version.
- Linux fallback: status=0 → no fallback, status=1 → sensible-browser fires, signal-kill (status=null + error) → fallback also fires. Last one was the pre-fix bug.
- Windows netstat fixture with the regression case: `LISTENING TCP :4747` PID 1234, `ESTABLISHED ... 8.8.8.8:4747` PID 9876, `UDP :4747` PID 5555, IPv6 LISTENING PID 1235. Asserts only 1234 + 1235 killed. Pre-fix `line.includes(':4747')` killed 9876 and 5555.
- Per-pid kill failure: middle pid EPERM, other two still attempted.

Schema: kept the 6 invalid-port integration tests (real MCP over stdio), added a 7th that pins `error.code === -32602` so a regression to `z.coerce.number().optional()` + handler crash can't satisfy the lenient `isError === true` check. Init response shape verified before tools/call.

2375 pass locally. macos + ubuntu green, windows pending.

Not tested: real `netstat -ano` on a Windows runner (only the parser w/ fixture), and concurrent ctx_insight TOCTOU — follow-up.

mksglu:
Let's resolve conflicts. @sebastianbreguel

sebastianbreguel:
rdy

ousamabenyounes:
Two follow-ups worth opening (security fix itself is solid — verified locally: `tsc` clean, full vitest 2359 pass / 25 skipped, source-grep confirms 0 `execSync(\`…${…}…\`)` in handler, helper unit tests with mocked `SpawnSyncFn` are present and exhaustive):

**1. Missing `{ timeout: ... }` on the new `spawnSync` calls in `openBrowserSync` / `killProcessOnPort`.** A hung `lsof` / `xdg-open` / `taskkill` will block the MCP tool indefinitely. Trivial defense-in-depth (e.g. `timeout: 5000`).

**2. Windows netstat parser breaks on non-English locales.** `state !== "LISTENING"` is locale-dependent — Windows-FR shows `À l'écoute`, Windows-DE `ABHÖREN`. A whitelist of connected-state names has the same problem.

The remote-address column IS locale-independent: a listening TCP socket always has remote `0.0.0.0:0` or `[::]:0`, a connected one has a real `addr:port`. So the locale-robust predicate is:

```ts
if (proto !== "TCP") continue;
if (!local.endsWith(portSuffix)) continue;
if (remote !== "0.0.0.0:0" && remote !== "[::]:0") continue;
```

The existing Windows regression fixture in `tests/core/server.test.ts` already covers it (LISTENING IPv4/IPv6 PIDs 1234/1235 must match, ESTABLISHED 9876 + UDP 5555 must not). Drop-in change, no test rewrite needed.

Not a security regression — failing-to-kill is safer than killing-wrong — but a real UX bug on non-EN Windows boxes. Worth a follow-up issue.

(Tests-existence note for the next reviewer who reads only the merge commit: helper unit tests landed in the follow-up refactor commits `12aa096` + `f4680f9`, not in `690d5e5`. Easy miss.)

PR #460, opened 2026-05-06
fix(adapters): honor CLAUDE_CONFIG_DIR in Claude Code adapter (#453)

### LINKED ISSUE #453: Claude Code adapter ignores $CLAUDE_CONFIG_DIR — session data lands in ~/.claude
## Summary

The Claude Code adapter writes session data to `~/.claude/...` and ignores `$CLAUDE_CONFIG_DIR`, even though Claude Code itself uses `$CLAUDE_CONFIG_DIR` as the canonical config root (which may be set to a non-default location like `~/.config/claude-code/`).

The fallback in `src/server.ts` was clearly *intended* to honor `CLAUDE_CONFIG_DIR`, but it's bypassed whenever a Claude Code adapter is detected — which is always, in Claude Code.

Notably, `hooks/session-helpers.mjs` **does** honor `CLAUDE_CONFIG_DIR`, so hooks and the server adapter diverge on where they think the config root lives. This can land session-scoped state in two different trees on the same machine.

## Reproduction

1. Set `CLAUDE_CONFIG_DIR=/Users/<you>/.config/claude-code/` (i.e., anywhere other than `~/.claude`).
2. Run Claude Code with the `context-mode` MCP server (e.g., via `npx -y context-mode@latest` in `.mcp.json`).
3. Use the session for a bit (any tool calls captured by the always-on PostToolUse hook).
4. Inspect both locations.

Observed (on my machine, `CLAUDE_CONFIG_DIR=/Users/ahby/.config/claude-code/`):

- `~/.claude/context-mode/sessions/` — 5 active `.db` files, mtime today. **This is where the running server writes.**
- `/Users/ahby/.config/claude-code/context-mode/sessions/3c3fc1e0cc7a19c2.db` — single stale file, mtime several days ago. (Likely written by a different invocation path / older code.)

Expected: session data lands under `$CLAUDE_CONFIG_DIR/context-mode/sessions/` whenever `$CLAUDE_CONFIG_DIR` is set.

## Root cause

Three converging facts in `context-mode`:

**1. `src/server.ts:1662`** — adapter takes precedence over `CLAUDE_CONFIG_DIR`:

```ts
const configDir = _detectedAdapter?.getConfigDir() ?? (process.env.CLAUDE_CONFIG_DIR || join(homedir(), ".claude"));
```

The `??` fallback only fires when no adapter is detected. Under Claude Code, an adapter is always detected, so `process.env.CLAUDE_CONFIG_DIR` is dead code on this path.

**2. `src/adapters/claude-code/index.ts:53`** — adapter hardcodes `.claude`:

```ts
constructor() { super([".claude"]); }   // sessionDirSegments = [".claude"]
```

**3. `src/adapters/base.ts:29, 62`** — base adapter joins `homedir() + ".claude"` unconditionally and `ClaudeCodeAdapter` does not override `getConfigDir()`:

```ts
const dir = join(homedir(), ...this.sessionDirSegments, "context-mode", "sessions");
// getConfigDir() default:
return join(homedir(), ...this.sessionDirSegments);
```

Result: on Claude Code, `getConfigDir()` returns `homedir() + "/.claude"` regardless of `$CLAUDE_CONFIG_DIR`.

For completeness, the npx env-propagation hypothesis is **moot**: child processes inherit env by default on macOS/Linux, `npx` does not strip variables, and Claude Code's MCP launcher passes parent env through. The variable arrives — it's just never read on this path.

## Hooks vs. server divergence

`hooks/session-helpers.mjs` (~lines 97 and 152) does honor `CLAUDE_CONFIG_DIR`. Tests at `tests/hooks/integration.test.ts:632–677` already cover that contract. So the behavior is inconsistent within `context-mode` itself: hooks respect `CLAUDE_CONFIG_DIR`, the server-side adapter does not.

## Suggested fix

Override `getConfigDir()` in `src/adapters/claude-code/index.ts` so the Claude Code adapter participates in the env-var contract that the rest of the codebase already follows:

```ts
getConfigDir(): string {
  return process.env.CLAUDE_CONFIG_DIR || resolve(homedir(), ".claude");
}
```

Additionally, the other hardcoded `homedir() + ".claude"` joins in `src/adapters/claude-code/index.ts` (settings, plugin registry, version cache around lines 73, 290–296, 314, 514, 1012, 1018) should probably be routed through `getConfigDir()` so they stay consistent with the new override.

Migration note: existing users will have data under `~/.claude/context-mode/sessions/`. After the fix, fresh sessions will write to `$CLAUDE_CONFIG_DIR/context-mode/sessions/`. A one-time migration step (or a soft-fallback that reads from the legacy path if the new one is empty) would avoid surprising users with "lost" session history.

## Environment

- macOS (Darwin 25.2.0)
- Claude Code (latest)
- `context-mode` invoked via `npx -y context-mode@latest`
- `CLAUDE_CONFIG_DIR=/Users/ahby/.config/claude-code/`

Happy to send a PR if the maintainer agrees with the approach.

halindrome:
## Summary

Fixes #453.

The Claude Code adapter hardcodes `~/.claude` for the session directory, settings file, and plugin registry, ignoring `$CLAUDE_CONFIG_DIR`. This causes server-side state to land under `~/.claude/` while `hooks/session-helpers.mjs` (which already honors the env var) writes under `$CLAUDE_CONFIG_DIR/`, splitting session data across two trees on the same machine.

This PR overrides `getConfigDir()` and `getSessionDir()` in `ClaudeCodeAdapter` so settings, sessions, and the adapter-owned plugin-registry reads/writes flow through `$CLAUDE_CONFIG_DIR` when set, falling back to `~/.claude` otherwise — matching the contract the hook helpers already follow, including tilde expansion for `~/foo`-style values.

### Scope

This PR fixes the **`ClaudeCodeAdapter`** call sites only:

- `getSettingsPath()` → `<configDir>/settings.json`
- `getSessionDir()` → `<configDir>/context-mode/sessions/`
- `getInstalledVersion()` → reads `<configDir>/plugins/installed_plugins.json` first, then `~/.claude` and `~/.config/claude` as legacy fallbacks
- `updatePluginRegistry()` → writes `<configDir>/plugins/installed_plugins.json`
- Diagnostic error message now reports the resolved settings path instead of a hardcoded one

`server.ts:1662` already calls `_detectedAdapter.getConfigDir()` first, so it picks up the override automatically.

### Out of scope (follow-ups)

QA round 1 (see PR comment) identified additional `~/.claude` hardcodes outside the adapter that the same env-var contract should reach in subsequent PRs:

- `src/server.ts:147` (pre-detection session-dir fallback)
- `src/server.ts:345, 348` (`healCacheMidSession` plugin-cache reads)
- `src/cli.ts:673, 874, 1012, 1018` (marketplace, registry, statusline)
- `src/security.ts:276, 342` (global settings reads)

These are tracked separately because they touch broader subsystems (CLI install flow, security policy) and warrant their own review.

### Migration note

Users who already have data under `~/.claude/context-mode/sessions/` and have `$CLAUDE_CONFIG_DIR` set elsewhere will see fresh sessions write to the new location after upgrading. Existing DBs/events files remain on disk but become orphaned for new server reads. No automatic copy/symlink is performed; a one-time `mv ~/.claude/context-mode "$CLAUDE_CONFIG_DIR/"` will restore continuity. A soft-fallback or migration helper can be added in a follow-up if maintainers prefer.

## Test plan

- [x] `npx vitest run tests/adapters tests/hooks/integration.test.ts` — all pass
- [x] New tests in `tests/adapters/claude-code-memory.test.ts` and `tests/adapters/claude-code.test.ts` cover both env-var-set and env-var-unset branches for `getConfigDir`, `getSettingsPath`, `getSessionDir`, `getMemoryDir`, plus the tilde-expansion and empty-string edge cases (added in QA round 1)
- [x] Pre-existing test failures in `tests/security.test.ts`, `tests/session-hooks-smoke.test.ts`, `tests/core/server.test.ts` confirmed to fail on `main` without these changes — unrelated to this PR

## QA rounds

Following [cmm-claude-code-setup CONTRIBUTING.md](https://github.com/halindrome/cmm-claude-code-setup/blob/main/CONTRIBUTING.md#pull-request-process), each QA round is posted as a separate PR comment with fixes in a separate commit. Round 1 complete (commit `a6a40ae`).

### DIFF
--- a/hooks/run-hook.mjs
+++ b/hooks/run-hook.mjs
@@ -13,7 +13,9 @@
  *     wrapper, where the handler is guaranteed to be live.
  *
  * Contract:
- *   - logs every failure to ~/.claude/context-mode/hook-errors.log
+ *   - logs every failure to <configDir>/context-mode/hook-errors.log,
+ *     where configDir honors $CLAUDE_CONFIG_DIR (incl. leading ~) and
+ *     falls back to ~/.claude — same contract as session-helpers.mjs (#453)
  *   - never propagates a non-zero exit (Claude Code surfaces non-zero as a
  *     "non-blocking hook error" on every tool call, which spams the user)
  *   - one-liner adoption for new hooks:
@@ -22,12 +24,23 @@
  */
 
 import { homedir } from "node:os";
-import { resolve } from "node:path";
+import { resolve, join } from "node:path";
 import { existsSync, mkdirSync, appendFileSync } from "node:fs";
+
+// Inlined to keep this wrapper dependency-free (parse-time imports must be
+// failure-proof). Mirrors session-helpers.mjs::resolveConfigDir for #453.
+function resolveClaudeConfigDir() {
+  const envVal = process.env.CLAUDE_CONFIG_DIR;
+  if (envVal) {
+    if (envVal.startsWith("~")) return join(homedir(), envVal.replace(/^~[/\\]?/, ""));
+    return envVal;
+  }
+  return resolve(homedir(), ".claude");
+}
 
 function logError(err) {
   try {
-    const dir = resolve(homedir(), ".claude", "context-mode");
+    const dir = resolve(resolveClaudeConfigDir(), "context-mode");
     if (!existsSync(dir)) mkdirSync(dir, { recursive: true });
     const line = `[${new Date().toISOString()}] pid=${process.pid} ${err?.stack || err?.message || String(err)}\n`;
     appendFileSync(resolve(dir, "hook-errors.log"), line);
--- a/src/adapters/claude-code/index.ts
+++ b/src/adapters/claude-code/index.ts
@@ -77,6 +77,13 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
    * `hooks/session-helpers.mjs::resolveConfigDir` already follows — including
    * tilde expansion for shells that pass `~/foo` through unchanged — so server
    * and hooks agree on where session-scoped state lives. See issue #453.
+   *
+   * Tilde regex `/^~[/\\]?/` only handles the current-user form (`~`, `~/`,
+   * `~\`); `~user/` is NOT expanded to a per-user homedir (matches
+   * `resolveConfigDir`). Non-tilde values are run through `resolve()` to
+   * normalize relative paths to absolute against cwd; the hook helper
+   * intentionally leaves them raw, but the adapter contract guarantees an
+   * absolute path (BaseAdapter.getConfigDir docstring).
    */
   getConfigDir(_projectDir?: string): string {
     const envVal = process.env.CLAUDE_CONFIG_DIR;
--- a/src/adapters/claude-code/index.ts
+++ b/src/adapters/claude-code/index.ts
@@ -74,13 +74,19 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
   /**
    * Honor `CLAUDE_CONFIG_DIR` (the canonical Claude Code config root) before
    * falling back to `~/.claude`. Mirrors the contract that
-   * `hooks/session-helpers.mjs` already follows so server and hooks agree on
-   * where session-scoped state lives. See issue #453.
+   * `hooks/session-helpers.mjs::resolveConfigDir` already follows — including
+   * tilde expansion for shells that pass `~/foo` through unchanged — so server
+   * and hooks agree on where session-scoped state lives. See issue #453.
    */
   getConfigDir(_projectDir?: string): string {
-    return process.env.CLAUDE_CONFIG_DIR
-      ? resolve(process.env.CLAUDE_CONFIG_DIR)
-      : resolve(homedir(), ".claude");
+    const envVal = process.env.CLAUDE_CONFIG_DIR;
+    if (envVal) {
+      if (envVal.startsWith("~")) {
+        return resolve(homedir(), envVal.replace(/^~[/\\]?/, ""));
+      }
+      return resolve(envVal);
+    }
+    return resolve(homedir(), ".claude");
   }
 
   getSessionDir(): string {
--- a/src/adapters/claude-code/index.ts
+++ b/src/adapters/claude-code/index.ts
@@ -6,9 +6,10 @@
  *
  * Claude Code hook specifics:
  *   - Session ID: transcript_path UUID > session_id > CLAUDE_SESSION_ID > ppid
- *   - Config: ~/.claude/settings.json
- *   - Session dir: ~/.claude/context-mode/sessions/
- *   - Plugin registry: ~/.claude/plugins/installed_plugins.json
+ *   - Config root: $CLAUDE_CONFIG_DIR (when set) or ~/.claude
+ *   - Settings: <configDir>/settings.json
+ *   - Session dir: <configDir>/context-mode/sessions/
+ *   - Plugin registry: <configDir>/plugins/installed_plugins.json
  */
 
 import {
@@ -18,6 +19,7 @@ import {
   readdirSync,
   chmodSync,
   accessSync,
+  mkdirSync,
   constants,
 } from "node:fs";
 import { resolve, join } from "node:path";
@@ -69,8 +71,26 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
 
   // ── Configuration ──────────────────────────────────────
 
+  /**
+   * Honor `CLAUDE_CONFIG_DIR` (the canonical Claude Code config root) before
+   * falling back to `~/.claude`. Mirrors the contract that
+   * `hooks/session-helpers.mjs` already follows so server and hooks agree on
+   * where session-scoped state lives. See issue #453.
+   */
+  getConfigDir(_projectDir?: string): string {
+    return process.env.CLAUDE_CONFIG_DIR
+      ? resolve(process.env.CLAUDE_CONFIG_DIR)
+      : resolve(homedir(), ".claude");
+  }
+
+  getSessionDir(): string {
+    const dir = join(this.getConfigDir(), "context-mode", "sessions");
+    mkdirSync(dir, { recursive: true });
+    return dir;
+  }
+
   getSettingsPath(): string {
-    return resolve(homedir(), ".claude", "settings.json");
+    return join(this.getConfigDir(), "settings.json");
   }
 
   generateHookConfig(pluginRoot: string): HookRegistration {
@@ -165,7 +185,7 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
       results.push({
         check: "PreToolUse hook",
         status: "fail",
-        message: "Could not read ~/.claude/settings.json",
+        message: `Could not read ${this.getSettingsPath()}`,
         fix: "context-mode upgrade",
       });
       return results;
@@ -290,9 +310,8 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
   getInstalledVersion(): string {
     // Primary: read from installed_plugins.json
     try {
-      const ipPath = resolve(
-        homedir(),
-        ".claude",
+      const ipPath = join(
+        this.getConfigDir(),
         "plugins",
         "installed_plugins.json",
       );
@@ -310,10 +329,13 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
     }
 
     // Fallback: scan common plugin cache locations
-    const bases = [
-      resolve(homedir(), ".claude"),
-      resolve(homedir(), ".config", "claude"),
-    ];
+    const bases = Array.from(
+      new Set([
+        this.getConfigDir(),
+        resolve(homedir(), ".claude"),
+        resolve(homedir(), ".config", "claude"),
+      ]),
+    );
     for (const base of bases) {
       const cacheDir = resolve(
         base,
@@ -510,9 +532,8 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
 
   updatePluginRegistry(pluginRoot: string, version: string): void {
     try {
-      const ipPath = resolve(
-        homedir(),
-        ".claude",
+      const ipPath = join(
+        this.getConfigDir(),
         "plugins",
         "installed_plugins.json",
       );
--- a/src/server.ts
+++ b/src/server.ts
@@ -126,25 +126,47 @@ let _detectedAdapter: HookAdapter | null = null;
 let _insightChild: ChildProcess | null = null;
 
 /**
+ * Resolve the Claude Code config root, honoring `CLAUDE_CONFIG_DIR` (incl.
+ * leading `~`) before falling back to `~/.claude`. Mirrors
+ * `hooks/session-helpers.mjs::resolveConfigDir` and
+ * `ClaudeCodeAdapter.getConfigDir` so the pre-detection path agrees with
+ * hooks/adapter on where Claude Code session data lives. See issue #453.
+ */
+function resolveClaudeConfigRoot(): string {
+  const envVal = process.env.CLAUDE_CONFIG_DIR;
+  if (envVal) {
+    if (envVal.startsWith("~")) return join(homedir(), envVal.replace(/^~[/\\]?/, ""));
+    return envVal;
+  }
+  return join(homedir(), ".claude");
+}
+
+/**
  * Get the platform-specific sessions directory from the detected adapter.
- * Falls back to ~/.claude/context-mode/sessions/ before adapter detection.
+ * Falls back to <CLAUDE_CONFIG_DIR>/context-mode/sessions/ (or
+ * ~/.claude/context-mode/sessions/) before adapter detection.
  */
 function getSessionDir(): string {
   if (_detectedAdapter) return _detectedAdapter.getSessionDir();
   // Pre-detection path (race window before MCP `initialize` completes):
   // call detectPlatform() (sync, env-var-based) and look up segments via
   // getSessionDirSegments() (sync map, no adapter instantiation). This keeps
-  // non-Claude platforms from spilling sessions into ~/.claude/.
+  // non-Claude platforms from spilling sessions into ~/.claude/. For Claude
+  // Code (segments=[".claude"]), reroute through the CLAUDE_CONFIG_DIR
+  // contract so the pre-detection window does not split-state with hooks.
   try {
     const signal = detectPlatform();
     const segments = getSessionDirSegments(signal.platform);
     if (segments) {
-      const dir = join(homedir(), ...segments, "context-mode", "sessions");
+      const root = segments.length === 1 && segments[0] === ".claude"
+        ? resolveClaudeConfigRoot()
+        : join(homedir(), ...segments);
+      const dir = join(root, "context-mode", "sessions");
       mkdirSync(dir, { recursive: true });
       return dir;
     }
-  } catch { /* fall through to .claude fallback */ }
-  const dir = join(homedir(), ".claude", "context-mode", "sessions");
+  } catch { /* fall through to claude fallback */ }
+  const dir = join(resolveClaudeConfigRoot(), "context-mode", "sessions");
   mkdirSync(dir, { recursive: true });
   return dir;
 }
@@ -1659,7 +1681,7 @@ server.registerTool(
         } catch { /* SessionDB unavailable — search ContentStore + auto-memory only */ }
       }
 
-      const configDir = _detectedAdapter?.getConfigDir() ?? (process.env.CLAUDE_CONFIG_DIR || join(homedir(), ".claude"));
+      const configDir = _detectedAdapter?.getConfigDir() ?? resolveClaudeConfigRoot();
 
       try {
       for (const q of queryList) {
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "134.9k+",
+  "message": "138.8k+",
   "color": "brightgreen",
-  "npm": "108.9k+",
+  "npm": "112.7k+",
   "marketplace": "26k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "127.7k+",
+  "message": "132.7k+",
   "color": "brightgreen",
-  "npm": "103.9k+",
+  "npm": "108.9k+",
   "marketplace": "23.7k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "132.7k+",
+  "message": "134.9k+",
   "color": "brightgreen",
   "npm": "108.9k+",
-  "marketplace": "23.7k+"
+  "marketplace": "26k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "125.4k+",
+  "message": "127.7k+",
   "color": "brightgreen",
   "npm": "103.9k+",
-  "marketplace": "21.4k+"
+  "marketplace": "23.7k+"
 }
--- a/tests/adapters/claude-code-memory.test.ts
+++ b/tests/adapters/claude-code-memory.test.ts
@@ -31,6 +31,16 @@ describe("ClaudeCodeAdapter memory conventions", () => {
     expect(adapter.getConfigDir()).toBe("/tmp/custom-claude-dir");
   });
 
+  it("getConfigDir expands leading ~ in CLAUDE_CONFIG_DIR (matches resolveConfigDir contract)", () => {
+    process.env.CLAUDE_CONFIG_DIR = "~/my-claude-cfg";
+    expect(adapter.getConfigDir()).toBe(join(homedir(), "my-claude-cfg"));
+  });
+
+  it("getConfigDir falls back to ~/.claude when CLAUDE_CONFIG_DIR is empty string", () => {
+    process.env.CLAUDE_CONFIG_DIR = "";
+    expect(adapter.getConfigDir()).toBe(join(homedir(), ".claude"));
+  });
+
   it("getInstructionFiles returns ['CLAUDE.md']", () => {
     expect(adapter.getInstructionFiles()).toEqual(["CLAUDE.md"]);
   });
--- a/tests/adapters/claude-code-memory.test.ts
+++ b/tests/adapters/claude-code-memory.test.ts
@@ -1,5 +1,5 @@
 import "../setup-home";
-import { describe, it, expect } from "vitest";
+import { afterEach, beforeEach, describe, it, expect } from "vitest";
 import { homedir } from "node:os";
 import { join } from "node:path";
 import { ClaudeCodeAdapter } from "../../src/adapters/claude-code/index.js";
@@ -11,16 +11,36 @@ import { ClaudeCodeAdapter } from "../../src/adapters/claude-code/index.js";
  */
 describe("ClaudeCodeAdapter memory conventions", () => {
   const adapter = new ClaudeCodeAdapter();
+  const savedConfigDir = process.env.CLAUDE_CONFIG_DIR;
 
-  it("getConfigDir returns ~/.claude", () => {
+  beforeEach(() => {
+    delete process.env.CLAUDE_CONFIG_DIR;
+  });
+
+  afterEach(() => {
+    if (savedConfigDir === undefined) delete process.env.CLAUDE_CONFIG_DIR;
+    else process.env.CLAUDE_CONFIG_DIR = savedConfigDir;
+  });
+
+  it("getConfigDir returns ~/.claude when CLAUDE_CONFIG_DIR is unset", () => {
     expect(adapter.getConfigDir()).toBe(join(homedir(), ".claude"));
+  });
+
+  it("getConfigDir honors CLAUDE_CONFIG_DIR when set (issue #453)", () => {
+    process.env.CLAUDE_CONFIG_DIR = "/tmp/custom-claude-dir";
+    expect(adapter.getConfigDir()).toBe("/tmp/custom-claude-dir");
   });
 
   it("getInstructionFiles returns ['CLAUDE.md']", () => {
     expect(adapter.getInstructionFiles()).toEqual(["CLAUDE.md"]);
   });
 
-  it("getMemoryDir returns ~/.claude/memory", () => {
+  it("getMemoryDir returns ~/.claude/memory by default", () => {
     expect(adapter.getMemoryDir()).toBe(join(homedir(), ".claude", "memory"));
+  });
+
+  it("getMemoryDir derives from CLAUDE_CONFIG_DIR when set", () => {
+    process.env.CLAUDE_CONFIG_DIR = "/tmp/custom-claude-dir";
+    expect(adapter.getMemoryDir()).toBe("/tmp/custom-claude-dir/memory");
   });
 });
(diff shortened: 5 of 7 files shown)


halindrome:
## QA Report — PR #460 (CLAUDE_CONFIG_DIR adapter fix) — Round 1

**Model used:** Claude Opus 4.7

### What Was Tested
- Commit narrative and diff (`8657218`) on branch `fix/claude-config-dir-honor`
- Modified file: `src/adapters/claude-code/index.ts` (new `getConfigDir`, `getSessionDir`, all `getConfigDir()` consumers)
- `src/adapters/base.ts` — to confirm `getSessionDir` override preserves `mkdirSync({recursive:true})`
- `src/server.ts:1662` (configDir resolution) and `src/server.ts:130-160` (pre-detection fallback)
- `hooks/session-helpers.mjs` — `resolveConfigDir` contract for tilde expansion
- New test coverage in `tests/adapters/claude-code.test.ts` and `tests/adapters/claude-code-memory.test.ts`
- Repo-wide grep for sibling hardcoded `~/.claude` sites the PR did not touch (`src/cli.ts`, `src/server.ts`, `src/security.ts`, `src/adapters/cursor/index.ts`, `src/adapters/detect.ts`)
- Other adapter override patterns (`opencode`, `openclaw`, `cursor`) for consistency
- Test setup harness (`tests/setup-home.ts`)

### Findings

#### F1 — Hook contract divergence: tilde expansion not handled in adapter — **major / confirmed**
`hooks/session-helpers.mjs::resolveConfigDir` explicitly expands a leading `~` (e.g. `CLAUDE_CONFIG_DIR=~/.claude-work` becomes `<homedir>/.claude-work`). The new `ClaudeCodeAdapter.getConfigDir` does NOT — it calls `resolve(process.env.CLAUDE_CONFIG_DIR)`, which on POSIX treats `~` as a literal directory name and produces e.g. `<cwd>/~/.claude-work`. The PR description claims the override "mirrors the contract" of `session-helpers.mjs`, but it does not.

Concrete impact: a user setting `CLAUDE_CONFIG_DIR=~/.claude-work` (a documented form, asserted in `tests/hooks/integration.test.ts:652`) will see the same split-state bug the PR claims to fix — hooks write to `$HOME/.claude-work/...`, server writes to `$CWD/~/.claude-work/...`. This is the exact regression mode the PR is supposed to close.

Tests do not catch this because they only set absolute paths (`/tmp/custom-claude-dir`, `join(fakeHome, ".config", "claude-code")`).

#### F2 — Empty-string `CLAUDE_CONFIG_DIR` falls back silently, but disagrees with hooks — **minor / confirmed**
`process.env.CLAUDE_CONFIG_DIR ? ... : ...` treats `""` as falsy and falls back to `~/.claude`. `resolveConfigDir` in the hook helper does the same (`if (envVal)`). Behaviorally consistent — fine. But `tests/hooks/integration.test.ts:620` already exercises the empty-string case for hooks; no equivalent test exists for the adapter. Nit-level test gap; behavior is OK.

#### F3 — Other call sites still hardcode `~/.claude` — **major / confirmed (out-of-scope but undermines the fix)**
The PR commit message says "all config-rooted paths (settings.json, sessions/, **plugins/**) flow through `$CLAUDE_CONFIG_DIR`." That is not true repo-wide:
- `src/server.ts:147` — fallback `getSessionDir` still hardcodes `~/.claude`
- `src/server.ts:345, 348` — `healCacheMidSession` reads `installed_plugins.json` and the cache root from `~/.claude/plugins/...` directly
- `src/server.ts:1662` — falls back to `process.env.CLAUDE_CONFIG_DIR || join(homedir(), ".claude")` (no tilde expansion either; same F1 bug)
- `src/cli.ts:673, 874, 1012, 1018` — marketplace dir, registry path, statusline candidate all hardcoded
- `src/security.ts:276, 342` — global settings path hardcoded
- `src/adapters/claude-code/index.ts:335-336` — fallback for `getInstalledVersion` still scans hardcoded `~/.claude` and `~/.config/claude` (these are deliberate as a fallback set, but the union doesn't include the user's `CLAUDE_CONFIG_DIR` — wait, it does, because `this.getConfigDir()` is the first entry; that one is fine)
- `src/adapters/cursor/index.ts:527` — references `~/.claude/settings.json` (cross-platform read; possibly intentional)

A user who set `CLAUDE_CONFIG_DIR` to keep state out of `~/.claude` will still see plugin self-heal, server pre-detection sessions, CLI install commands, and security policy reads scribble in `~/.claude`. The PR claim of full coverage is overbroad.

#### F4 — `getSessionDir` override correctly preserves `mkdirSync({recursive:true})` — **confirmed OK**
The new override at lines 86-89 mirrors `BaseAdapter.getSessionDir` exactly (`mkdirSync(dir, { recursive: true })`). No regression. `getSessionDBPath` and `getSessionEventsPath` inherit from base and call `this.getSessionDir()` — virtual dispatch routes through the override correctly. Verified via the base.ts source.

#### F5 — Migration: silent loss of existing `~/.claude/context-mode/sessions/` data — **major / confirmed**
For users who already had `CLAUDE_CONFIG_DIR` set in their shell but were running an older context-mode that wrote to `~/.claude` regardless: after upgrade, the server starts reading/writing at `$CLAUDE_CONFIG_DIR/context-mode/sessions/`. Existing FTS5 DBs, `-events.md` files, and stats JSON under `~/.claude/context-mode/sessions/` become orphaned. No migration code, no warning, no symlink heal. Sessions are not "lost" (still on disk) but are silently inaccessible and indexing rebuilds from scratch.

This is a behavior change worth a CHANGELOG note at minimum. Not a blocker if intentional, but undocumented.

#### F6 — `process.env.CLAUDE_CONFIG_DIR` read on every call — **minor / hypothetical**
`getConfigDir` evaluates `process.env.CLAUDE_CONFIG_DIR` on each call. Server-side `_detectedAdapter` is a process-lifetime singleton, but `process.env` mutation mid-process (uncommon but possible — e.g. test pollution, `EnterWorktree`, agent subprocesses) would flip every subsequent path. The new tests in `claude-code-memory.test.ts` and the `config paths` describe in `claude-code.test.ts` rely on this — they mutate the env var between cases. That is the desired test ergonomics, but it also means a stray `process.env.CLAUDE_CONFIG_DIR = ...` somewhere else in the test suite (in a parallel-running file under vitest's default `isolate: true` per-file, but same-file tests share process) could leak. Currently low risk because `setup-home.ts` doesn't touch `CLAUDE_CONFIG_DIR`.

Caching in a private field (computed once in the constructor or memoized) would be safer and cheaper, but the PR's chosen approach matches `session-helpers.mjs` style. Acceptable.

#### F7 — `_projectDir` parameter accepted but ignored — **nit / confirmed**
Both new methods accept `_projectDir` for interface symmetry but ignore it. Consistent with `BaseAdapter.getConfigDir` and the `opencode` override. Fine.

#### F8 — Other adapters unaffected — **confirmed OK**
`cursor`, `vscode-copilot`, `jetbrains-copilot`, `kiro`, `antigravity`, `openclaw`, `opencode`, `qwen-code` all define their own `getConfigDir`. None inherit from `ClaudeCodeAdapter`. `ClaudeCodeBaseAdapter` is the parent; its other subclasses (if any — only `ClaudeCodeAdapter` extends it per the file) are unaffected. Override scope clean.

#### F9 — Test isolation: `savedConfigDir` snapshot at module load — **nit / confirmed**
Both new tests capture `savedConfigDir = process.env.CLAUDE_CONFIG_DIR` at module load (top of `describe`). Vitest typically loads test files in fresh workers, so this is correct. But if `setup-home.ts` (loaded first) ever sets `CLAUDE_CONFIG_DIR`, the snapshot becomes the fake value. Currently `setup-home.ts` doesn't, so safe. Worth a comment.

### Overall Recommendation
**Needs fixes** — F1 (tilde expansion) is a real defect that recreates the bug the PR is claimed to fix, just for a documented input form. F3 (incomplete coverage) and F5 (migration) are scope/communication issues that should be acknowledged in the PR description and/or CHANGELOG even if deferred. F4 confirms the core mechanic is sound. Suggested minimum to land: fix F1 by reusing the same tilde-expansion logic as `resolveConfigDir`, add a test for `CLAUDE_CONFIG_DIR=~/foo`, and either narrow the PR description's scope claim or follow up with the remaining call sites in F3.

---

**Round 1 fix commit:** `a6a40ae` — addresses F1 (tilde expansion now matches `resolveConfigDir`) and F2 (empty-string test added). F3 (broader call-site coverage) and F5 (migration note) are tracked for follow-up; the PR description has been narrowed to clarify scope is the Claude Code adapter only.

halindrome:
## QA Report — PR #460 (CLAUDE_CONFIG_DIR adapter fix) — Round 2

**Model used:** Claude Opus 4.7

### What Was Tested
- Full diff `main..HEAD` (commits `8657218`, `a6a40ae`) over `src/adapters/claude-code/index.ts`, `tests/adapters/claude-code.test.ts`, `tests/adapters/claude-code-memory.test.ts`.
- Tilde-expansion regex `/^~[/\\]?/` against 10 edge cases (`""`, `undefined`, `/tmp/x`, `~/foo`, `~`, `~/`, `~user/foo`, `~~/foo`, `relative/path`, `./rel`).
- Cross-check between adapter `getConfigDir()` and hook `hooks/session-helpers.mjs::resolveConfigDir`.
- `validateHooks` diagnostic message path.
- `getInstalledVersion` legacy fallback dedup via `Set`.
- `getSessionDir` `mkdirSync({recursive:true})` semantics.
- Test isolation in `claude-code-memory.test.ts` and `claude-code.test.ts` (savedConfigDir capture vs `setup-home` ordering).
- Hardcoded `~/.claude` survey across `src/cli.ts`, `src/server.ts`, `src/security.ts`, `hooks/`.
- `getMemoryDir` (BaseAdapter) inheritance through `getConfigDir`.
- Vitest run for the three target test files (132 tests passed).
- `tsc --noEmit` (clean).

### Findings

#### F1 — Adapter and hook diverge on relative-path inputs (minor, confirmed)
Adapter's `getConfigDir()` uses `resolve(envVal)`; hook's `resolveConfigDir()` returns `envVal` raw for non-tilde inputs. For `CLAUDE_CONFIG_DIR=relative/path`, adapter returns `<cwd>/relative/path`, hook returns `relative/path`. Same divergence the PR claims to close, just for an undocumented input shape. Probably not user-hit in practice (Claude Code documents absolute paths), but the PR description claims "matches resolveConfigDir's contract" — the contracts are not byte-equal. Either: (a) document that adapter intentionally normalizes to absolute, or (b) drop `resolve()` and trust the env var as-is. Not a blocker since absolute env values produce identical results.

#### F2 — `~user/foo` and `~~/foo` produce surprising paths but are consistent with the hook (nit, confirmed)
Regex `/^~[/\\]?/` only strips `~`, `~/`, or `~\`. So `~user/foo` becomes `<HOME>/user/foo` (silently wrong vs POSIX shell expansion which is per-user homedir lookup). `~~/foo` becomes `<HOME>/~/foo`. Both behaviors match `hooks/session-helpers.mjs::resolveConfigDir` exactly (same regex), so the contract is preserved. Acceptable for parity, but worth a code comment that `~user/` is NOT supported.

#### F3 — `getSettingsPath()` in `validateHooks` error message is robust to `readSettings` failure modes (confirmed OK)
`readSettings()` (`src/adapters/claude-code/index.ts`) catches everything and returns `null`. `getSettingsPath()` is pure path arithmetic (no I/O), so the diagnostic message path is safe regardless of whether the underlying failure was ENOENT or EACCES. No regression.

#### F4 — Legacy `~/.claude` fallback in `getInstalledVersion` Set-dedup works (confirmed OK)
`Array.from(new Set([this.getConfigDir(), resolve(homedir(), ".claude"), resolve(homedir(), ".config", "claude")]))` — when `CLAUDE_CONFIG_DIR` is unset, `getConfigDir()` returns `resolve(homedir(), ".claude")` which string-equals the legacy entry, so the Set collapses to 2. When `CLAUDE_CONFIG_DIR=/custom`, all three remain. Behavior is correct.

#### F5 — `mkdirSync({recursive:true})` race: idempotent and safe (confirmed OK)
`recursive:true` swallows `EEXIST` for the leaf and intermediate dirs. Two processes calling `getSessionDir()` simultaneously cannot produce a partially-readable dir from this call alone (fs creates the dir atomically per path component). No issue.

#### F6 — Wider `~/.claude` hardcodes still present in `src/cli.ts`, `src/server.ts`, `src/security.ts`, `hooks/run-hook.mjs` (major, confirmed — but explicitly deferred)
Confirmed by grep:
- `src/server.ts:147` — `join(homedir(), ".claude", "context-mode", "sessions")` fallback.
- `src/server.ts:2796` — `join(homedir(), ".context-mode", ...)` (different path, OK).
- `hooks/run-hook.mjs:30` — `resolve(homedir(), ".claude", "context-mode")` for hook-errors.log.
- `src/security.ts:253` — comment only; the resolution code in security.ts uses cwd + ADAPTER paths.

Of these, `hooks/run-hook.mjs:30` is the most concerning leftover: hook-errors.log will land in `~/.claude/context-mode/hook-errors.log` even when `CLAUDE_CONFIG_DIR` points elsewhere. This re-creates the very split the PR fixes for sessions, just for the error log. The PR description acknowledges follow-ups, but `run-hook.mjs` is not called out by name and appears trivial to fix in this PR. Reviewer's call.

#### F7 — Test isolation in `claude-code-memory.test.ts` is correct (confirmed OK)
`const savedConfigDir = process.env.CLAUDE_CONFIG_DIR;` runs at describe-block module evaluation time, after `setup-home` has executed (which does NOT touch `CLAUDE_CONFIG_DIR`). `beforeEach`/`afterEach` cycle deletes/restores. `setup-home`'s `vi.mock("node:os")` makes `homedir()` return `fakeHome`, so the assertions `join(homedir(), ".claude")` resolve through the mock — internally consistent. No leak.

#### F8 — Test isolation in `claude-code.test.ts > config paths` describe block: marginal concern (nit, hypothetical)
The `savedConfigDir` constant is captured at module-load time, but the outer `describe("ClaudeCodeAdapter")` has its own `beforeEach`/`afterEach` that does `savedEnv = { ...process.env }` / `process.env = savedEnv` in the parsePreToolUseInput sub-block. Since both isolate independently (different `describe` scope), no interference. Verified by tests passing.

#### F9 — `openclaw`/`cursor` adapters do not import or depend on `~/.claude` paths (confirmed OK)
Grep over `src/adapters/openclaw` and `src/adapters/cursor` shows no references to `.claude` or `ClaudeCodeAdapter` internals. `WorkspaceRouter` (server.ts) handles non-Claude platforms via separate code path that is untouched by this PR.

#### F10 — Commit message accurately reflects scope (confirmed OK)
`a6a40ae` commit message accurately credits round-1 F1 (tilde) and F2 (empty string), references the underlying hook contract correctly. PR title (`fix(adapters): honor CLAUDE_CONFIG_DIR in Claude Code adapter`) is accurate; per the task description, the PR description was narrowed to acknowledge cli.ts/server.ts/security.ts as follow-ups. No overclaim.

#### F11 — Tests green, typecheck clean (confirmed OK)
`npx vitest run tests/adapters/claude-code.test.ts tests/adapters/claude-code-memory.test.ts tests/hooks/integration.test.ts` → 3 files, 132 tests passed, 5.22s.
`npx tsc --noEmit` → no output (clean).

### Overall Recommendation
**PASS with minor notes.** The targeted fix is correct, the tilde-expansion regex matches the hook contract on every input shape that matters, tests are green, and typecheck is clean. F1 (relative-path divergence) and F2 (`~user/`) are nits worth a follow-up code comment but do not block merge. F6 is the only real concern — `hooks/run-hook.mjs:30` retains a hardcoded `~/.claude` for `hook-errors.log` and is in the same conceptual class as the bug being fixed; consider folding it into this PR or filing it explicitly as a tracked follow-up issue.

---

**Round 2 fix commit:** `0f94395` — F6 addressed by routing `hooks/run-hook.mjs::logError` through an inlined `resolveClaudeConfigDir()` helper that mirrors `session-helpers.mjs::resolveConfigDir` (tilde expansion included). Test harness updated to strip `CLAUDE_CONFIG_DIR` from the spawned env so legacy-path assertions stay meaningful, and a new test asserts the env-var path is honored. F1/F2 nits documented inline with a clarifying comment on `getConfigDir()`.

halindrome:
## QA Report — PR #460 (CLAUDE_CONFIG_DIR adapter fix) — Round 3

**Model used:** Claude Opus 4.7

### What Was Tested
- Read full diff (3 commits, 5 files, +154/-24)
- Verified `npx tsc --noEmit` clean
- Verified `npx vitest run tests/adapters tests/hooks` — 855 passed, 1 skipped
- Compared `run-hook.mjs::resolveClaudeConfigDir`, `session-helpers.mjs::resolveConfigDir`, and `claude-code/index.ts::getConfigDir` for contract drift
- Audited remaining hardcoded `~/.claude` references in `src/` for cross-process consistency
- Reviewed `tests/hooks/run-hook.test.ts` for env pollution and `tmpHome` cleanup correctness
- Walked the `getSessionDir → getSessionDBPath → SessionDB` startup path

### Findings

#### F1. Server-side session DB still hardcodes `~/.claude/context-mode/sessions` — adapter fix incomplete (major, confirmed)
`src/server.ts:147` — `const dir = join(homedir(), ".claude", "context-mode", "sessions");` — does NOT consult the adapter or `CLAUDE_CONFIG_DIR`. So when `CLAUDE_CONFIG_DIR=/foo` is set, the hooks (mjs) write events under `/foo/context-mode/sessions/` while the MCP server reads/writes its session DB under `~/.claude/context-mode/sessions/`. This is exactly the cross-process split the PR claims to close (#453). Other lurking sites: `src/session/analytics.ts:530`, `src/cli.ts:673,874,1012,1018`, `src/security.ts:276,342`, `src/adapters/cursor/index.ts:527`. The PR title says "Claude Code adapter" — fair scope — but the PR description and adapter docstring both promise "server and hooks agree on where session-scoped state lives," which is not true while server.ts:147 stands.

#### F2. Three near-duplicate `CLAUDE_CONFIG_DIR` resolvers with subtly different contracts (major, confirmed)
Three implementations now live in the tree:
1. `hooks/session-helpers.mjs::resolveConfigDir` — returns raw `envVal` (no `resolve()` normalization).
2. `hooks/run-hook.mjs::resolveClaudeConfigDir` — returns raw `envVal` (matches #1).
3. `src/adapters/claude-code/index.ts::getConfigDir` — wraps non-tilde `envVal` in `resolve(envVal)` (intentional per docstring).

Drift surface: a relative `CLAUDE_CONFIG_DIR=foo` resolves differently in the server (cwd-anchored absolute) than in either mjs path (raw `"foo"`). If hook subprocess cwd ≠ server cwd they end up in different directories. The adapter docstring acknowledges this divergence but does not justify it; better to align all three on one helper. Recommend extracting into `hooks/session-helpers.mjs` (already exported) and importing where TS-side allows, OR documenting the relative-path contract as "undefined behavior — set absolute paths only."

#### F3. Empty-string `CLAUDE_CONFIG_DIR` handled inconsistently with whitespace-only (minor, confirmed)
All three resolvers use `if (envVal)` so `""` correctly falls through to `~/.claude`. But `CLAUDE_CONFIG_DIR=" "` (single space) passes the truthy check and is used verbatim — `mkdirSync(" /context-mode")` creates a literal space-prefixed directory in cwd. Low impact (no real user does this) but session-helpers and run-hook silently diverge from adapter (which `resolve()`s it to absolute). Consider trimming.

#### F4. `tmpHome` cleanup for the new test correctly handles nested `customCfg` (nit, confirmed)
`customCfg = join(tmpHome, "custom-claude-cfg")` lives inside `tmpHome`, and `rmSync(tmpHome, { recursive: true, force: true })` in `afterEach` cleans both. No leak. ✓

#### F5. Parent-env `CLAUDE_CONFIG_DIR` strip in `runScript()` is correct and not order-dependent (nit, confirmed)
`delete parentEnv.CLAUDE_CONFIG_DIR` operates on a copy (`{ ...process.env }`) — never mutates the vitest worker's `process.env`. Vitest test isolation is preserved. The new test's explicit `CLAUDE_CONFIG_DIR: customCfg` correctly overrides the strip via spread order. ✓

#### F6. `resolveClaudeConfigDir()` cannot throw under uncaughtException (nit, confirmed)
`process.env` access cannot throw, `homedir()` is sync and only fails if HOME is unset on POSIX (then throws — but `runHook` wraps `logError` in `try/catch {}` so logging still no-ops safely). `mkdirSync(..., {recursive:true})` on `ENAMETOOLONG` does throw, but the outer `try/catch` swallows it; log is silently lost rather than crashing the hook — matches the "never fail logging" contract. Acceptable. ✓

#### F7. `getSessionDir()` in adapter calls `mkdirSync` on every call — silently creates empty dir if migrating (minor, confirmed)
`src/adapters/claude-code/index.ts::getSessionDir()` unconditionally `mkdirSync({recursive:true})`s the new path. If a user previously had data under `~/.claude/context-mode/sessions/` and now sets `CLAUDE_CONFIG_DIR=/new`, the adapter happily creates `/new/context-mode/sessions/` empty — old session DB is orphaned, no migration, no warning. The PR makes no migration claim but should at least log "no existing session DB at <new>; legacy <old> exists" once at startup. Combined with F1 this can produce a *split-brain* where new hook events go to /new and old DB stays at ~. Recommend a one-time migration shim or at minimum a startup advisory log.

#### F8. Adapter `legacyConfigSearchPaths` still pins `~/.claude` (minor, confirmed)
`src/adapters/claude-code/index.ts:348` lists `resolve(homedir(), ".claude")` as a legacy search path even when `CLAUDE_CONFIG_DIR` is set, which is correct *as a fallback* but the array does NOT include the new `getConfigDir()` value if it's set — it only checks the hardcoded legacy locations. Out-of-scope for this PR (the function name says "legacy") but worth flagging.

### Overall Recommendation
**Needs fixes** — F1 is a confirmed cross-process inconsistency that contradicts the PR's stated goal. The adapter change in isolation is correct, but `src/server.ts:147` writes session DB outside the adapter, defeating the fix in any session that uses the MCP server alongside hooks. Either expand the PR to route `server.ts:147` through the adapter, or narrow the docstring/PR description to "hooks-only fix; server.ts session DB still at legacy path — tracked in #XXX."

F2 (helper duplication) is the right cleanup-now moment — three copies will drift. F7 (silent empty-dir migration) deserves at least a startup log line.

---

**Round 3 fix commit:** `42e4b3e` — F1 closed by routing the pre-detection branch and final fallback in `src/server.ts::getSessionDir()` through a new `resolveClaudeConfigRoot()` helper that mirrors the existing tilde-expansion contract. The earlier configDir resolution at line 1684 also now uses the helper. F2 (helper duplication across `.ts`/`.mjs`) tracked as a follow-up — consolidating across that boundary requires shared-util plumbing that's larger than this fix. F3 (whitespace-only env), F7 (no migration warning), and F8 (legacy search-paths) tracked for follow-up; behavior is documented in the PR description's "Out of scope" section.

Note: cursor-hooks flake (`stop.mjs > parses BOM-prefixed stdin` / `sessionstart.mjs > returns native additional_context JSON` intermittently fails 1/3 runs in the broader `tests/adapters tests/hooks` invocation) is **pre-existing on main** — confirmed by stashing this branch's changes and reproducing the same flake. Not introduced by this PR.

halindrome:
## QA Report — PR #460 (CLAUDE_CONFIG_DIR adapter fix) — Round 4

**Model used:** Claude Opus 4.7

### What Was Tested
- Re-read full `git diff main..HEAD -- src/` (HEAD=42e4b3e, 4 commits since main).
- Inspected new `resolveClaudeConfigRoot()` helper at `src/server.ts:125-145` and rewritten `getSessionDir()` (lines ~150-175) and `configDir` site at line 1684.
- Cross-checked `ClaudeCodeAdapter.getConfigDir()` in `src/adapters/claude-code/index.ts` and `hooks/session-helpers.mjs::resolveConfigDir` (line 156).
- Verified `getSessionDirSegments()` map in `src/adapters/detect.ts:89-` (every supported platform listed).
- Reviewed test diff: `tests/adapters/claude-code-memory.test.ts`, `tests/adapters/claude-code.test.ts`, plus existing `tests/core/server.test.ts:2942-2998` (segments map + getSessionDir source-shape assertions).
- `npx tsc --noEmit` -> clean.
- `npx vitest run tests/adapters` -> 489/489 pass. `npx vitest run tests/hooks/cursor-hooks.test.ts` (isolated) -> 18/18 pass. Combined run reproduced the known cursor-hooks BOM flake (1 fail) which is documented pre-existing.

### Findings

#### 1. Logic duplication across three call sites — Severity: Low (non-blocking)
`resolveClaudeConfigRoot()` in `src/server.ts`, `ClaudeCodeAdapter.getConfigDir()` in `src/adapters/claude-code/index.ts`, and `resolveConfigDir()` in `hooks/session-helpers.mjs` each independently implement the same CLAUDE_CONFIG_DIR + tilde-expansion + fallback contract. The server-side helper is intentional (avoids importing the adapter pre-detection to dodge the race window described in the comment) and the hook helper cannot import TS, so this is a justified triplicate. The JSDoc explicitly cites the mirror relationship. Acceptable, but a future consolidation (e.g., a shared `.mjs` resolver imported by both server and hooks) would be cleaner.

One subtle inconsistency: the adapter wraps the env value in `resolve(envVal)` (forces absolute against cwd, per `BaseAdapter.getConfigDir` contract), while `resolveClaudeConfigRoot()` returns `envVal` raw. In practice this only matters if a user sets `CLAUDE_CONFIG_DIR=relative/path`. Pre-detection then yields a relative session dir; post-detection (adapter) yields an absolute one — possible split-state. Real-world relative `CLAUDE_CONFIG_DIR` is exotic, so non-blocking, but worth a follow-up note.

#### 2. `[".claude"]` dispatch — Severity: None (correct)
Confirmed in `src/adapters/detect.ts:89-105`: only `claude-code` returns `[".claude"]`. No other platform shares that exact single-segment shape (`cursor`→`[".cursor"]`, `codex`→`[".codex"]`, `kilo`→`[".config","kilo"]`, etc.). The dispatch `segments.length === 1 && segments[0] === ".claude"` is uniquely correct for Claude Code.

#### 3. Information loss at line 1684 — Severity: None
Old: `_detectedAdapter?.getConfigDir() ?? (process.env.CLAUDE_CONFIG_DIR || join(homedir(), ".claude"))`. New: `... ?? resolveClaudeConfigRoot()`. The new helper additionally honors `~`-prefixed values (the old expression returned `~/foo` literally — a bug). Empty-string env value: old `||` falls through to `~/.claude`; new helper checks truthy `if (envVal)` so empty also falls through. **Behavior is strictly improved or equivalent on every platform.**

#### 4. Cursor / non-Claude platforms — Severity: None (correct)
Cursor segments are `[".cursor"]`, which fail the dispatch and route through the `join(homedir(), ...segments)` branch. Confirmed: no non-Claude adapter is rerouted through `CLAUDE_CONFIG_DIR`. The `tests/core/server.test.ts:2947` describe block locks down all platform→segments mappings.

#### 5. Test coverage of new server.ts behavior — Severity: Low (gap, not blocker)
`tests/core/server.test.ts:2973-2998` asserts `getSessionDir`'s source contains `detectPlatform` and `getSessionDirSegments` and that `.claude` appears only as fallback. **However, no test asserts that `resolveClaudeConfigRoot()` is invoked when segments=`[".claude"]`** — i.e., the round-3 fix itself has no behavioral test. The adapter tests in `tests/adapters/claude-code-memory.test.ts` cover `getConfigDir`/`getMemoryDir` with CLAUDE_CONFIG_DIR set/unset/empty/tilde, which validates the post-detection path well. A source-grep test (analogous to the existing one) ensuring the body of `getSessionDir` references `resolveClaudeConfigRoot` for the `.claude` branch would lock in this round's fix. Recommend adding in a follow-up.

#### 6. Blocking bugs — None found
tsc clean. Adapter and hook test suites pass in isolation. The combined-run flake is the documented pre-existing cursor-hooks BOM stdin race, unrelated to this PR.

### Overall Recommendation
**PASS with minor notes.** Ship it. Two non-blocking follow-ups for a future PR: (a) add a source-shape test that asserts `getSessionDir`'s `.claude` branch invokes `resolveClaudeConfigRoot()`; (b) consider deduplicating the three CLAUDE_CONFIG_DIR resolvers and aligning the relative-path normalization (adapter calls `resolve()`, server helper does not).

---

**Round 4 outcome:** Clean. No fix commit. Per [CONTRIBUTING.md QA cycle](https://github.com/halindrome/cmm-claude-code-setup/blob/main/CONTRIBUTING.md#pull-request-process): "Continue until a round comes back clean or only has hypothetical/minor findings." This round has only minor non-blocking findings — QA cycle complete.

PR #462, opened 2026-05-07
fix: skip unusable python launcher on Windows

### LINKED ISSUE #454: Windows: commandExists('python3') matches Microsoft Store launcher stub → executor spawns broken python3 (exit 9009)
## Summary

On default Windows 10 / 11 installs, `where python3` resolves to the **Microsoft Store App Execution Alias** at `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python3.exe`. This is a 0-byte launcher stub that, when Python is not installed via the Store, exits non-zero and prints a "go install Python from the Store" message — even when a real Python interpreter is already on `PATH` somewhere else.

`src/runtime.ts:60–68` defines `commandExists` purely in terms of `where <cmd>` exit code:

```ts
function commandExists(cmd: string): boolean {
  try {
    const check = isWindows ? `where ${cmd}` : `command -v ${cmd}`;
    execSync(check, { stdio: "pipe" });
    return true;
  } catch {
    return false;
  }
}
```

So `commandExists("python3")` returns `true` because the Store stub is "found", and `detectRuntimes()` (lines 174–178) selects `"python3"` as the executor binary:

```ts
python: commandExists("python3")
  ? "python3"
  : commandExists("python")
    ? "python"
    : null,
```

At runtime `executor.#spawn(["python3", scriptPath], ...)` resolves to the Store stub, the stub exits 9009, and **every Python sandbox call fails on the affected machine** even though a fully working `python` (or `py`) is one fallback away.

## Reproduction

On a Windows machine with the App Execution Alias enabled (default) and a separate Python install:

```powershell
where.exe python3
# C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python3.exe   ← stub
# E:\anaconda3\python.exe                                            ← real
# (etc.)

python3 --version
# (no output, exit code != 0, opens Store page or prints "Python was not found")

python --version
# Python 3.12.7
```

Then in this repo:

```powershell
npm test
# tests/executor.test.ts — Python: hello world
# AssertionError: 9009 !== 0
# (12 Python tests fail, 1 server.test.ts cap test fails)
```

`ctx_execute({ language: "python", code: "print('hi')" })` from any MCP client also fails with exit 9009.

## Suggested fix

`commandExists` for `python3` / `python` on Windows must additionally verify the binary actually runs. Two-line patch in `src/runtime.ts`:

```ts
function commandExists(cmd: string): boolean {
  try {
    const check = isWindows ? `where ${cmd}` : `command -v ${cmd}`;
    execSync(check, { stdio: "pipe" });
    return true;
  } catch {
    return false;
  }
}

// Stricter probe for runtimes that have notorious stubs (Windows MS Store).
function runnableExists(cmd: string): boolean {
  if (!commandExists(cmd)) return false;
  try {
    execFileSync(cmd, ["--version"], {
      shell: process.platform === "win32",
      stdio: "pipe",
      timeout: 5000,
    });
    return true;
  } catch {
    return false;
  }
}
```

Then in `detectRuntimes()`:

```ts
python: runnableExists("python3")
  ? "python3"
  : runnableExists("python")
    ? "python"
    : runnableExists("py")
      ? "py"
      : null,
```

Adding `py` (the Windows Python launcher) as a third fallback also covers the very common case where a user has Python installed but only `py.exe` is registered on PATH.

Optional belt-and-suspenders: explicitly skip paths under `…\Microsoft\WindowsApps\` when scanning `where` results, since those are *always* stubs.

## Why this matters

Windows is one of the three target OSes (#45). Today, on a fresh Windows install where the user hasn't disabled the App Execution Alias (Settings → Apps → App execution aliases) and hasn't installed Python from the Store, the executor is silently broken for any Python sandbox call. This affects benchmarks, ecosystem tests, and any user who runs `ctx_execute({ language: "python", ... })`.

## Environment

- Windows 10/11
- `where python3` → `C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\python3.exe` (App Execution Alias)
- Real interpreter: `Python 3.12.7` at `E:\anaconda3\python.exe`
- Node `v24.11.1`
- context-mode `1.0.111` (current `next`)

## Related

- #45 (Windows beta testing umbrella)
- Test fallout: 12 of 12 `executor.test.ts` Python cases + 1 `core/server.test.ts` cap-trigger case fail with exit 9009 on this configuration.

nightt5879:
Fixes #454.

## Summary

- verify Python candidates by running `--version` instead of trusting `where` / `command -v` alone
- fall back from a broken `python3` command to `python`, then to the Windows `py` launcher
- add Windows-focused regression coverage for the Microsoft Store App Execution Alias case

## Root cause

On Windows, `where python3` can resolve to `%LOCALAPPDATA%\Microsoft\WindowsApps\python3.exe`, the Microsoft Store App Execution Alias. That makes `commandExists("python3")` return true even though invoking `python3 --version` exits with 9009. Runtime detection then selects `python3`, so every Python sandbox execution fails despite a working `python` being available later on PATH.

## Validation

Validated on a Windows host that reproduces the launcher-stub setup:

- `where python3` resolves to `%LOCALAPPDATA%\Microsoft\WindowsApps\python3.exe`
- `python3 --version` exits with 9009
- `where python` resolves first to a real local Python install before the WindowsApps alias
- `python --version` reports `Python 3.13.5`
- `detectRuntimes().python` now returns `python`

Commands run:

- `cmd /c corepack pnpm exec vitest run tests/runtime.test.ts`
- `cmd /c corepack pnpm exec vitest run tests/executor.test.ts -t "Python"`
- `cmd /c corepack pnpm run build`
- `git diff --check`


mksglu:
Hi @nightt5879 Thank you so much! But it's duplicated atm I guess.

nightt5879:
> Hi @nightt5879 Thank you so much! But it's duplicated atm I guess.

Thanks for the clarification, I just noticed that someone was already working on this in #457. I missed that context earlier......

PR #470, opened 2026-05-07
fix(routing): skip Bash nudge for structurally-bounded commands (#463)

### LINKED ISSUE #463: Feature: user-configurable Bash command allowlist to skip flood-warning nudge
## Problem

The PreToolUse Bash hook fires its `<context_guidance>` nudge ("May produce large output. Use ctx_batch_execute / ctx_execute…") on every Bash invocation, including commands whose output is structurally small and where flooding context is impossible. Examples from real session traffic:

- `pwd`, `whoami`, `hostname`, `date`
- `echo …`, `printf …`
- `which …`, `type …`, `command -v …`
- `readlink …`, `basename …`, `dirname …`
- `cd …`, `mkdir -p …`, `touch …`
- `mv`, `cp`, `rm` (small, output-bounded by failures only)
- `ls`, `ls -la …` (bounded by tty width)
- `git status`, `git rev-parse …`, `git remote -v`, `git branch`, `git config --get …`, `git log -<small N>`, `git diff --stat`, `git diff --name-only`, `git stash list`, `git tag`
- `*--version`, `*-V` style probes

Each nudge is ~85 tokens. Across a ~50-Bash-call session, that's a few hundred wasted tokens — small individually, but the bigger cost is the friction: the nudge is pure noise on commands where the user/agent KNOWS no flood is possible, and trains the agent to ignore the nudge over time, weakening it for the cases where it actually matters (`find /`, `cat large-file`, unbounded `grep`).

## Proposed feature

A user-configurable allowlist that suppresses the Bash nudge for commands matching specified patterns. Two-tier:

1. **A built-in default allowlist** for commands whose stdout is structurally bounded — `pwd`, `whoami`, `echo`, `which`, `readlink`, `basename`, `dirname`, `mkdir`, `cd`, `mv`, `cp`, `rm`, `ls`, `git status`, `git rev-parse`, `git remote`, `git branch`, `git config --get`, `git log -<N>`, `git diff --stat`, `git diff --name-only`, `git stash list`, `git tag`, `*--version`. Conservatively skip anything piped/chained/redirected/subshelled.

2. **User extension** via config (e.g. `context-mode.allowlist` array in plugin config or a sibling `bash-allowlist.json`) so users can add their own safe commands without forking.

The allowlist would short-circuit the nudge but still let context-mode see the call (so PostToolUse session capture and stats accounting are unaffected).

## Why a flag rather than smarter detection

Output-size detection requires running the command, which defeats the purpose. Pattern-based allowlisting is the only way to suppress the nudge BEFORE the call without losing the safety net for unbounded commands.

## Current workaround

I'm running a local wrapper at `~/.claude/scripts/cm-bash-allowlist.sh` that gates the call before delegating to `pretooluse.mjs`, with a SessionStart hook that re-applies the patch to `${CLAUDE_PLUGIN_ROOT}/hooks/hooks.json` after each context-mode upgrade (since hooks.json gets reset on upgrade). It works, but it's brittle:

- The patch lives outside the plugin and silently breaks if the upgrade restructures hooks.json
- Other users can't benefit without copying the wrapper
- Future context-mode changes that depend on its own hook running first could be silently bypassed by the wrapper

A first-class allowlist in the plugin would solve all three.

## Happy to contribute

If a contributor PR is welcome, I can prepare one against `core/routing.mjs` adding the allowlist short-circuit before the size-detection logic, gated by a config flag default-on with the conservative defaults above. Let me know if there's a CONTRIBUTING.md or design preference I should follow.

— filed from a real-user workflow on `context-mode@1.0.111` with `claude-code@2.1.132`.

ousamabenyounes:
## What

Fixes #463.

The PreToolUse Bash routing nudge ("May produce large output. Use ctx_…") fires on every Bash invocation that isn't otherwise redirected. Useful for `find /` or `cat huge-log` — pure noise for `pwd`, `whoami`, `git status`, or `--version` probes. ~85 tokens per nudge, and it trains the agent to ignore the warning where it actually matters.

This adds a built-in allowlist of structurally-bounded commands. When the command's first token matches a safe pattern AND no shell control operator is present (pipe, redirect, command substitution, `&&`, `||`, `;`), the nudge is short-circuited and routing returns null. Unknown commands still get the nudge — fail-safe default.

## Why

`isStructurallyBounded()` requires both:
1. Match against a conservative pattern list (per the issue's proposal).
2. No shell control operator anywhere — `pwd | xargs cat`, `git status && tail -F log`, `echo $(find /)` all bypass the short-circuit.

The shell-operator gate is the load-bearing safety net: any operator can compose a safe command with an unbounded sink, so the allowlist refuses to short-circuit them.

The user-extensible config tier (issue's second paragraph) is deferred to a follow-up to keep this PR scope tight to the structural-allowlist mechanism.

## Allowlist

- System probes: `pwd`, `whoami`, `hostname`, `date`, `echo`, `printf`, `which`, `type`, `command -v`, `readlink`, `basename`, `dirname`
- Silent fs ops: `cd`, `mkdir`, `touch`, `mv`, `cp`, `rm`
- `ls` (rejects `-R` / `--recursive`)
- Git read-only: `status`, `rev-parse`, `remote`, `branch`, `config --get`, `diff --stat`, `diff --name-only`, `stash list`, `tag`
- `git log -<N>` (digit-bounded only — raw `git log` still nudges)
- `--version` anywhere, `cmd -V`

## Coverage added (`tests/core/routing.test.ts`)

8 new test cases under `Bash structurally-bounded allowlist (#463)`:
- Each allowlist group (system probes / git read-only / version probes)
- `ls` without `-R` allowed; `ls -R` and `ls --recursive` still nudge
- Unbounded commands (`find /`, `cat`, `grep -r`, `ps aux`, `git log`, `git diff`) still nudge
- Composition guard — `pwd | cat`, `git status && cat huge.log`, `echo $(find /)` etc. all still nudge
- Direct unit checks of `isStructurallyBounded()`

Regression updates to existing tests that used now-allowlisted commands as proxies for "any non-routed Bash command":
- `tests/guidance-throttle.test.ts` — bash-throttle proxies switched from `ls` / `pwd` to `npm install` / `find /` so they keep exercising the throttle, not the new allowlist.
- `tests/hooks/core-routing.test.ts` — `git status` / `mkdir` cases flipped to the new "allowlisted → null" behavior; gradle-in-quotes test rewritten to use `find` so it still exercises the strip-quotes-then-no-match path.
- `tests/hooks/integration.test.ts` — `Bash + git status` / `Bash + mkdir` end-to-end tests replaced with `npm install` / `find` (unbounded) and a new positive test pinning that `git status` now emits empty stdout.
- `tests/hooks/tool-naming.test.ts` — openclaw bash-guidance proxies switched from `ls` to `npm install`.
- `tests/hooks/vscode-hooks.test.ts` — `run_in_terminal` probe switched from `git status` to `npm install`.

## Test plan

- [x] `npm run build`
- [x] `npm test` — 73 files, **2408 pass / 25 skipped / 0 fail**
- [x] `npx vitest run tests/core/routing.test.ts` — 16 pass (8 pre-existing + 8 new)
- [x] `see-real-bug` repro: `pwd`, `whoami`, `git status`, `ls -la`, `echo hello`, `node --version` all fired the nudge on `next` @ `1f70bee`; after the fix all six return null. Repro artifact saved locally.

## Notes

- Targets `next` per CONTRIBUTING.md.
- No bundles committed (regenerated bundles reverted before commit).
- No new dependencies, no new test files.

### DIFF
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -161,13 +161,17 @@ const SAFE_COMMAND_PATTERNS = [
   /^readlink(?:\s+.+)?$/,
   /^basename(?:\s+.+)?$/,
   /^dirname(?:\s+.+)?$/,
-  // Filesystem ops (silent on success, errors on stderr only)
+  // Filesystem ops (silent on success, errors on stderr only).
+  // For cp / mv / rm we explicitly refuse `-v` / `--verbose`: verbose
+  // mode prints one line per file and can flood on big trees
+  // (recursive copy of /etc, mass rename, etc.). The "silent on
+  // success" invariant only holds without -v.
   /^cd(?:\s+.+)?$/,
   /^mkdir(?:\s+.+)?$/,
   /^touch\s+.+$/,
-  /^mv\s+.+$/,
-  /^cp\s+.+$/,
-  /^rm\s+.+$/,
+  /^mv(?!\s+-[a-zA-Z]*v\b)(?!\s+--verbose\b)\s+.+$/,
+  /^cp(?!\s+-[a-zA-Z]*v\b)(?!\s+--verbose\b)\s+.+$/,
+  /^rm(?!\s+-[a-zA-Z]*v\b)(?!\s+--verbose\b)\s+.+$/,
   // ls — refuse recursive (-R / --recursive) to keep output bounded.
   /^ls(?!\s+-[a-zA-Z]*R)(?!\s+--recursive)(?:\s+.+)?$/,
   // git read-only / status subcommands
@@ -187,7 +191,14 @@ const SAFE_COMMAND_PATTERNS = [
   /^\S+\s+-V(?:\s|$)/,
 ];
 
-const SHELL_CONTROL_OPERATORS = /[|`]|\$\(|>>|>|<(?!<)|&&|\|\||;/;
+// Bash shell control operators that can compose a safe command with an
+// unbounded sink. Any match disqualifies the command from the allowlist.
+//
+// Note `&` (single — background + sequence): listed BEFORE `&&` in the
+// alternation so the regex engine doesn't accidentally short-match `&&`
+// when `&` is itself a separator (`date & cat huge.log`). Without this,
+// `^date(?:\s+.+)?$` would match the whole string and bypass the gate.
+const SHELL_CONTROL_OPERATORS = /[|`]|\$\(|>>|>|<(?!<)|&(?!&)|&&|\|\||;/;
 
 /**
  * @param {string} command Raw Bash command string from the hook payload.
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -130,6 +130,78 @@ function stripQuotedContent(cmd) {
     .replace(/"[^"]*"/g, '""');                   // double-quoted strings
 }
 
+/**
+ * Built-in allowlist of structurally-bounded Bash commands (#463).
+ *
+ * The PreToolUse Bash nudge ("May produce large output. Use ctx_…") is
+ * tuned for unbounded commands like `find /` or `cat large-file`. On
+ * commands whose stdout is structurally bounded (system probes, version
+ * checks, simple git read subcommands), the nudge is pure noise — a
+ * recurring ~85 tokens that trains the agent to ignore the warning.
+ *
+ * isStructurallyBounded() returns true ONLY when the command:
+ *   1. Has no shell control operators (pipe, redirect, command
+ *      substitution, &&, ||, ;) — any of those can compose with an
+ *      unbounded command and re-introduce flooding.
+ *   2. Matches one of the conservative patterns below.
+ *
+ * Unknown commands are treated as unbounded (false) — fail-safe default.
+ */
+const SAFE_COMMAND_PATTERNS = [
+  // System probes (no stdout, or one short line)
+  /^pwd$/,
+  /^whoami$/,
+  /^hostname(?:\s+-[a-zA-Z]+)?$/,
+  /^date(?:\s+.+)?$/,
+  /^echo\s/,
+  /^printf\s/,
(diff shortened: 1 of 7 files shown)


ousamabenyounes:
Two holes patched per review feedback (commit ee39e92):

**1. Single `&` missing from operator gate.** Confirmed: `date & cat /var/log/syslog` was allowlisted on pre-patch code — the regex covered `&&` but not the standalone background+sequence `&`. Added `&(?!&)` to the alternation.

**2. `-v` / `--verbose` on `cp` / `mv` / `rm` floods.** Same shape as the `ls -R` carve-out — `(?!\s+-[a-zA-Z]*v\b)(?!\s+--verbose\b)` so `cp -v -r /etc /tmp` and friends fall through to the nudge. Silent-success path remains allowlisted.

Tests:
- 3 new single-`&` fixtures appended to the existing composition-guard block (`date & cat …`, `whoami & find /`, `pwd & tail -F …`).
- New `cp / mv / rm with -v / --verbose` block — 9 fixtures.
- New `cp / mv / rm without -v` sanity block — 5 fixtures pinning the silent-success path stays allowlisted.

Confirmed TDD: both new groups were red on `a65cc25`, green on `ee39e92`. `npm test` full suite: 73 files, 2410 pass / 25 skipped, 0 assertion failures (the 2 worker-pool teardown timeouts on `kiro-hooks` + `insight-cors` are pre-existing flakes — both files pass in isolation).

Edge cases left as known limitations (documented in issue/PR thread):
- `cat huge.txt --version` matches the `--version` pattern. CLIs usually short-circuit on `--version`, but it's not universal — flagged as accepted.
- `-V` is not universally a version flag (e.g. `tar -V label`). Same — accepted.
- The regex runs on raw command, not a shell parse, so `echo "pipe | inside string"` is conservatively flagged as having a pipe. False negative on the allowlist (nudge fires when it could skip), not a security risk.

PR #472, opened 2026-05-07
fix(pi): bridge MCP tools into pi.registerTool() so the LLM can call them (#426)

### LINKED ISSUE #426: Pi extension registers 0 tools — routing block tells LLM to use ctx_execute/ctx_search but they don't exist
## Bug Summary

The Pi extension (`build/pi-extension.js`) injects a routing block that instructs the LLM to use `ctx_execute`, `ctx_search`, `ctx_batch_execute`, `ctx_fetch_and_index`, etc. — but **none of these tools are registered** via `pi.registerTool()`.

The LLM sees the routing directive but cannot find the tools, so it falls back to ignoring the block entirely. This makes context-mode a **pure cost with zero benefit** in Pi.

### What the Pi extension DOES register:
```
pi.on("session_start", ...)        ✅ session tracking
pi.on("tool_call", ...)            ✅ event capture
pi.on("tool_result", ...)          ✅ event capture
pi.on("before_agent_start", ...)   ✅ routing block injection
pi.registerCommand("ctx-stats")    ✅ command
pi.registerCommand("ctx-doctor")   ✅ command
```

### What the Pi extension DOES NOT register:
```
pi.registerTool("ctx_execute")         ❌ MISSING
pi.registerTool("ctx_execute_file")    ❌ MISSING
pi.registerTool("ctx_search")          ❌ MISSING
pi.registerTool("ctx_batch_execute")   ❌ MISSING
pi.registerTool("ctx_fetch_and_index") ❌ MISSING
pi.registerTool("ctx_index")           ❌ MISSING
pi.registerTool("ctx_doctor")          ❌ MISSING (only a command, not a tool)
pi.registerTool("ctx_stats")           ❌ MISSING (only a command, not a tool)
pi.registerTool("ctx_purge")           ❌ MISSING
pi.registerTool("ctx_upgrade")         ❌ MISSING
pi.registerTool("ctx_insight")         ❌ MISSING
```

### Impact measurements (18 sessions over 2 days):

| Metric | Value |
|--------|-------|
| System prompt cost per window | ~2,500 tokens (routing block + 12 skill descriptions + MCP tool stubs) |
| Tool calls via ctx_* tools | **0** (never called) |
| Tokens saved by context-mode | **0** |
| Knowledge base events recorded | 447 (never retrieved) |
| Net ROI | **Negative** |

### Root cause

The server.bundle.mjs implements 13 MCP tools via the MCP protocol (stdio JSON-RPC), which works for Claude Code and other MCP hosts. But Pi extensions use `pi.registerTool()` API instead of MCP protocol. The Pi extension file (`pi-extension.js`) never bridges these two worlds.

## Suggested fix

Add `pi.registerTool()` calls in `pi-extension.js` that wrap the server-side tool handlers. Example pattern from Pi docs:

```javascript
pi.registerTool({
  name: "ctx_execute",
  label: "ctx_execute",
  description: "Execute code and index output for later search",
  parameters: Type.Object({
    language: Type.String({ description: "Programming language" }),
    code: Type.String({ description: "Code to execute" }),
  }),
  async execute(toolCallId, params, signal, onUpdate, ctx) {
    // Bridge to the existing server-side handler
    const result = await serverHandler.executeCode(params);
    return {
      content: [{ type: "text", text: result }],
      details: {},
    };
  },
});
```

This needs to be done for all 13 tools defined in server.bundle.mjs.

## Environment

- context-mode: 1.0.111
- pi: 0.73.0
- OS: macOS (Apple Silicon)
- Node: v22.20.0

ousamabenyounes:
## What

Fixes #426.

Pi 0.73.x has no native MCP support. From [pi-mono README](https://github.com/badlogic/pi-mono/tree/main/packages/coding-agent):

> **No MCP.** Build CLI tools with READMEs (see Skills), or build an extension that adds MCP support.

Without a bridge in the context-mode Pi extension, the routing block tells the LLM to call `ctx_execute` / `ctx_search` / etc. but those tools never enter Pi's tool list and the LLM cannot reach them. The reporter measured 18 sessions over 2 days: ~2,500 tokens of system-prompt overhead per window, **0 actual ctx_* calls**, 447 events recorded but never retrieved. ROI on Pi was net negative.

## Fix

A new `src/pi-mcp-bridge.ts` module:
- Spawns `server.bundle.mjs` as a long-lived MCP child via stdio.
- Performs the standard MCP handshake (`initialize` → `notifications/initialized`).
- Calls `tools/list` once, then registers each returned tool through `pi.registerTool({ name, label, description, parameters, execute })`.
- `execute()` forwards into the child via `tools/call`. Errors are thrown — Pi's contract for "tool failed" — so the LLM sees the MCP server's diagnostic text.

`src/pi-extension.ts` adds the bootstrap call (fire-and-forget, so spawn / handshake latency does NOT block session start) and a `session_shutdown` cleanup that SIGTERMs the child.

## Why JSON Schema instead of TypeBox

MCP `tools/list` returns JSON Schema. Pi's parameter validator accepts JSON Schema directly — TypeBox just adds Symbol metadata to JSON Schema for type inference. Passing the schema through unchanged avoids a runtime translation pass and keeps the bridge a true thin layer over the MCP protocol. Same code path Claude Code, Gemini CLI and the other adapters use.

## No new runtime dependencies

Pure `node:child_process` + `node:path`. The `@earendil-works/pi-*` packages are NOT pulled in as build deps — `pi` is typed structurally (matching the existing `src/pi-extension.ts` style of `pi: any`) and the bridge only touches the documented `pi.registerTool()` shape.

## Defense-in-depth

A missing `server.bundle.mjs` or any spawn / handshake error is surfaced once on stderr, then the extension keeps running with the existing hooks + slash commands. The bridge cannot break a Pi session for users with broken installs.

## Coverage added (`tests/pi-extension.test.ts`)

Two new `describe` blocks (no new test files — added to the existing domain file per CONTRIBUTING):

1. **`MCPStdioClient`** — 5 unit tests pinning the wire protocol with fake stdio servers:
   - Id-matched responses over `\n`-delimited JSON
   - Concurrent in-flight requests with out-of-order delivery (id-map dispatch)
   - Child-exit cancellation rejects every pending request
   - Timeout instead of hanging on a silent server
   - Non-JSON noise tolerance (banner lines, garbage)

2. **`bootstrapMCPTools`** — 2 integration tests against the real MCP server (`start.mjs`):
   - Asserts the canonical `ctx_*` set is registered: `ctx_execute`, `ctx_execute_file`, `ctx_search`, `ctx_index`, `ctx_batch_execute`, `ctx_fetch_and_index`, `ctx_doctor`, `ctx_stats`, `ctx_purge`
   - Round-trips `ctx_index` through `tools/call` (server returns "Indexed N sections … from: pi-bridge-smoke" — pin proves args make it across the bridge)

## Test plan

- [x] `npm run build`
- [x] `npm test` — 73 files, **2406 pass / 25 skipped / 0 fail**
- [x] `npx vitest run tests/pi-extension.test.ts` — 44 pass (37 pre-existing + 7 new for the bridge)
- [x] **see-real-bug** repro — three independent confirmations on `next` @ `1f70bee`:
  1. Static: `pi.registerTool` count = 0 in installed `build/pi-extension.js` (only `pi.registerCommand("ctx-stats")` and `pi.registerCommand("ctx-doctor")`)
  2. Pi documentation: explicit "No MCP" stance — Pi 0.73.x ignores `~/.pi/agent/mcp.json` for context-mode purposes
  3. Issue reporter's 18-session measurements: 0 ctx_* calls, 447 events captured but never retrieved
- [-] Live Pi LLM probe: free-tier Gemini quota was exhausted on every available key during the fix session, so the LLM-side regression contract is enforced via the integration test (which spawns the same MCP server Pi would, against the same Pi-facing `registerTool` surface).

## Out of scope

- The `~/.pi/agent/mcp.json` step in the Pi install README. Once this lands the step is harmless but no longer load-bearing — cleanup deferred to a docs-only follow-up.
- In-process refactor of `server.ts` handlers. The subprocess bridge matches every other adapter; an in-process inline is its own scope.


ousamabenyounes:
### Self-review (mode: empirical) — PR #472

Head SHA: `1734c787` · Base: `next`

**Phase A** (bug reproduces without fix): PASS — reverted `src/pi-extension.ts` and removed `src/pi-mcp-bridge.ts`, the 7 new tests all fail. **Caveat**: the failure mode is import-not-found, not a behavior assertion. See C1 below.

**Phase B** (test passes with fix): PASS — 7/7 green on head.

**Quality lenses**:
- C1 Coverage strength: 1 HIGH
- C2 Copy-paste: 1 LOW
- C3 Dead code: 1 LOW
- C4 Clarity: PASS
- C5 Performance: skipped (spawn-once-per-session, no hot path)
- C6 Security: PASS (env passthrough to same-package child, trust boundary unchanged)

#### Findings

**[HIGH] No test asserts the wiring in `src/pi-extension.ts` actually runs at load time.** Phase A passes only because removing `src/pi-mcp-bridge.ts` produces an import error, not because the bug reproduces behaviorally. If a future refactor drops the `bootstrapMCPTools(pi, serverBundle).then(...)` line from `src/pi-extension.ts` while keeping the bridge module intact, every one of the 7 new bridge tests stays green and the bug silently re-enters. Of the 36 existing `registerPiExtension(api)` tests in `tests/pi-extension.test.ts`, none assert that any `ctx_*` tool is registered through `pi.registerTool()` after the extension's default export runs.

Fix: add a test that calls `await registerPiExtension(api)`, waits for the fire-and-forget bootstrap to settle (export `_mcpBridgeReady` as a Promise from `pi-extension.ts` for tests), then asserts `api.registerTool.mock.calls` includes at least one entry whose `name` matches `/^ctx_/`. The mock already tracks calls — `api.registerTool` is `vi.fn()` in `createMockPiApi()`.

**[LOW] Dead getter `running` on `MCPStdioClient`.** `src/pi-mcp-bridge.ts:189` exports `get running(): boolean` but no caller exists in `src/` or `tests/`. Either drop it or wire it into one of the unit tests as a sanity probe.

**[LOW] Path resolution duplicated across the two integration tests.** `tests/pi-extension.test.ts:887-892` and `:931-934` both compute `mcpEntry = path.resolve(here, "..", "start.mjs")` from scratch. Lift to a `const` at the top of the `bootstrapMCPTools — registers every ctx_* tool with Pi` describe block.

#### Verdict

REQUEST_CHANGES — only for the C1 wiring-test gap. C2/C3 are nits, ship-or-not at maintainer discretion.

ousamabenyounes:
Self-review findings addressed in `b2baec9`:

**[HIGH C1] Wiring test added.** Exported `_mcpBridgeReady: Promise<void>` from `src/pi-extension.ts`; new test `pi-extension.ts wiring (#426 regression guard)` calls `registerPiExtension(api)`, awaits `_mcpBridgeReady`, asserts `api.registerTool.mock.calls` includes at least `ctx_execute` / `ctx_search` / `ctx_index` / `ctx_batch_execute` / `ctx_fetch_and_index`. Verified red-on-revert: revert ONLY the `bootstrapMCPTools(pi, ...)` call (keep the bridge module intact) → `registeredNames: []`, test fails. Restore → green.

**[LOW C2] Path resolution lifted.** `mcpEntry` + `mcpEnv` consts at top of the `bootstrapMCPTools — ...` describe, used by both integration tests.

**[LOW C3] Dead `running` getter dropped.** 5-line removal on `MCPStdioClient`.

Suite: 73 files, 2407 pass / 25 skipped / 0 fail. Bridge + wiring tests: 8/8.

ousamabenyounes:
Pushed structure refactor on top of the bridge fix per Mert's feedback. Three atomic commits, each with its own message + test plan:

**3243caf** `refactor(openclaw): consolidate src/openclaw/* into src/adapters/openclaw/`
- `git mv src/openclaw/{mcp-tools,workspace-router}.ts → src/adapters/openclaw/`
- Updates the 3 imports in the openclaw plugin + 2 test refs
- Deletes the now-empty src/openclaw/ dir

**5bb1b0f** `refactor: flatten src/concurrency/runPool.ts → src/runPool.ts`
- One-file directory was structural noise; flattened
- 2 import-path updates (server.ts, server.test.ts)

**4911c07** `refactor: relocate plugin entry files into src/adapters/<platform>/`
- `src/pi-extension.ts → src/adapters/pi/extension.ts`
- `src/pi-mcp-bridge.ts → src/adapters/pi/mcp-bridge.ts`
- `src/openclaw-plugin.ts → src/adapters/openclaw/plugin.ts`
- `src/opencode-plugin.ts → src/adapters/opencode/plugin.ts`
- Internal import paths updated (depth +2)
- Runtime path resolution updated (`resolve(buildDir, "..")` → `resolve(buildDir, "..", "..", "..")`)
- `package.json`: `pi.extensions`, `openclaw.extensions`, `main`, `exports.*` paths follow the build-output relocation
- `.pi/extensions/context-mode/index.ts` + `.openclaw-plugin/index.ts` thin wrappers updated
- 4 test-side dynamic-import path updates + 1 source-grep path update in `tests/core/cli.test.ts`

**Test plan** (full suite re-run after each commit):
- [x] `npm test` — 73/73 files, 2407 pass / 25 skipped / 0 fail
- [x] No stale `src/pi-extension`, `src/openclaw-plugin`, `src/opencode-plugin`, `build/pi-extension.js`, `build/openclaw-plugin.js`, `build/opencode-plugin.js` refs anywhere in src/ tests/ package.json or the thin-wrapper dirs (grep clean).

No force-push — the 3 refactor commits sit cleanly on top of the bridge fix so each step is reviewable on its own.

ousamabenyounes:
## CI fix + independent review

### CI fix (3b0bfd5)
- OpenClaw E2E was failing because `scripts/test-openclaw-e2e.sh` and `scripts/install-openclaw-plugin.sh` still pointed at `build/openclaw-plugin.js` — the structural refactor moved the build output to `build/adapters/openclaw/plugin.js`.
- Updated both scripts (with a fall-back to the legacy path for transition safety).
- Bonus: the opencode + openclaw plugins were calling `routing.initSecurity(buildDir)` with the adapter sub-dir instead of the build root, so every plugin load emitted a noisy "security.js not found" WARNING. Fixed — now passes `resolve(buildDir, "..", "..")`.
- Local: `bash scripts/test-openclaw-e2e.sh` → 39/39 pass.
- Latest CI run: OpenClaw E2E success ✅ (1m10s on 3b0bfd5).

### Independent review (general-purpose subagent, no prior context)

Empirical Phase A on the wiring guard test commit (b2baec9) was the most interesting:
- Reverted ONLY the `bootstrapMCPTools(pi, serverBundle).then(...)` call line in `src/adapters/pi/extension.ts` while keeping `src/adapters/pi/mcp-bridge.ts` and the import statement intact.
- Result: **exactly 1 test fails** — the wiring guard, with `registeredNames: []` instead of `arrayContaining(["ctx_execute", ...])`. The other 7 bridge tests stay green. Confirms the wiring guard catches the precise regression class no other test catches.

Quality lenses:
- C1 Coverage strength: **PASS** — every prod path is exercised behaviorally; arrayContaining is additive-friendly while still pinning the canonical names.
- C2 / C3 / C4 / C6: **PASS**.
- C5 Performance: SKIPPED (spawn-once-per-session, no hot path).

Refactor sanity:
- Stale path strings, package.json paths, thin wrappers, runtime resolve depths, `initSecurity` arg: all PASS.
- One LOW finding — 3 stale paths in `docs/adapters/openclaw.md`. **Fixed in `fa60719`.**
- One INFO — `let _mcpBridgeReady` export binding semantics; existing docstring already covers it.

**Verdict: APPROVE** — bug reproduces empirically, fix asserts the regression behaviorally, refactor is import-clean across src/, tests/, scripts/, package.json, thin wrappers, and runtime resolve depths.

mystilleef:
After testing I found this issue. Pi subagents may miss `ctx_*` tools due to an async registration race.

Each subagent starts a fresh `pi --mode json -p --no-session` process. context-mode’s Pi adapter registers `ctx_*` tools through the MCP bridge, but that bootstrap runs fire-and-forget via `_mcpBridgeReady`. The child process can start the prompt before registration finishes, so the initial tool registry may not include `ctx_execute`, `ctx_batch_execute`, etc.

This does not look like a frontmatter allowlist issue; the agents already include the ctx tool names. context-mode likely needs to await bridge setup, or block first agent start until `_mcpBridgeReady` settles.

PR #474, opened 2026-05-08
fix: keep upstream credits outside skills directory

wo7864:
## Summary\n- move UPSTREAM-CREDITS.md from skills/ to the package root so skill scanners do not treat it as a skill\n- update vendored skill footer links to ../../UPSTREAM-CREDITS.md\n- include UPSTREAM-CREDITS.md in the npm package files allowlist\n\n## Validation\n- verified skills/ has no root-level .md files\n- verified vendored SKILL.md files point to the new credits location


mksglu:
Hi @wo7864 Let's resolve conflicts

PR #476, opened 2026-05-08
fix(security): close DNS rebinding gap in ctx_fetch_and_index SSRF guard

aaronjmars:
## Summary

`ctx_fetch_and_index` runs `ssrfGuard()` in the parent process to resolve the URL's hostname and validate that the returned IP isn't link-local / IMDS / multicast / reserved. The actual HTTP fetch then happens inside a subprocess spawned by `executor.execute(buildFetchCode(url, ...))`, which calls `await fetch(url)` — and that subprocess does its **own** DNS lookup. Between the parent's pre-flight resolution and the subprocess's connect-time resolution, an attacker controlling a low-TTL DNS record can rebind the hostname from a public IP (passes ssrfGuard) to a blocked IP like `169.254.169.254` (cloud metadata service). The existing comment in `ssrfGuard` claims to defend against DNS rebinding, but the implementation only validates the parent's lookup — the subprocess's lookup is unchecked.

## Impact

- LLM-driven prompt injection in any indexed page (or a malicious URL passed via tool output) can call `ctx_fetch_and_index("https://attacker-controlled.example/")`.
- The attacker's DNS server returns a public IP for the parent's `ssrfGuard` lookup, then `169.254.169.254` (or any other blocked range) for the subprocess's actual connect.
- On AWS / GCP / Azure / Cloud9 / Codespaces / EC2 instances — anywhere the metadata service is reachable from the host running the MCP — the subprocess fetches IMDS and the returned content (e.g. IAM role credentials on IMDSv1) flows back into the LLM's context as "fetched URL content."
- The same primitive also reaches arbitrary RFC1918 internal services when the user hasn't enabled `CTX_FETCH_STRICT=1`.

## Location

`src/server.ts` — `buildFetchCode()` (around line 1779) generates the subprocess code that calls `await fetch(url)` without any post-`ssrfGuard` validation. `ssrfGuard()` runs in the parent at `src/server.ts:1877`, then `fetchOneUrl()` spawns the subprocess at `src/server.ts:2049` — the two lookups are independent.

## Fix

Embed `classifyIp` (already exported from `src/server.ts`) into the subprocess and wrap `dns.lookup` so every connect-time resolution runs through the same policy:

- Parent's `ssrfGuard` is unchanged — still fails fast for `file://`, link-local hostnames, etc.
- Subprocess monkey-patches `dns.lookup` before any `fetch()` call. `undici`'s connector goes through `dns.lookup` to resolve the socket target; the patched lookup re-classifies every returned record and returns an error before TCP if any verdict is `"block"` (or `"private"` when `CTX_FETCH_STRICT=1`).
- `classifyIp.toString()` keeps the subprocess in sync with the parent automatically — no policy duplication.

The two-layer defense closes the rebinding window: even if the resolver returns different addresses across the parent and child, the connect-time check runs at the moment the socket is opened, so the IP that ends up in the TCP `connect()` is also the IP that was just classified.

## Detected by

Aeon + manual review of the `ssrfGuard` / `buildFetchCode` parent/child boundary.

- Severity: medium
- CWE-918 (Server-Side Request Forgery)
- CWE-441 (Unintended Proxy or Intermediary, "DNS rebinding" sub-class)

## Verification

- `npx tsc --noEmit` — clean.
- `npx vitest run` — 2256/2256 tests that pass on `main` still pass on this branch. The 7 pre-existing failures (5 sudo-deny hook tests + 2 IDEA_INITIAL_DIRECTORY env tests) reproduce identically on `main` before the patch — unrelated to this change.
- `tests/core/server.test.ts` `classifyIp` block (8 tests) all pass — the policy table is unchanged.

## Notes

- No new dependencies. The patch uses only `node:dns` (already a Node built-in).
- The wrapper handles the `dns.lookup` overload variants (`(host, cb)`, `(host, family, cb)`, `(host, options, cb)`, `{ all: true }`). Tested by reading Node's docs; the existing test suite's lookup paths still resolve normally for non-malicious hostnames.
- `CTX_FETCH_STRICT` propagates to the subprocess via `process.env` — preserved.

---

Filed by [Aeon](https://github.com/aaronjmars/aeon-aaron).


mksglu:
cc @ousamabenyounes

ousamabenyounes:
---

Nice catch on the parent/child DNS asymmetry. A few items before merge:

**1. `dns/promises` and `dns.resolve*` not patched.**
Patching `dns.lookup` doesn't affect `dnsPromises.lookup` — separate function references. Today undici uses callback-`lookup` so default `fetch` is covered, but the invariant is fragile.

```js
const dnsP = require('node:dns/promises');
const _origLookupP = dnsP.lookup;
dnsP.lookup = async function(hostname, options) {
  const opts = Object.assign({}, options || {}, { all: true, verbatim: true });
  const records = await _origLookupP(hostname, opts);
  for (const r of records) {
    const v = classifyIp(r.address);
    if (v === 'block' || (STRICT && v === 'private')) {
      throw new Error('SSRF blocked: ' + hostname + ' → ' + r.address);
    }
  }
  return options && options.all ? records : { address: records[0].address, family: records[0].family };
};
```

**2. Proxy env bypass.**
Subprocess inherits `HTTP_PROXY` / `HTTPS_PROXY` / `ALL_PROXY`. If a proxy is set, target DNS happens at the proxy → local guard never sees it.

```js
delete process.env.HTTP_PROXY;
delete process.env.HTTPS_PROXY;
delete process.env.ALL_PROXY;
delete process.env.http_proxy;
delete process.env.https_proxy;
delete process.env.all_proxy;
```

**3. `classifyIp.toString()` is brittle.**
Breaks silently if `classifyIp` ever closes over module scope. Pin it with a test:

```js
// tests/core/server.test.ts
it('classifyIp is self-contained (safe to .toString() into subprocess)', () => {
  const fn = new Function('return (' + classifyIp.toString() + ')')();
  expect(fn('169.254.169.254')).toBe('block');
  expect(fn('8.8.8.8')).toBe('allow');
});
```

**4. Dead branch.**
`all: true` is forced, so `records` is always an array.

```js
// remove this block:
if (!Array.isArray(records)) {
  records = [{ address: records, family: (options && options.family) || 4 }];
}
```

**5. No regression test for the actual fix.**
The PR says overloads were "tested by reading Node's docs". Minimum:

```js
it('blocks rebinding to IMDS at connect-time', async () => {
  vi.spyOn(dns, 'lookup').mockImplementation((h, o, cb) => cb(null, [{ address: '169.254.169.254', family: 4 }]));
  await expect(fetchOneUrl('https://attacker.example/')).rejects.toThrow(/SSRF blocked/);
});

it('blocks RFC1918 under CTX_FETCH_STRICT=1', async () => { /* 10.0.0.1 */ });
it('catches 30x redirect to IMDS', async () => { /* redirect → 169.254.169.254 */ });
```

mksglu:
Hi @aaronjmars Let's resolve CR comments

Issue #473, opened 2026-05-07
Feature request: env var(s) to configure storage directory(ies)


binhex:
Currently context-mode stores data in platform-specific directories under `~/.{platform}/context-mode/` (e.g. `~/.pi/`, `~/.claude/`), split across two locations:

- Session events → adapter-detected platform dir
- Content index (FTS5 BM25) → always `~/.claude/context-mode/content/`
- Session stats → `~/.claude/context-mode/sessions/`

For users running non-native harnesses (e.g. Oh My Pi/OMP which shares the Pi adapter path), or users who want all context-mode data in a single custom directory, there is currently no way to redirect storage.

**Request:** Add one or more environment variables to override storage paths.

**Preferred approach — single root env var:**
```
CONTEXT_MODE_DIR=~/.omp/context-mode
```
This would redirect all three (session events DB, content index, session stats) under that root. Falls back to the current platform-detected path when unset.

**Alternative — individual overrides:**
- `CONTEXT_MODE_SESSION_DIR` — session events database
- `CONTEXT_MODE_CONTENT_DIR` — BM25 FTS5 content index
- `CONTEXT_MODE_STATS_DIR` — process-level stats JSON files

This would allow OMP, Pi, Claude Code, and other harnesses to coexist without sharing or conflicting over storage directories.

Happy to submit a PR if this direction is agreeable.

mksglu:
@binhex @ousamabenyounes @mikij @sebastianbreguel 

Hey, thanks for opening this and for offering to submit a PR.

Before we land an ENV var I want to push back a bit on the framing. Every flag we ship without a clear, validated use case becomes dead code and tech debt later, and right now we don't have data on how many people actually hit this. I'd rather understand the root cause than patch it with a workaround, and I think this is worth doing properly.

If you're still up for the PR, here's what I'd like you to do before we agree on a direction:

1. Map the full picture across all 14 adapters. Why does each AI coding agent get a different path today? What does that split actually buy us, and what breaks if we collapse it? I want a clear answer to this before we add a knob on top.
2. Find the root cause, not a workaround. Is this strictly the non-native harness case (OMP sharing the Pi adapter path), or is it broader? If it's only OMP/Pi, an adapter-level fix is probably cleaner than a global ENV that every adapter has to respect. The Pi Agent repo is open source, clone it and use it as a reference for how non-native harnesses interact with our paths.
[...]

PR #480, opened 2026-05-08
feat(adapters): add OMP (Oh My Pi) platform adapter (#473)

ousamabenyounes:
## What

Adds a dedicated `OMPAdapter` (mcp-only paradigm) so context-mode users running under [Oh My Pi](https://github.com/can1357/oh-my-pi) get their session DB, stats, and content index rooted at `~/.omp/context-mode/` instead of leaking into `~/.claude/`.

Closes the adapter gap behind #473.

## Why

#473 reported: *"Context Index and Session Stats were located in `~/.claude/`"* after switching from Pi to OMP.

I audited the storage path resolution from `next` and confirmed the in-thread observation by @nightt5879:
- `getSessionDir()` (src/adapters/base.ts) already honors per-adapter `sessionDirSegments`.
- `getStorePath()` (src/server.ts:238) derives the content DB from the same session dir.
- `getStatsFilePath()` (src/server.ts:462) joins `getSessionDir()` with `stats-<sessionId>.json`.

So the storage layout is already adapter-rooted — there is no hardcoded `~/.claude/...` for content/stats. The remaining failure mode is **adapter selection**: there is no `OMPAdapter`, so detection either falls through to `pi` (and from there to the Claude fallback in `getAdapter()`) or directly to Claude. Either way the segments end up `[".claude"]`.

This PR closes that gap by adding OMP as a first-class adapter so `getAdapter("omp")` returns an instance whose `sessionDirSegments = [".omp"]`. **No new env-var override layer** — keeping the design constraint @mksglu raised in the issue: "every flag we ship without a clear, validated use case becomes dead code and tech debt."

### Scope notes

- **PiAdapter still does not exist.** `getAdapter("pi")` currently falls through the `default:` branch and returns a `ClaudeCodeAdapter`. That is a pre-existing bug, not part of #473's stated scope, and is not addressed here. Happy to file a follow-up issue if you want — wanted to keep this PR review-light.
- **No `CONTEXT_MODE_DIR` env var** added. Discussed in-thread, parking it until there's a validated use case.

## Coverage added

- `tests/adapters/detect.test.ts`
  - Detects `omp` from `OMP_PROCESSING_AGENT_DIR`.
  - Prefers `omp` over `pi` when both env vars are set (precedence regression).
  - `omp` accepted in the platform fallback assertion.
  - `getAdapter("omp")` returns `OMPAdapter` instance.
- `tests/adapters/omp.test.ts` (new — mirrors `tests/adapters/antigravity.test.ts`)
  - Identity (name, paradigm).
  - All `capabilities` are false (mcp-only).
  - All `parse*Input` methods throw with `OMP does not support hooks`.
  - All `format*Response` methods return `undefined`.
  - `generateHookConfig`, `configureAllHooks`, `setHookPermissions` return empty.
  - `getSessionDir()` resolves under `~/.omp/context-mode/sessions/`.
  - `getSessionDBPath` / `getSessionEventsPath` contain `.omp` and **never** `.claude` (the issue's exact failure mode).
  - Default `getSettingsPath()` → `~/.omp/agent/mcp_config.json`, default `getConfigDir()` → `~/.omp/agent`.
  - `OMP_PROCESSING_AGENT_DIR` overrides settings + config dir resolution.
  - `getInstructionFiles()` returns `["PI.md"]` (Pi compat).

## Test plan

- [x] `npm run build` — bundles regenerate (reverted before commit per repo convention; CI rebuilds)
- [x] `npm test` — full suite green: **2439 passed, 25 skipped, 0 failed**
- [x] `npx vitest run tests/adapters/{omp,detect,detect-config-dir,client-map}.test.ts` — 106 passed
- [x] No new test file fragmentation: per-adapter test file is the existing convention (`antigravity.test.ts`, `cursor.test.ts`, `kiro.test.ts`, etc.); `omp.test.ts` follows that pattern.
- [x] Bundles (`cli.bundle.mjs` / `server.bundle.mjs`) excluded from the commit per repo rules.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -40,6 +40,7 @@ src/
     opencode/      → OpenCode adapter
     codex/         → Codex CLI adapter
     vscode-copilot/ → VS Code Copilot adapter
+    omp/           → OMP (Oh My Pi) adapter — MCP-only, isolated ~/.omp/ storage (#473)
   openclaw/
     workspace-router.ts → Workspace path resolution for Pi Agent sessions
   openclaw-plugin.ts   → OpenClaw gateway plugin entry (sync register)
--- a/README.md
+++ b/README.md
@@ -807,6 +807,45 @@ Full configs: [`configs/kiro/mcp.json`](configs/kiro/mcp.json) | [`configs/kiro/
 </details>
 
 <details>
+<summary><strong>OMP (Oh My Pi)</strong> — MCP-only, isolated storage under <code>~/.omp/</code></summary>
+
+**Prerequisites:** Node.js 18+, [Oh My Pi](https://github.com/can1357/oh-my-pi) installed.
+
+**Install:**
+
+1. Install context-mode globally:
+
+   ```bash
+   npm install -g context-mode
+   ```
+
+2. Add to `~/.omp/agent/mcp_config.json` (or `$OMP_PROCESSING_AGENT_DIR/mcp_config.json` if the env var is set):
+
+   ```json
+   {
+     "mcpServers": {
+       "context-mode": {
+         "command": "context-mode"
+       }
+     }
+   }
+   ```
+
+3. Copy routing instructions (OMP has no hook support):
+
+   ```bash
+   cp node_modules/context-mode/configs/omp/PI.md ./PI.md
+   ```
+
+4. Restart OMP.
+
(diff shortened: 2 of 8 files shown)


mksglu:
Updated docs (readme, support adaptors) etc.

PR #486, opened 2026-05-08
chore: ignore temporary checkout dirs

mkc909:
## Summary
- Adds ignore coverage for temporary checkout directories created during local plugin work.

## Verification
- Git branch was pushed and local status is clean.


mikij:
I think you need to remove these bundle files from git cache in order to gitignore works over previously tracked files

PR #365, opened 2026-04-27
fix(mcp): resolve ctx_index relative paths from project dir

Tomodad:
## Summary
- Resolve `ctx_index(path)` relative file paths against context-mode's detected project directory instead of the MCP server process cwd.
- Preserve the original relative path as the source label when no explicit `source` is provided.
- Add a regression test that starts the MCP server with `CONTEXT_MODE_PROJECT_DIR` and indexes a relative file path via `ctx_index`.

## Test plan
- [x] `pnpm run typecheck`
- [ ] `pnpm exec vitest run tests/core/server.test.ts --testNamePattern \"ctx_index file paths\"` — blocked locally by missing `better-sqlite3` native binding under Node 24 (`Could not locate the bindings file`).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #492, opened 2026-05-09
[codex] add compact hook support

MarcusNeufeldt:
## Summary

Adds Codex-specific hook setup for context-mode, including PreCompact snapshot persistence and a SessionStart compact-resume bug fix.

## What changed

- Adds `hooks/codex/precompact.mjs` to persist resume snapshots, compact counts, and compaction summary events.
- Updates Codex hook generation, validation, and upgrade to configure all Codex hook entries plus `[features].hooks = true`.
- Honors `CODEX_HOME` consistently in Codex adapter paths and the MCP server pre-detection fallback.
- Fixes Codex compact resume restore: the previous `SessionStart` compact branch marked the resume consumed before appending the snapshot to `additionalContext`; this now injects the snapshot first, then marks it consumed.
- Expands Codex PreToolUse matcher coverage for shell, apply_patch/file edits, and context-mode MCP tool names, while avoiding read-only hook interception.
- Makes doctor script discovery tolerate both nested hook config entries and Cursor-style flat command entries.
- Preserves malformed Codex `hooks.json` by backing it up before replacement.
- Updates Codex install/platform docs and regenerated shipped bundles.

## Validation

- `npm run typecheck`
- `npm run build`
- `npx vitest run tests/adapters/codex.test.ts tests/core/cli.test.ts tests/hooks/core-routing.test.ts`
- `npx vitest run tests/core/server.test.ts -t "ctx_doctor hook script checks|ctx_doctor|doctor|getSessionDir honors CODEX_HOME"`
- `git diff --check`

## Notes

Codex `PreCompact` is treated as runtime-gated: context-mode generates the hook entry, but missing PreCompact validation is a warning because upstream Codex parity docs still describe the event as partial.


ousamabenyounes:
Thanks for the work — test coverage on the hook scripts is solid, and the `routing.mjs` consolidation into `matchesContextModeTool` is clean. A few things to address before merge, in priority order:

### 🔴 P1 — `getCommandsFromHookEntry`, `extractHookScriptPath`, `getHookScriptPaths` are duplicated across `src/cli.ts` and `src/server.ts`

Verified with `diff`: the three helpers are identical in both files. Suggest extracting to `src/util/hook-config.ts` and importing from both:

```ts
// src/util/hook-config.ts
import type { HookAdapter } from "../adapters/types.js";

export function getCommandsFromHookEntry(entry: unknown): string[] { /* … */ }
export function extractHookScriptPath(command: string): string | null { /* … */ }
export function getHookScriptPaths(adapter: HookAdapter, pluginRoot: string): string[] { /* … */ }
```

### 🔴 P2 — `getCodexWorktreeSuffix` / `normalizeCodexProjectDir` reimplement existing helpers, uncached

`src/session/db.ts` already exports `getWorktreeSuffix` with documented memoization (the comment notes `~12ms per fork on macOS, 50ms+ Windows` — that's why it's cached). The new `normalizeCodexProjectDir` is identical to the existing `normalizeWorktreePath` modulo parameter name, and `getCodexWorktreeSuffix` duplicates `getWorktreeSuffix` without the cache.

```ts
// src/session/db.ts — add `export` to normalizeWorktreePath
export function normalizeWorktreePath(path: string): string { /* … */ }

// src/adapters/codex/index.ts — reuse, drop both new helpers
import { getWorktreeSuffix, normalizeWorktreePath } from "../../session/db.js";

getSessionDBPath(projectDir: string): string {
  const normalized = normalizeWorktreePath(projectDir);
  const hash = createHash("sha256").update(normalized).digest("hex").slice(0, 16);
  return join(this.getSessionDir(), `${hash}${getWorktreeSuffix(normalized)}.db`);
}
```

`CONTEXT_MODE_SESSION_SUFFIX` env override is already handled by the existing function, so no behavior change.

### 🟡 P3 — `resolveCodexConfigRoot` (server.ts) and `resolveCodexConfigDir` (adapter) are the same function under two names. Pick one and reuse.

### 🟡 P4 — A few `~/.codex/hooks.json` strings remain hardcoded in `validateHooks` / `configureAllHooks` error messages and the `Failed to update ~/.codex/hooks.json` throw, even though the rest of the function uses `this.getHooksPath()`. Misleading for users with `CODEX_HOME` set.

### 🟢 P5 — `PRE_TOOL_USE_MATCHER_PATTERN` includes `Read`, which fires on every file read — intentional? And the alternation has redundancy: `mcp__.*__ctx_execute` already covers `mcp__plugin_context-mode_context-mode__ctx_execute`.

One small thing worth surfacing in the PR description: the `sessionstart.mjs` change isn't just "restores snapshots" — prior to this PR, the `compact` branch called `markResumeConsumed` immediately without ever appending the snapshot to `additionalContext`. It's a behavioral fix, not just plumbing.

Happy to re-review once these are addressed.

MarcusNeufeldt:
Thanks for the review. I pushed a cleanup commit addressing the before-merge items:

- extracted the shared hook-config command/script discovery into `src/util/hook-config.ts` and reused it from CLI + server doctor
- reused `normalizeWorktreePath()` / cached `getWorktreeSuffix()` from `src/session/db.ts` in the Codex adapter
- moved Codex config-dir resolution into `src/adapters/codex/paths.ts` and reused it from adapter + server pre-detection
- replaced remaining Codex hook fix/error strings with the resolved hooks path
- removed `Read` from the Codex PreToolUse matcher and removed the redundant plugin-qualified MCP matcher alternations
- updated the PR description to describe the `SessionStart` compact-resume path as a behavioral bug fix

Validation on the rebased branch:

- `npm run typecheck`
- `npm run build`
- `npx vitest run tests/adapters/codex.test.ts tests/core/cli.test.ts tests/hooks/core-routing.test.ts`
- `npx vitest run tests/core/server.test.ts -t "getSessionDir honors CODEX_HOME|ctx_doctor|doctor"`
- `git diff --check`

One note: I also ran the broader focused set including full `tests/core/server.test.ts`; it reached 468 passing tests and then hit the existing Windows temp-dir cleanup `EBUSY` in `ctx_index: projectRoot path resolution (#365)`, so I reran the changed server coverage narrowly and it passed.

MarcusNeufeldt:
Tiny follow-up pushed after another pass:

- `ctx_doctor` now resolves direct `.mjs` hook script paths against `pluginRoot`, matching CLI doctor behavior for relative hook paths.
- Added a static regression in `tests/core/server.test.ts` for that path resolution.
- Updated the PR description to explicitly describe the `SessionStart` compact-resume consume-before-append bug fix.

Additional validation:

- `npm run typecheck`
- `npm run build`
- `npx vitest run tests/core/server.test.ts -t "ctx_doctor hook script checks|ctx_doctor|doctor|getSessionDir honors CODEX_HOME"`
- `git diff --check`

PR #497, opened 2026-05-09
Fix/pi extension routing and session start

ToniBig:
What / Why / How

 Two bugs in the Pi coding agent adapter (src/pi-extension.ts) that caused context-mode MCP tools to be invisible
 to the LLM after turn 1, and session IDs to be non-deterministic.

 Bug 1 — Routing block disappears after turn 1
 The _routingInjected gate was ported from the Claude Code adapter where the SessionStart hook fires once and its
 output persists for the entire session. In Pi, before_agent_start is called on every turn and the system prompt is
 rebuilt from scratch each time. The gate meant the <context_window_protection> block was only present on turn 1 —
 from turn 2 onwards the LLM lost all awareness of context-mode MCP tools (ctx_search, ctx_batch_execute,
 ctx_execute, etc.). Fix: remove the _routingInjected set and gate entirely; inject the routing block on every
 before_agent_start call.

 Bug 2 — Wrong session_start handler signature
 The handler was registered as (ctx) but Pi's handler signature is (event, ctx). The first argument received was
 the event object ({ reason: "startup" }), not ExtensionContext. This caused ctx.sessionManager to always be
 undefined inside deriveSessionId, falling back to pi-${Date.now()} and breaking session continuity in the DB. Fix:
 correct the signature to (event, ctx). This change is Pi-only — other platforms use independent hook mechanisms
 and are unaffected.

 Affected platforms

 - [ ] Claude Code
 - [ ] Cursor
 - [ ] JetBrains Copilot
 - [ ] Gemini CLI
 - [ ] Qwen Code
 - [ ] OpenCode
 - [ ] KiloCode
 - [ ] Codex CLI
 - [ ] OpenClaw (Pi Agent)
 - [x] Pi
 - [ ] Kiro
 - [ ] Antigravity
 - [ ] Zed
 - [ ] All platforms

 Test plan

 Unit tests (tests/pi-extension.test.ts):
 - Updated Slice 7: renamed "does not re-inject the routing block on subsequent calls" → "re-injects the routing
 block on every subsequent call". Old test asserted the buggy behaviour (0 occurrences on turn 2). New test asserts
 <context_window_protection> is present on turns 1, 2, and 3.
 - Added Slice 4 test: "session_start receives (event, ctx)" — passes a real sessionManager as the second argument
 and confirms the extension initialises without falling back to a timestamp-based session ID.

 Integration test (Pi runtime, manual):

 ```
   # Turn 1
   pi -p --append-system-prompt "Output ROUTING_PRESENT if you see <context_window_protection> in your system
 prompt, ROUTING_ABSENT if not." "do you see context_window_protection?"
   # → ROUTING_PRESENT ✅

   # Turn 2 (continued session)
   pi -c -p --append-system-prompt "Output ROUTING_PRESENT if you see <context_window_protection> in your system
 prompt, ROUTING_ABSENT if not." "do you still see context_window_protection?"
   # → ROUTING_PRESENT ✅  (was ROUTING_ABSENT before this fix)
 ```

 Checklist

 - [x] npm test passes
 - [x] npm run typecheck passes
 - [x] Targets next branch (unless hotfix)

 <details>
 Our CI runs on Ubuntu, macOS, and Windows.

 - Use path.join() / path.resolve(), never hardcode / separators
 - Use event-based stdin reading — readFileSync(0) breaks on Windows
 - Use os.tmpdir(), never hardcode /tmp

 </details>


mksglu:
@ToniBig Thanks but the target branch should be `next`

mikij (approved):
LGTM

mksglu:
Let's resolve conflict @ToniBig

ToniBig:
> @ToniBig Thanks but the target branch should be `next`

Next time I'll know. Thanks!

Issue #511, opened 2026-05-10
Dynamic require of "node:fs" is not supported


SkipXS:
### Platform

Pi

### context-mode version

v1.0.118

### Debug script output (REQUIRED)

```json
context_mode_ctx_doctor
 context-mode doctor

 [OK] Runtimes: 5/11 (45%) — javascript, shell, typescript, python, perl
 [OK] Performance: FAST (Bun)
 [OK] Server test: PASS
 [OK] FTS5 / SQLite: PASS — native module works
 [OK] PreToolUse hook: PreToolUse hook configured
 [OK] SessionStart hook: SessionStart hook configured
 [OK] Hook script: PASS — /home/sebas/.nvm/versions/node/v26.1.0/lib/node_modules/context-mode/hooks/pretooluse.mjs
 [OK] Hook script: PASS — /home/sebas/.nvm/versions/node/v26.1.0/lib/node_modules/context-mode/hooks/posttooluse.mjs
 [OK] Hook script: PASS — /home/sebas/.nvm/versions/node/v26.1.0/lib/node_modules/context-mode/hooks/precompact.mjs
 [OK] Hook script: PASS — /home/sebas/.nvm/versions/node/v26.1.0/lib/node_modules/context-mode/hooks/userpromptsubmit.mjs
 [OK] Hook script: PASS — /home/sebas/.nvm/versions/node/v26.1.0/lib/node_modules/context-mode/hooks/sessionstart.mjs
 [OK] Version: v1.0.118


 Thinking...

 scripts/ctx-debug.sh does not exist anywhere on disk. No match under /home/sebas, context-mode, or pi-coding-agent.

 Ran ctx_doctor as the closest equivalent — all checks pass:

 
   [OK] Runtimes: 5/11 (45%) — javascript, shell, typescript, python, perl
   [OK] Performance: FAST (Bun)
   [OK] Server test: PASS
   [OK] FTS5 / SQLite: PASS
   [OK] PreToolUse hook: configured
   [OK] SessionStart hook: configured
   [OK] All 5 hook scripts: PASS
   [OK] Version: v1.0.118
 

 What were you trying to debug? If you have the script contents, paste them and I'll create + run it.
```

### Exact prompt that triggered the bug (REQUIRED)

```text
Please check if the app updates the feeds on startup
```

### Full error output (REQUIRED)

```text
Dynamic require of "node:fs" is not supported
```

### Steps to reproduce (REQUIRED)

[...]

PR #513, opened 2026-05-10
fix(core): make project-dir transcript fallback ESM-safe

Scotchy49:
## Summary

Fixes the ESM/bundled runtime path that can throw `Dynamic require of "node:fs" is not supported` before context-mode tools do useful work.

Changes:
- replace the transcript project-dir resolver's inline `require("node:fs")` / `require("node:path")` calls with ESM-safe static imports
- add regression coverage that imports the compiled `build/util/project-dir.js` module and exercises the transcript fallback path under Node/npm and Bun

## Root cause

Commit `2ac2ea9af263595a2ef2cb0ff3cb694394eaaadd` added `resolveProjectDirFromTranscript()` with inline CommonJS `require("node:fs")` and `require("node:path")` calls. The package is ESM and the published MCP entrypoint is bundled as `.mjs`, so those calls become dynamic-require helper calls in the bundle and can throw when `require` is unavailable.

This affects paths such as `ctx_search`, `ctx_index`, and `ctx_execute_file` when project-dir resolution falls into the transcript heuristic before `PWD`/`cwd` fallback.

Related issue: #511

## Validation

Passed:

```text
npm run build
npx vitest run tests/util/project-dir.test.ts
```

The focused test covers:

```text
compiled ESM resolver runs under Node without CommonJS require
compiled ESM resolver runs under Bun without CommonJS require
```

Full suite note:

```text
npm test
```

currently fails locally on `next` in unrelated statusline rendering expectations and one Codex stop-hook test:

```text
tests/statusline.test.ts: 7 failed
tests/statusline-sqlite.test.ts: 2 failed
tests/adapters/codex.test.ts: outputs valid JSON without requesting continuation
```


Scotchy49:
Implementation note after re-checking the original inline-import comments:

The previous code kept `require("node:fs")` / `require("node:path")` inside `resolveProjectDirFromTranscript()` so call sites that did not enter the transcript heuristic stayed easy to test or stub. That constraint is real, but the inline CommonJS `require()` is the part that fails once this ESM package is bundled and run through the MCP entrypoint.

This PR keeps the same `fs.*` / `path.*` call style, but makes the module ESM-safe with namespace imports:

```ts
import * as fs from "node:fs";
import * as path from "node:path";
```

I intentionally did not add dependency injection for `fs`/`path`: there are no current project-dir tests stubbing this module, and adding an injectable API would broaden the resolver for a hypothetical need. Namespace imports are the minimal compromise: ESM-safe for Node/npm and Bun, readable, and still compatible with normal module-mocking approaches if a future test needs it.

PR #515, opened 2026-05-10
test(statusline): align with v1.0.118 byte format + add kb() regression suite

murataslan1:
## Context

`efbd294` ("fix(statusline): drop $ math, mirror ctx_stats byte metrics") landed in v1.0.118 and rewrote `bin/statusline.mjs` around the new exported `kb()` formatter. The test assertions weren't updated alongside it — running `vitest` against the v1.0.118 tag on a clean checkout currently fails 9 statusline tests for the same root cause (still matching on `$[0-9]` and `saved this session`).

This PR catches the tests up and pins the new contract so the dual-currency drift can't reappear silently.

## What changed

**`tests/statusline.test.ts`** — 4 resolver/override blocks and 3 fallback paths now assert on the v1.0.118 copy (`this chat`, `lifetime`, `kept out`, `preserved across compact, restart & upgrade`). Fallback tests also scope `HOME` to the temp dir so `getMultiAdapterLifetimeStats()` can't pull lifetime data from the developer's real home and skip the brand-new headline branch.

**`tests/statusline-sqlite.test.ts`** — the lifetime-$ assertion swapped for a `\b\d+(\.\d+)?\s*(KB|MB|GB)\b` regex; the empty-dir fallback gets the same `HOME` scoping fix.

**`tests/kb-formatter.test.ts`** (new) — regression suite around the unified formatter:
- `kb()` boundary cases across B / KB / MB / GB, including the 100-unit decimal switch in each scale
- `kb` is exported from analytics and `bin/statusline.mjs` imports it
- statusline output never contains `$` in ACTIVE or FRESH state
- ACTIVE renders `{kb} this chat` and ends in `kept out`
- FRESH renders `{kb} kept out · preserved across compact` and omits the ACTIVE `this chat` block

## Test plan

- [x] `npx vitest run tests/kb-formatter.test.ts tests/statusline.test.ts tests/statusline-sqlite.test.ts tests/adapters/omp.test.ts tests/adapters/omp-plugin.test.ts` → 64/64
- [x] `npx vitest run` (full suite) → 2843 passed, 20 skipped. Two `tests/session/continuity.test.ts` cases were flaky under full-suite parallelism but pass cleanly when the file is run in isolation; they're unrelated to this PR.
- [x] Live `node bin/statusline.mjs <<< '{}'` against a populated home produces `1016 KB kept out · 29.9 KB/day · preserved across compact, restart & upgrade` — no `$`.

## Notes for the reviewer

- The `HOME=temp_dir` scoping in the fallback tests is purely a test-isolation fix — it doesn't claim production behavior should change. If you'd rather see `getMultiAdapterLifetimeStats()` take an explicit `home` arg from statusline.mjs (so the test seam matches the production seam), happy to follow up.
- I considered an end-to-end "statusline and `ctx_stats` produce the same kb-formatted byte total for the same DB" parity test, but it's hard to do cleanly without an in-process MCP runtime. Left it out for now; the `kb` import-source check in `kb-formatter.test.ts` is the lightweight stand-in.
- Verified against `omp v14.9.3` with context-mode linked via `omp plugin link` — statusline renders the new format end-to-end.

### DIFF
--- a/tests/adapters/omp-statusline-integration.test.ts
+++ b/tests/adapters/omp-statusline-integration.test.ts
@@ -11,6 +11,16 @@ import "../setup-home";
  * a realistic session lifecycle, then spawn bin/statusline.mjs against
  * the DB the plugin actually wrote to and assert the v1.0.118 contract
  * holds end-to-end (no '$', kb()-formatted numbers, expected copy).
+ *
+ * Slices:
+ *   1. ACTIVE render — basic session_start + tool_results → statusline shows
+ *      "{kb} this chat · {kb} lifetime · % kept out"
+ *   2. Routing — tool_call(curl) is blocked AND statusline still renders the
+ *      v1.0.118 format afterward (block side effects don't poison the DB)
+ *   3. Compaction — session_before_compact persists a resume snapshot AND
+ *      statusline continues to render the v1.0.118 format
+ *   4. Tolerance — malformed events don't take statusline down and the
+ *      output stays free of '$'
  */
 
 import { describe, it, expect, beforeEach, afterEach } from "vitest";
@@ -24,6 +34,11 @@ import { fakeHome } from "../setup-home";
 const STATUSLINE = resolve(process.cwd(), "bin", "statusline.mjs");
 
 type HandlerFn = (...args: unknown[]) => unknown | Promise<unknown>;
+
+interface BlockResult {
+  block?: boolean;
+  reason?: string;
+}
 
 function createMockOmpApi() {
   const handlers: Record<string, HandlerFn[]> = {};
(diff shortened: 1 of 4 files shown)


murataslan1:
Two follow-up commits expanded coverage beyond the initial statusline-only scope:

**`tests/adapters/omp-statusline-integration.test.ts`** — split into four slices:
1. **ACTIVE render** — DB lands at `OMPAdapter.getSessionDir()`; statusline shows `{kb} this chat · {kb} lifetime · % kept out` with no `$`; rendered bytes go up monotonically with more events (catches a routing miss between plugin tool_results and statusline reads); non-matching `CLAUDE_SESSION_ID` falls back to FRESH cleanly.
2. **`tool_call` routing** — bash `curl` is blocked with a context-mode reason; blocked call leaves no trace in the SessionDB statusline reads (would have caught a regression where the block fell through to the event extractor); non-bash `tool_call` returns undefined.
3. **`session_before_compact`** — resume snapshot persisted, `compact_count` incremented, AND statusline keeps rendering the v1.0.118 format across the compaction boundary.
4. **Malformed-event tolerance** — a `tool_result` missing `toolName` / `content` is swallowed and a follow-up legit event still flows through to statusline.

**`tests/kb-formatter.test.ts`** — added the two render paths that were thin on the first pass:
- BRAND_NEW state (`saves ~98% of context window` headline survives the v1.0.118 refactor, no `$`, no `this chat`, no `kept out`)
- ACTIVE percent label (v1.0.118 swapped `efficient` for `kept out` on the percent block)

**`tests/statusline-sqlite.test.ts`** — the multi-adapter `across N tools` test now also pins no `$`; the multi-USD figure was the other place the pre-v1.0.118 statusline reached for dollar math.

Updated count: **76/76** on the focused subset (52 originally, plus 24 new across these three files).

murataslan1:
Heads-up while testing this against a real DB: the v1.0.118 fix landed for `bin/statusline.mjs` but the narrative renderer in `src/session/analytics.ts` still emits `$`. Running `ctx_stats` against my local DB returns:

\`\`\`
Persistent memory  ✓ preserved across compact, restart & upgrade
  1.3K events · 334 sessions · ~$5.05 saved lifetime
...
─────────────────────────────────────────────────────────────────
Your AI talks less, remembers more, costs less.
$0.00 this session  ·  $0.00 lifetime
─────────────────────────────────────────────────────────────────
\`\`\`

The leftover dollar sites I spotted:

- `src/session/analytics.ts:2127` — `~${tokensToUsd(lifetimeTokens)} saved lifetime` in the persistent-memory block
- `src/session/analytics.ts:2205` — `${sessionUsd} this session  ·  ${lifetimeUsd} lifetime` (footer)
- `src/session/analytics.ts:2236` — `${lifetimeUsd} saved with context-mode  ·  ${savedPct.toFixed(1)}% reduction`

The release notes for v1.0.118 said both surfaces would be unified on bytes, but efbd294 only touched the statusline. Not blocking this PR — happy to do the analytics-side cleanup in a follow-up if it's wanted, or leave it for you. Flagging because it explains why the kb-formatter regression test only pins "no $ in statusline" and not "no $ anywhere".

mksglu:
Closing as superseded by [v1.0.119](https://github.com/mksglu/context-mode/releases/tag/v1.0.119).

Your test contract for the v1.0.118 byte-based statusline render (commits in this PR) was the right diagnosis — the existing assertions did need updating. The single commit `59ac5fd` covering the assertion update landed on `next` directly, and v1.0.119's broader Windows test isolation work (commits `a16c3dc`, `bd5821a`) extends `tests/setup-home.ts` with a `withIsolatedEnv()` helper that scopes `APPDATA`/`LOCALAPPDATA`/`XDG_*`/`TMPDIR` — closing the Windows runner gap your PR's BRAND_NEW assertions were hitting.

This PR's branch is in CONFLICT with `main` and the work is already incorporated, so closing without merge to clean up. Your contribution is fully credited in the v1.0.119 release notes.

Thank you for the deep statusline test work and for catching the v1.0.118 contract drift — both made v1.0.119 a much tighter ship. ✨

PR #537, opened 2026-05-12
fix: eliminate DEP0190 DeprecationWarning on Windows

frhino:
## Problem

Node.js DEP0190 fires when `spawn()`/`execFileSync()` are called with `shell: true` and arguments passed as a separate array. On Windows, the args array is unsafely concatenated into the shell command string — Node.js rightfully warns about this.

This warning appears on every pi startup (visible as `[mcp-bridge] (node:XXXX) [DEP0190] DeprecationWarning` in the terminal).

## Root Cause

Three call sites in the source use the args-array form with `shell: true` on Windows:

1. `runtime.ts` — `runnableExists()`: `execFileSync(cmd, [--version], { shell: isWindows })`
2. `runtime.ts` — `getVersion()`: `execFileSync(cmd, args, { shell: process.platform === win32 })`
3. `executor.ts` — `#spawn()`: `spawn(spawnCmd, spawnArgs, { shell: needsShell })` when `needsShell` is true (`tsx`/`ts-node`/`elixir`/`bun` on Windows)

## Fix

When shell is needed on Windows, collapse the command and arguments into a single string before passing to the child_process function. Leave the shell-free code path (POSIX) unchanged.

- **`runtime.ts`**: split into `execSync(cmdString)` on Windows vs `execFileSync(cmd, args)` without `shell` on POSIX
- **`executor.ts`**: when `needsShell` is true, join `spawnCmd + spawnArgs` into a single string, pass empty args array; otherwise keep the args-array form with `shell: false`

## Testing

Ran the original (pre-fix) bundle and the fixed bundle through the same MCP handshake + `ctx_execute` tool call:

**Test methodology:** spawned server.bundle.mjs as a child process, performed full MCP handshake (initialize → notifications/initialized → tools/list → tools/call for ctx_execute), captured all stderr.

| Bundle | stderr | Result |
|--------|--------|--------|
| Original (main) | 264 bytes, contains `[DEP0190]` | ❌ |
| Fixed (this PR) | 0 bytes | ✅ |

Also ran `npm test`: all 103 executor tests pass, all 136 cli tests pass (1 unrelated pre-existing failure in plugin manifest substitution).


mksglu:
@frhino CI broken

frhino:
Thank you for the cleanup! I thought the full test suite was run but it wasn't. Won't let that happen again.

PR #549, opened 2026-05-13
fix(claude-code): quote pluginRoot in generated hook commands so paths with spaces parse correctly

jstan2525:
## Summary

On Windows where the user folder contains a space (e.g. `C:\Users\High Ground Services\...`), `ctx doctor` reports all 5 hook scripts as **`Hook script exists: FAIL`** at a doubled path like:

```
C:\Users\High Ground Services\AppData\Roaming\npm\node_modules\context-mode\Services\AppData\Roaming\npm\node_modules\context-mode\hooks\pretooluse.mjs
```

The scripts are actually present at the correct location and the hooks themselves execute fine — only the doctor's diagnostic mis-reports.

## Root cause

`ClaudeCodeAdapter.generateHookConfig` builds command strings with an **unquoted** `${pluginRoot}` interpolation:

```ts
// src/adapters/claude-code/index.ts (before)
const preToolUseCommand = `node ${pluginRoot}/hooks/pretooluse.mjs`;
```

When `pluginRoot` contains spaces the rendered command looks like:

```
node C:\Users\High Ground Services\AppData\Roaming\npm\node_modules\context-mode/hooks/pretooluse.mjs
```

`extractHookScriptPath` in `src/util/hook-config.ts` uses:

```ts
const match = command.match(/(?:"([^"]+\.mjs)"|'([^']+\.mjs)'|(\S+\.mjs))/);
```

With no quotes to anchor the match, the regex falls through to its `\S+\.mjs` branch and grabs only the path tail after the last space (`Services\AppData\...\pretooluse.mjs`). The cli.ts caller then does `resolve(pluginRoot, scriptPath)` which appends that tail to pluginRoot, producing the doubled path that fails `accessSync`.

## Fix

Wrap each `${pluginRoot}/hooks/<name>.mjs` in double quotes, matching the style already used by [`hooks/hooks.json`](../blob/main/hooks/hooks.json):

```ts
// after
const preToolUseCommand = `node \"${pluginRoot}/hooks/pretooluse.mjs\"`;
```

The existing `extractHookScriptPath` regex already has a `\"([^\"]+\.mjs)\"` branch that handles this correctly.

No behavior change on platforms / installs where `pluginRoot` has no spaces — quoted paths execute identically on every shell that consumes hooks.json.

## Regression test

Added `tests/adapters/claude-code.test.ts` test that:
- Builds the hook config with a Windows-style `pluginRoot` (`C:\Users\High Ground Services\...`)
- Pipes each command through `extractHookScriptPath`
- Asserts the extracted path still starts with `pluginRoot` (not a relative tail)

All 53 tests in `claude-code.test.ts` pass.

## Repro before this fix

On Windows with a space in the user folder, after `npm install -g context-mode@latest`:

```
$ context-mode doctor
...
■ Hook script exists: FAIL — not found at C:\Users\<First Last>\AppData\Roaming\npm\node_modules\context-mode\<Last>\AppData\Roaming\npm\node_modules\context-mode\hooks\pretooluse.mjs
(repeated 5x for posttooluse / precompact / userpromptsubmit / sessionstart)
```

After this fix, `ctx doctor` reports all 5 hooks as PASS on the same machine.

## Test plan

- [x] `npx vitest run tests/adapters/claude-code.test.ts` — 53 passed
- [ ] CI green
- [ ] Manual verification on Windows host with spaces in user folder

### DIFF
--- a/src/adapters/claude-code/index.ts
+++ b/src/adapters/claude-code/index.ts
@@ -106,7 +106,13 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
   }
 
   generateHookConfig(pluginRoot: string): HookRegistration {
-    const preToolUseCommand = `node ${pluginRoot}/hooks/pretooluse.mjs`;
+    // Paths must be double-quoted so that hosts (or our own diagnostic
+    // regex) parse them correctly when pluginRoot contains spaces — common
+    // on Windows where the user folder is e.g. "C:\Users\First Last\...".
+    // Without quotes, extractHookScriptPath's `\S+\.mjs` fallback grabs
+    // only the tail after the last space, producing a doubled-path FAIL
+    // in `ctx doctor`. Matches the quoting style used in hooks/hooks.json.
+    const preToolUseCommand = `node "${pluginRoot}/hooks/pretooluse.mjs"`;
     const preToolUseMatchers = [...PRE_TOOL_USE_MATCHERS];
 
     return {
@@ -120,7 +126,7 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
           hooks: [
             {
               type: "command",
-              command: `node ${pluginRoot}/hooks/posttooluse.mjs`,
+              command: `node "${pluginRoot}/hooks/posttooluse.mjs"`,
             },
           ],
         },
@@ -131,7 +137,7 @@ export class ClaudeCodeAdapter extends ClaudeCodeBaseAdapter implements HookAdap
           hooks: [
             {
               type: "command",
-              command: `node ${pluginRoot}/hooks/precompact.mjs`,
+              command: `node "${pluginRoot}/hooks/precompact.mjs"`,
             },
           ],
         },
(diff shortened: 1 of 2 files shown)


mksglu:
Merged and shipped in v1.0.125. Thank you @jstan2525 — precise diagnosis with empirical reproduction (the `\S+\.mjs` fallback in `extractHookScriptPath` grabbing only tail-after-last-space, then `resolve(pluginRoot, scriptPath)` doubling). 5 surgical quote additions, regression test included. This is the kind of PR that lands clean.

## 15-adapter audit (per our MUST-3 invariant)

Verified the bug lived ONLY in `claude-code/index.ts`. The other 14 hook-emitting adapters route through `buildNodeCommand` (`src/adapters/types.ts:332`) which already double-quotes both `nodePath` and `scriptPath` and forward-slash-normalizes (introduced for #369/#372). Your fix is complete for the reported bug; no other adapter is affected.

## Open follow-up (not blocking, future cleanup)

The architectural trend is to migrate all hook command builders to `buildNodeCommand` so claude-code stops being the lone outlier. That's a 1-line cost per adapter that brings claude-code into parity with the canonical helper and eliminates an entire class of future Windows quoting bugs. Tracked separately, not blocking your fix.

Thank you for the Windows-aware contribution — paths-with-spaces is a long-standing class of bugs and your fix unblocks an entire user population. Release notes: https://github.com/mksglu/context-mode/releases/tag/v1.0.125

PR #570, opened 2026-05-14
docs: prohibit head/tail truncation inside ctx_execute sandbox

### LINKED ISSUE #569: docs: ctx_execute sandbox truncation anti-pattern missing from guidance
## Problem

Claude correctly routes large-output commands through `ctx_execute` (per existing guidance), but then applies `head`/`tail` *inside* the shell code to cap the output — e.g.:

```shell
# What Claude does today
ctx_execute(language="shell", code="grep ERROR app.log | head -50")
```

This defeats the purpose in a subtle way that isn't covered anywhere in the current docs:

- Inside `ctx_execute`, stdout goes to the **FTS5 sandbox**, not to context.
- There is **no output-size penalty** for letting the command run in full.
- Truncating inside the sandbox discards data **before indexing** — the rest is permanently lost and unsearchable.
- The correct pattern is programmatic analysis: filter/count/aggregate and `console.log()` only the findings.

## Why it happens

The existing anti-pattern bullet in `SKILL.md` reads:

> Piping Bash output through `| head -20` → you lose the rest. Use `ctx_execute` to analyze ALL data and print summary.

Claude learns to move the command into `ctx_execute`, but carries over the truncation reflex because nothing explains that **the sandbox has a different model** from context. The mental model gap: Claude treats the sandbox like a smaller context window rather than an indexed store.

## Proposed fix

Three files:

1. **`skills/context-mode/SKILL.md`** — add a second bullet to the Anti-Patterns section making the inside-sandbox case explicit.
2. **`skills/context-mode/references/anti-patterns.md`** — new section §8 with before/after examples and the corrected programmatic analysis pattern.
3. **`configs/claude-code/CLAUDE.md`** — one-liner callout under *Think in Code*.

Also fixes a pre-existing issue: the `SKILL.md` frontmatter `description` field uses a multi-line `|` block, which breaks plugin/skill discovery. Collapsed to a single line.

A PR with the changes is ready: https://github.com/halindrome/context-mode/compare/main...halindrome:context-mode:docs/no-head-tail-inside-ctx-execute

halindrome:
Closes #569

## Summary

- **`skills/context-mode/SKILL.md`**: adds a second Anti-Patterns bullet covering `head`/`tail` truncation *inside* `ctx_execute` (distinct from the existing Bash `| head` rule); also collapses the multi-line `description: |` frontmatter to a single line to fix plugin/skill discovery
- **`skills/context-mode/references/anti-patterns.md`**: new section §8 — "Using head/tail Inside ctx_execute" — with before/after shell and JS examples and the corrected programmatic-analysis pattern
- **`configs/claude-code/CLAUDE.md`**: one-liner callout added under *Think in Code* so the rule surfaces in the installed project config

## Why this matters

The existing guidance says "don't pipe through `| head -20` in Bash — use `ctx_execute` instead." Claude follows that, but then applies the same truncation reflex *inside* `ctx_execute` shell code. The two cases have opposite failure modes:

| Location | Effect of `head`/`tail` |
|---|---|
| Bash command | output floods context window |
| Inside `ctx_execute` | data discarded before FTS5 indexing — permanently lost |

Nothing in the current docs explains that the sandbox has no output-size penalty and that truncation inside it is always wrong.

## Test plan

- [ ] Read `skills/context-mode/references/anti-patterns.md` section §8 — examples are clear and the rule is unambiguous
- [ ] Read `skills/context-mode/SKILL.md` Anti-Patterns section — new bullet distinguishes Bash-level vs sandbox-level truncation
- [ ] Read `configs/claude-code/CLAUDE.md` Think in Code section — callout is present and concise
- [ ] Confirm `SKILL.md` frontmatter `description` is now a single line


mksglu:
The PR is not up to date
There are files that shouldn't be there

PR #571, opened 2026-05-14
fix(win32): detect VS year via vswhere displayName for VS 2026+

### LINKED ISSUE #566: [Bug]: VS 2026 not recognised by node-gyp during better-sqlite3 build
### Platform

Claude Code

### context-mode version

1.0.131

```json
NA
```

```text
NA
```

```text
NA
```

### Steps to reproduce (REQUIRED)

- OS: Windows 11 (10.0.26200)
  - Node.js: v20.19.6
  - npm: 10.8.2 (at time of failure)
  - Visual Studio: 2026 Community (18.6.11806.211) with "Desktop development with C++" workload
  - context-mode: v1.0.122 → upgrading to v1.0.131

### What have you tried to fix it?

Running `/ctx-upgrade` fails during `npm install` because `better-sqlite3` has no
  prebuilt binary for this Node/platform combination, so it falls back to compiling
  from source via `node-gyp`. The compilation fails because node-gyp cannot identify
  the installed Visual Studio version.

  ## Root Cause (3-layer chain)

  1. **No prebuilt binary** — `prebuild-install@7.1.3` (bundled inside `better-sqlite3`)
     cannot find a prebuilt `.node` file for the running Node version, so it falls
     through to `node-gyp rebuild`.

  2. **node-gyp uses a hardcoded version table** — It calls `vswhere.exe` which returns
     the internal VS version number (`18.6.11806.211` for VS 2026). Older node-gyp (≤10.x,
     bundled with npm ≤10.x) has a hardcoded map of `{ 15: '2017', 16: '2019', 17: '2022' }`
     and produces `"unknown version"` for major version 18.

  3. **Even node-gyp 12.x requires an explicit hint** — After upgrading npm (which bundles
     node-gyp 12.3.0 that does recognise VS 2026), the build still fails unless
     `npm_config_msvs_version=2026` is set in the environment, because VS 2022 is the
     default expected version.

 ## Error Output

  npm error prebuild-install warn install No prebuilt binaries found
    (target=20.19.6 runtime=node arch=x64 libc= platform=win32)
  npm error gyp ERR! find VS unknown version "undefined" found at
    "C:\Program Files\Microsoft Visual Studio\18\Community"
  npm error gyp ERR! find VS could not find a version of Visual Studio 2017 or newer to use

## Workaround (confirmed working)

  1. Upgrade npm: `npm install -g npm@latest`
  2. Add to global npmrc (`npm config get globalconfig`):
     msvs_version=2026

## Proposed Fix

  `scripts/heal-better-sqlite3.mjs` already has a `buildSafeEnv()` function that
  patches the child process environment for known platform quirks (conda Python on
  macOS/Linux — see #408, #533). The same pattern should be applied for Windows VS
  detection.

  Instead of hardcoding a year map (which would break again with VS 2029, VS 2032,
  etc.), use `vswhere.exe -latest -property catalog_productLineVersion` which returns
  the human-readable year string directly (`"2026"`, `"2022"`, etc.) — no mapping table
  needed, future-proof for any VS release.
  This fix also needs to be applied to the upgrade CLI's main npm install invocation
  (the temp-dir install), not just the heal layers. 

**I will raise a PR for the fix.**

References

  - Same class of issue as #408 (Windows prebuild-install PATH) and #533 (conda Python)
  - catalog_productLineVersion vswhere property: always returns the year string,
  stable across all VS versions

### Pre-submission checklist

### Operating System

Windows (Git Bash)

### JS Runtime

_No response_

Kishan08:
Fixes #566

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [x] All platforms

## Test plan

12 tests in tests/util/heal-better-sqlite3-windows-vs.test.ts

detectWindowsVsYear() unit tests (dependency-injected, run on any OS):

  1. exports detectWindowsVsYear() as a function
  2. returns null on non-Windows platforms — darwin, linux, freebsd
  3. returns null when vswhere.exe is absent
  4. extracts year from vswhere displayName for VS 2026
  5. works for all VS editions (Community, Professional, Enterprise)
  6. works for older VS versions (2017, 2019, 2022)
  7. trims whitespace and CRLF from vswhere output before matching
  8. returns null when vswhere throws
  9. returns null when vswhere returns a string with no 4-digit year — e.g. "18", "", "unknown"

  source-code assertions on buildSafeEnv:

  10. uses displayName to get the year and extracts it with a regex — asserts -property displayName and /20\d{2}/ are in source
  11. buildSafeEnv sets npm_config_msvs_version on Windows via vswhere
  12. buildSafeEnv only sets npm_config_msvs_version when not already present — respects user's existing npmrc value

## Checklist

- [x] `npm test` passes


mksglu:
do NOT create seperare unit test for that pls use existing one

Kishan08:
@mksglu merged into existing test file.

PR #572, opened 2026-05-14
[codex] Bound SQLite write hot paths

sborisov88:
## Summary
- Replaces CPU-spinning SQLite retry waits with parked sync waits plus an async retry primitive.
- Adds per-DB write queue/backpressure for FTS and SessionDB write paths, including coalesced session stats/event flushes.
- Defers synchronous ctx_batch_execute fresh-search/vocabulary work for large outputs and returns deterministic ctx_search guidance.

## Root Cause
Large command outputs and rapid tool responses could concentrate synchronous SQLite/FTS work on the response hot path: blocking retry loops, unqueued multi-instance FTS writes, per-response SessionDB open/write/close storms, and fresh searches immediately after large batch indexing.

## Validation
- npm run typecheck
- npm run build
- CI=1 npx vitest run --maxWorkers=1
- git diff --check HEAD~1..HEAD


EPPCOM-Solutions (changes requested):
## Request Changes

The write-queue architecture in `src/db-write-queue.ts` is sound — `resolve(dbPath)` as queue key is cross-platform safe, failure isolation prevents stranded jobs, and the test coverage across 20 test files is substantial.

However, before merging a +6589/-697 diff, two confirmations are required:

### Required before merge

**1. CI green on all 3 OS**
The PR validation section lists `npm run typecheck`, `npm run build`, `npx vitest run` — please link the CI run showing green on Windows, Linux, and macOS. The diff touches `src/db-write-queue.ts`, `src/session/db.ts`, `src/store.ts`, and native binding paths — all of which have OS-specific behavior.

**2. Bundle regeneration confirmed**
The diff includes changes to `hooks/session-db.bundle.mjs`, `cli.bundle.mjs`, `server.bundle.mjs`, and `hooks/security.bundle.mjs`. Please confirm these were regenerated from the modified source (not manually edited) and that `npm run build` was the last step before pushing.

### Already verified
- `db-write-queue.ts` — per-DB serialization with failure isolation, correct
- `tests/util/db-write-queue.test.ts` + `tests/store-write-queue.test.ts` — write-queue behavior covered
- `tests/perf-smoke.ts` + `tests/perf-baseline.json` — regression baseline added
- No schema migrations — this is a pure concurrency layer, rollback is a revert
- Windows path handling: `resolve(dbPath)` is correct

Once CI links are provided and bundle regen confirmed, this is ready to merge.

mksglu:
What're u doing there be alone? 😂 @sborisov88

PR #575, opened 2026-05-15
fix(codex): harden plugin hook routing

TyceHerrman:
## Summary

- ship the Codex plugin hook manifest and package it with npm
- route Codex exec_command `{ cmd }` payloads through the same Bash policy path as `{ command }`
- force bundled Codex hook commands to run with `CONTEXT_MODE_PLATFORM=codex` while preserving the current plugin-root fallback
- clarify Codex install docs around MCP verification, plugin hook trust, and manual fallback hooks

## Validation

- `./node_modules/.bin/vitest run tests/hooks/core-routing.test.ts tests/plugins/codex-manifest.test.ts tests/codex/marketplace-layout.test.ts tests/core/cli.test.ts`
- `git diff --check`
- `NPM_CONFIG_CACHE=/tmp/context-mode-npm-cache npm pack --dry-run --json`
- `npm run assert-bundle`


TyceHerrman:
CI follow-up after `a6eb8cd`:

- Fixed the branch-owned Windows failure in `tests/hooks/core-routing.test.ts` by using `CODEX_HOME` for the Codex settings path.
- Current PR CI now shows `test (windows-latest)` passing, plus macOS and OpenClaw jobs passing.
- Remaining red check is `test (ubuntu-latest)`, failing in `tests/analytics/format-report.test.ts` with `RangeError: Incorrect locale information provided` from `src/session/analytics.ts`.
- That Ubuntu analytics failure is already present on upstream `next` at the PR base (`914b0c6`): run https://github.com/mksglu/context-mode/actions/runs/25887250971 shows the same `RangeError` in the same test, so it is not introduced by this Codex plugin branch.

Validation run locally before push:

```bash
./node_modules/.bin/vitest run tests/hooks/core-routing.test.ts tests/plugins/codex-manifest.test.ts tests/codex/marketplace-layout.test.ts tests/core/cli.test.ts
git diff --check
NPM_CONFIG_CACHE=/tmp/context-mode-npm-cache npm pack --dry-run --json
npm run assert-bundle
```

mksglu:
Hi @TyceHerrman Did you tested manually? Thx.

TyceHerrman:
Hi @mksglu, yes. I tested it locally after the CI fix, including a manual Codex plugin smoke test.

Local automated checks run:

```bash
./node_modules/.bin/vitest run tests/hooks/core-routing.test.ts tests/plugins/codex-manifest.test.ts tests/codex/marketplace-layout.test.ts tests/core/cli.test.ts
git diff --check
NPM_CONFIG_CACHE=/tmp/context-mode-npm-cache npm pack --dry-run --json
npm run assert-bundle
```

Results:

- Vitest passed: 4 files / 252 tests.
- `git diff --check` was clean.
- `npm pack --dry-run --json` included `.codex-plugin/hooks.json`.
- `npm run assert-bundle` passed without requiring bundle changes.

Manual Codex smoke test:

- Used isolated temp `CODEX_HOME`, so no real `~/.codex` config was changed.
- Ran `codex plugin marketplace add /Users/tyceherrman/GitHub/forks/context-mode`.
- Verified temp Codex config had `hooks = true`, `plugin_hooks = true`, and the local marketplace source.
- Verified the plugin hook manifest command:
  - sets `CONTEXT_MODE_PLATFORM='codex'`
  - resolves plugin root from `PLUGIN_ROOT || CLAUDE_PLUGIN_ROOT`
- Invoked the actual `.codex-plugin/hooks.json` `PreToolUse` command with Codex-shaped stdin:
  - `exec_command` with `{ "cmd": "echo blocked" }` and temp `$CODEX_HOME/settings.json` deny policy returned `permissionDecision: "deny"`.
  - Same test passed with only `CLAUDE_PLUGIN_ROOT` set, proving the compatibility fallback works.
  - `{ "cmd": "echo harmless" }` returned normal `PreToolUse` output with no deny.
- Also ran `codex debug prompt-input --enable hooks --enable plugin_hooks`, which saw `context-mode` and `.codex-plugin`.

One limit: I did not run a full authenticated interactive Codex session with the real hook trust UI, because I kept the test isolated from my real Codex config/auth. The branch-owned runtime path and plugin manifest path were tested locally.

EPPCOM-Solutions (approved):
## Approved ✅

Codex `exec_command` payload routing and platform settings path wiring are both correct and necessary.

**What was verified:**
- `getShellCommand()` at `routing.mjs:499–505` — correctly normalizes Codex `exec_command { cmd }` payloads alongside Claude `{ command }`; without this, Codex shell commands bypassed Bash policy checks entirely
- `getPlatformSettingsPath()` at `routing.mjs:513–516` — threads `~/.codex/settings.json` (or `$CODEX_HOME`) into `security.readBashPolicies()` and `security.readToolDenyPatterns()`; Codex users' deny rules were previously silently ignored
- `core-routing.test.ts` — `exec_command cmd` denial tests + full policy-enforcement fixture with proper `beforeEach`/`afterEach` env cleanup

**Non-blocking note:**
`.codex-plugin/hooks.json:16` — PostToolUse entry has no `matcher` field (PreToolUse has one). For Claude Code the wildcard behavior is identical, but being explicit would improve reviewability. Not a blocker.

This PR does **not** fix issue #587 (macOS cold-start race in `ensure-deps.mjs`) — that needs a separate PR.

.codex-plugin/hooks.json
```diff
@@ -0,0 +1,65 @@
...
+          }
+        ]
+      }
+    ],
+    "Stop": [
+      {
+        "hooks": [
+          {
+            "type": "command",
+            "command": "node -e \"const r=process.env.PLUGIN_ROOT||process.env.CLAUDE_PLUGIN_ROOT;if(!r)throw new Error('PLUGIN_ROOT missing');process.env.CONTEXT_MODE_PLATFORM='codex';Promise.all([import('node:path'),import('node:url')]).then(([p,u])=>import(u.pathToFileURL(p.join(r,'hooks/codex/stop.mjs')).href))\""
```
mksglu: I had never seen in that way in code-base why we follow that code:

"command": "node -e \"const r=process.env.PLUGIN_ROOT||process.env.CLAUDE_PLUGIN_ROOT;if(!r)throw new Error('PLUGIN_ROOT missing');process.env.CONTEXT_MODE_PLATFORM='codex';Promise.all([import('node:path'),import('node:url')]).then(([p,u])=>import(u.pathToFileURL(p.join(r,'hooks/codex/stop.mjs')).href))\""

mksglu (changes requested):

.codex-plugin/hooks.json
mksglu: @TyceHerrman

TyceHerrman: The inline `node -e` was a conservative workaround for resolving the plugin root and forcing `CONTEXT_MODE_PLATFORM=codex` before the shared hook code loads, but I rechecked codex upstream and found a fix was merged so now codex handles placeholder substitution. openai/codex#19705 injects and substitutes `PLUGIN_ROOT` / `CLAUDE_PLUGIN_ROOT` for plugin-provided hooks, so we do not need this JS.

I'll simplify this to `node "${PLUGIN_ROOT}/hooks/codex/stop.mjs"`, and move the codex platform forcing into the `hooks/codex/*.mjs` wrapper modules instead.

ousamabenyounes:
Hey @TyceHerrman, @mksglu — I ran an empirical check on the test-discipline side of this PR. Process: checked out the branch, reverted *only* `hooks/core/routing.mjs` back to `HEAD^` while keeping the new tests in place, ran the routing test file, restored, ran again.

## Verified

Four test files cited in the description on the PR branch:
- `tests/hooks/core-routing.test.ts tests/plugins/codex-manifest.test.ts tests/codex/marketplace-layout.test.ts tests/core/cli.test.ts` → **252 / 252 pass**

Bypass demonstration (`hooks/core/routing.mjs` reverted, all new tests kept):

| State | `core-routing.test.ts` | `exec_command { cmd }` policy bypass | `~/.codex/settings.json` deny bypass |
|---|---|---|---|
| Production code reverted, new tests in place | **2 FAIL / 77 pass** | open | open |
| PR applied (current HEAD) | **79 / 79 pass** | closed | closed |

The two failures with the fix reverted are exactly:
- `routePreToolUse > Bash tool > denies Codex exec_command cmd payloads like Bash command payloads`
- `routePreToolUse > Codex exec_command security policy > denies Codex exec_command cmd payloads from .codex settings`

So the new tests genuinely fail when the patch is reverted — test discipline is satisfied, and the 77 other tests in the same file stay green (no latent regression). CI is already green on ubuntu/macos/windows.

## On the inline `node -e "..."` (@mksglu's comment on hooks.json:59)

FWIW the pattern is doing four things that all need to happen *before* the hook module loads:

1. Codex's `hooks.json` schema accepts only a command string, not a script file reference — so the wrapper has to be inline.
2. `CONTEXT_MODE_PLATFORM=codex` must be set **before** the hook module is imported, otherwise the routing layer can't differentiate Codex from Claude. Inline assignment guarantees that ordering.
3. `PLUGIN_ROOT || CLAUDE_PLUGIN_ROOT` fallback covers Codex's env-var name plus the Claude-style name the manual fallback path expects.
4. `pathToFileURL(...).href` is needed for ESM `import()` on Windows — plain file paths don't satisfy the URL scheme requirement.

A cleaner future iteration could factor this into a single `hooks/codex/_loader.mjs <hookName>` shim — if Codex's `command` field accepts args (`"node ${PLUGIN_ROOT}/hooks/codex/_loader.mjs stop"`), the 6 duplicated wrappers collapse to one.

**My read:** mergeable as-is given the security value (real bypass closed, with tests that fail at revert); the wrapper cleanup is worth a separate cosmetic PR rather than blocking this one.

TyceHerrman:
@ousamabenyounes The revert-test evidence is useful for showing the new routing tests for cmd are working. the inline `node -e` analysis is stale after the latest commits. 

@mksglu PR ready for review

.codex-plugin/hooks.json
mksglu: No need inline for that @TyceHerrman

PR #578, opened 2026-05-15
fix(linux): re-exec with Bun on Linux to avoid better-sqlite3 SIGSEGV (#564)

### LINKED ISSUE #564: Linux: SIGSEGV on first tool call caused by node/bun runtime mismatch (with workaround)
Related to #556, which was closed without a confirmed root cause on Linux.

**Environment**
- OS: Linux (Ubuntu, kernel 6.17)
- Claude Code: 2.1.141
- context-mode: 1.0.131
- Node: 20.20.2 (nvm)
- Bun: 1.3.13

**Symptoms**

The MCP server connects successfully at session startup (243ms), but crashes with SIGSEGV the first time any `ctx_*` tool is invoked, roughly 10–12 seconds later. The skill falls back to CLI. Subsequent tool calls in the same session fail with "dynamic tool unavailable" because the server is dead.

**Root cause**

The plugin loader invokes `start.mjs` via `node` (per `.claude-plugin/plugin.json`'s `"command": "node"`). On Linux, `better-sqlite3`'s native addon segfaults under node's V8 — it requires bun's runtime environment. The doctor's SQLite check passes because it runs under bun; the actual server process doesn't.

Confirmed via a thin wrapper around `start.mjs` that logs the child process exit signal:

```
stdin data 306b       ← initialize
stdin data 202b       ← notifications/initialized
stdin data 177b       ← first tool call arrives
child exit code=null sig=SIGSEGV   ← 343ms later
```

**Workaround**

Backup the original and replace `start.mjs` with a wrapper that spawns `start.real.mjs` using bun:

```js
#!/usr/bin/env node
import { spawn } from 'child_process';
import { fileURLToPath } from 'url';
import { dirname, join } from 'path';

const __dirname = dirname(fileURLToPath(import.meta.url));
const child = spawn(process.env.HOME + '/.bun/bin/bun', [join(__dirname, 'start.real.mjs')], {
  stdio: ['pipe', 'inherit', 'inherit'],
  env: process.env,
});

process.stdin.on('data', (chunk) => { if (!child.stdin.destroyed) child.stdin.write(chunk); });
process.stdin.on('end', () => {});
const _k = setInterval(() => {}, 2147483647);
child.on('exit', (code) => { clearInterval(_k); process.exit(code ?? 0); });
```

With bun spawning the real server, `ctx_doctor` and other tools work correctly.

**Suggested fix**

The plugin manifest could specify `bun` as the command (with a node fallback), or `start.mjs` could detect the active runtime and re-exec itself under bun if it was invoked via node.

---

Note: this is separate from #559 (zombie processes after upgrade), though both surfaced during the same debugging session.

sghaskell:
## Problem

The fix in v1.0.132 (`hasModernSqlite()` in `ensure-deps.mjs`) skips the pre-flight probe on Node 22.5+, but `server.bundle.mjs` still has only two SQLite code paths at runtime:

- `bun:sqlite` — safe, activated when `typeof globalThis.Bun !== "undefined"`
- `require("better-sqlite3")` — SIGSEGV on Linux under Node's V8

Skipping the probe doesn't prevent the crash — it happens when the bundle loads better-sqlite3 at runtime, not during the probe. The result: v1.0.132–v1.0.134 still segfaults on Linux when invoked via `node`.

**Affected users:** Anyone on Linux with Node 20 (LTS), Node 22.0–22.4, or Node 22.5+ where the built-in SQLite wasn't compiled with FTS5.

## Fix

When `start.mjs` detects it's running under node on Linux, it searches for a Bun installation and re-execs itself under Bun. The node process then parks itself forwarding stdin while the Bun child owns the MCP session. Bun's runtime activates the `bun:sqlite` path in the bundle, avoiding better-sqlite3 entirely.

```
Claude Code → node start.mjs → (Linux + Bun found) → bun start.mjs → bun:sqlite ✓
Claude Code → node start.mjs → (Linux, no Bun)    → node continues  → existing behavior
```

Bun is detected at (in order):
1. `$BUN_INSTALL/bin/bun`
2. `~/.bun/bin/bun` (default install location)
3. `/usr/local/bin/bun`
4. `/usr/bin/bun`

Falls back to existing node behavior if Bun is not found — no regression for users without Bun.

## Testing

Tested on Linux (Ubuntu, kernel 6.17, Node v22.22.3, Bun 1.3.13, Claude Code v2.1.142):

**Happy path** — `node start.mjs` re-execs under Bun, full MCP handshake works:
```
$ echo '{"jsonrpc":"2.0","method":"initialize",...}' | node start.mjs
{"result":{"serverInfo":{"name":"context-mode","version":"1.0.134"},...}}
# tools/list: 11 tools returned, no segfault
```

**Real session** — `/ctx-doctor` calls MCP tool directly (no CLI fallback), `ctx_batch_execute` works correctly.

**Negative test** — Bun binary removed; server falls back to node and still responds to `initialize` without crashing.

## Notes

This is the approach suggested in the original bug report (#564). The `node:sqlite` path exists in `src/db-base.ts` (commit `302bb7b`) but is not yet in the published bundle for all affected Node versions — this fix is complementary and works without requiring a bundle rebuild.

Closes #564


sghaskell:
One additional note for context: even with this fix in place, Linux users will need to manually register the MCP server via `claude mcp add --scope user` to get tool access in Claude Code sessions. Plugin-registered MCP servers (via `plugin.json` mcpServers) connect successfully but their tools don't appear as callable to the AI.

This is a Claude Code bug tracked here: https://github.com/anthropics/claude-code/issues/59310

The workaround:
```bash
claude mcp add --scope user plugin_context-mode_context-mode \
  /path/to/bun -- /path/to/plugin/start.mjs
```

Ideally the `/ctx-upgrade` script could detect and update an existing `~/.claude.json` user-level registration when upgrading to a new version, so users don't have to manually update the hardcoded version path after each upgrade.

mksglu:

```bash
```

PR #579, opened 2026-05-15
fix(upgrade): heal ~/.claude.json user MCP registrations after version bump

sghaskell:
## Problem

Users on Linux who work around [anthropics/claude-code#59310](https://github.com/anthropics/claude-code/issues/59310) (plugin-registered MCP servers don't expose tools to the AI) register the context-mode server manually:

```bash
claude mcp add --scope user plugin_context-mode_context-mode \
  /path/to/bun -- ~/.claude/plugins/cache/context-mode/context-mode/1.0.134/start.mjs
```

This writes a hardcoded absolute path to a specific version dir in `~/.claude.json`. After `/ctx-upgrade` installs a new version, the path is stale and MCP tools stop working until the user manually updates the version number.

## Fix

During upgrade, after the registry is synced, call `healClaudeJsonMcpArgs()` to scan `~/.claude.json` for any `mcpServers` entries whose args contain a path inside the context-mode plugin cache. Update the version dir segment to point at the new `pluginRoot`.

The heal function is extracted into `scripts/heal-installed-plugins.mjs` (the shared heal module) so it follows the same pattern as existing heals and is unit-testable.

- Best-effort: never throws, never blocks the upgrade
- Idempotent: no-op if args already point at the new version
- Logs when args are updated

## Tests

5 new unit tests in `tests/util/heal-installed-plugins.test.ts`:
- ✅ Updates stale version path to new pluginRoot
- ✅ No-op when args already point at new version  
- ✅ Skips when `~/.claude.json` doesn't exist
- ✅ Skips when `mcpServers` key is absent
- ✅ Ignores args that are not inside the context-mode cache

All 32 tests pass (27 existing + 5 new).

## Related

- PR #578 — fixes the Linux SIGSEGV that necessitates the manual registration workaround
- [anthropics/claude-code#59310](https://github.com/anthropics/claude-code/issues/59310) — the underlying Claude Code bug


mksglu:

```bash
```

Issue #378, opened 2026-04-28
Windows Claude Code: repeated hook loader failures when plugin uses 


Asureu:
### Summary

On native Windows Claude Code, `context-mode` has repeatedly hit startup/runtime hook failures after upgrade or reinstall when plugin-managed files use `${CLAUDE_PLUGIN_ROOT}` in hook commands and MCP startup args.

Observed failures include:

- `Failed with non-blocking status code: node:internal/modules/cjs/loader:1479`
- `SessionStart:startup hook error`
- `PostToolUse:Edit hook error`
- `PostToolUse:TaskUpdate hook error`
- `UserPromptSubmit hook error`

This has now been observed across multiple versions, including `1.0.98` and `1.0.103`.

### What keeps reproducing

After upgrade/reinstall, these plugin-managed files revert to placeholder-based paths again:

- `.claude-plugin/plugin.json`
- `hooks/hooks.json`

Examples from fresh plugin state:

```json
"args": ["${CLAUDE_PLUGIN_ROOT}/start.mjs"]
```

```json
"command": "node \"${CLAUDE_PLUGIN_ROOT}/hooks/sessionstart.mjs\""
```

In this Windows environment, that placeholder family repeatedly correlates with runtime loader failures.

### Important observations

- `ctx-doctor` can still pass.
- Direct `node --check` on hook scripts can still pass.
- Direct manual `node <absolute-path>` execution of hook scripts can still pass.
- Failures show up during real Claude runtime hook execution.
- Upgrades overwrite the local workaround and the problem returns.

### Working local workaround

Replacing `${CLAUDE_PLUGIN_ROOT}` with absolute Windows-safe paths in both places restores stability locally:

1. `hooks/hooks.json`
   - `posttooluse.mjs`
   - `precompact.mjs`
   - `pretooluse.mjs`
   - `userpromptsubmit.mjs`
   - `sessionstart.mjs`
2. `.claude-plugin/plugin.json`
   - replace `${CLAUDE_PLUGIN_ROOT}/start.mjs` with absolute path

After that:

- startup hook errors stop
- slash commands return
- MCP tools return
- `ctx-doctor` passes

### Why I think this is upstream-worthy

- It has reproduced across multiple versions.
- It is not durable: upgrade restores placeholder-based registration and the issue comes back.
- The current working fix is stable locally.
- This suggests either:
  1. `context-mode` should not rely on `${CLAUDE_PLUGIN_ROOT}` on native Windows Claude Code, or
[...]

PR #582, opened 2026-05-15
fix(linux): normalize hooks.json on Linux so bare node resolves for /bin/sh

sghaskell:
## Problem

`hooks/hooks.json` ships with `node "${CLAUDE_PLUGIN_ROOT}/hooks/posttooluse.mjs"` (and similar). Claude Code resolves `${CLAUDE_PLUGIN_ROOT}` but invokes the command via `/bin/sh`, which doesn't have nvm in PATH. On Linux with nvm, bare `node` fails:

```
PostToolUse:Edit hook error
Failed with non-blocking status code: /bin/sh: 1: node: not found
```

The hooks fail silently (non-blocking) but no tool usage is recorded in the session DB — context-mode's session tracking is effectively broken on Linux with nvm.

`normalizeHooksOnStartup()` already fixes this on Windows by rewriting bare `node` to `process.execPath` on every MCP boot. The fix was gated behind `platform !== "win32"` because the original issue (#378) was Windows-specific (MSYS path mangling). The same root cause — runtime not in PATH — affects Linux.

## Fix

Extend `normalizeHooksOnStartup()` to also run on Linux. macOS ships a system `node` at `/usr/bin/node` so it resolves reliably from `/bin/sh` — no change needed there.

On Linux, `process.execPath` is now bun (via PR #578 re-exec). All hook scripts are verified to work correctly under bun.

## Testing

- Verified hooks.json is normalized on Linux after MCP boot (absolute bun path replaces bare `node`)
- Verified PostToolUse/PreToolUse hooks run without error in a live Claude Code session
- Session DB `event_count` increments correctly after tool calls
- 40 tests pass (13 hook normalization + 27 heal)

## Related

- PR #578 — bun re-exec fix (makes `process.execPath` = bun on Linux)
- Issue #378 — original Windows normalization fix


mksglu:

```bash
```

PR #594, opened 2026-05-16
feat: Make context-mode output collapsible

mrinfinidy:
## What / Why / How

When using context-mode with Pi it prints all the output to the agent. This clutters the conversation history.
To solve this problem this PR adds collapsibles that only show the the title of the tool use:
image
The output is collapsed by default.
image
By pressing Ctrl+O (default Pi keybinding) the collapsed output can be displayed.

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [x] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

No automated tests added. The collapsible wrapping is pure output formatting in the MCP bridge adapter layer. It applies markdown syntax around existing tool output and doesn't introduce new logic that can regress silently. Behavior was manually verified against a live Pi installation using the dev build (see screenshot): output renders collapsed by default, and Ctrl+O expands it as expected.

## Checklist

- [x] `npm test` passes


DaniBedz:
This would be great to add. The way it works currently in pi is very noisy.

Issue #592, opened 2026-05-16
Claude Code adapter ctx_* tools still go stale after idle shutdown on v1.0.135 (#583 fix was Pi-only)


skoal:
## Summary

The idle self-shutdown introduced in v1.0.132 (PR #568) still strands the **Claude Code** adapter's `ctx_*` tools on v1.0.135. The Pi-adapter respawn fix delivered in v1.0.135 (closing #583) does not extend to Claude Code: once `start.mjs` exits cleanly on the 15-minute idle timer, Claude Code surfaces every registered `ctx_*` tool as `MCP server disconnected` and does not auto-respawn the child.

Effective state on 1.0.135 + Claude Code: the symptom from #576 ("MCP server (start.mjs) dies silently mid-session") is still reproducible if MCP tools idle past `CONTEXT_MODE_IDLE_TIMEOUT_MS` while the host editor remains active on non-MCP work.

## Environment

- **context-mode**: v1.0.135 (also v1.0.131 in cache; v1.0.135 is the resolved version)
- **Claude Code**: 2.1.143 (Opus 4.7 model, 1M context)
- **OS**: macOS 26.3.1 (Darwin 25.3.0, arm64)
- **Node**: v26.0.0 (Homebrew, `/opt/homebrew/bin/node`)
- **Bun**: 1.3.13 (installed but not the runtime here — start.mjs's `runtimeIsBun` gate is false on macOS)
- **Shell**: zsh

## Reproduction

1. Start a Claude Code session with the context-mode plugin enabled. Default env (no `CONTEXT_MODE_IDLE_TIMEOUT_MS` override).
2. Run a burst of `ctx_*` calls — in our case ~25 calls including `ctx_execute`, `ctx_execute_file`, `ctx_batch_execute` for log analysis and small probe scripts. All succeed.
3. Switch to non-MCP work in the same session (file edits via Edit/Write, Bash for git/diagnostics, sub-agent dispatch). For our session, ~15 minutes elapsed without any `ctx_*` call.
4. Attempt the next `ctx_*` call.

## Observed

Claude Code injects the following system-reminder before the next assistant turn:

```
The following deferred tools are no longer available (their MCP server disconnected). Do not search for them — ToolSearch will return no match:
mcp__plugin_context-mode_context-mode__ctx_batch_execute
mcp__plugin_context-mode_context-mode__ctx_doctor
mcp__plugin_context-mode_context-mode__ctx_execute
[...]

Issue #583, opened 2026-05-15
Pi ctx_* tools become stale after MCP idle shutdown (regression in 1.0.132)


zer09:
## Summary

In Pi, `ctx_*` tools become permanently stale after the context-mode MCP child exits due to idle shutdown. The regression starts in `context-mode@1.0.132`: `1.0.131` survives the same short-idle test, while `1.0.132` fails with `MCP server has exited`.

## Environment

- OS: Linux WSL2
- Kernel: `6.6.87.2-microsoft-standard-WSL2`
- Platform: Pi coding agent
- Node: `v24.14.1`
- Bun: `1.3.13`
- npm global package path: `~/.nvm/versions/node/v24.14.1/lib/node_modules/context-mode`
- Test env: `CONTEXT_MODE_IDLE_TIMEOUT_MS=30000`

## Reproduction

1. Install `context-mode@1.0.131`.
2. Reload Pi with:

   ```bash
   CONTEXT_MODE_IDLE_TIMEOUT_MS=30000 pi
   ```

3. Run `ctx_doctor` and an immediate `ctx_execute`.
4. Wait more than 30 seconds.
5. Run `ctx_execute` again.
6. Repeat the same steps with `context-mode@1.0.132`.

### `context-mode@1.0.131`

- `ctx_doctor`: OK
- immediate `ctx_execute`: OK
- after more than 30 seconds idle: `ctx_execute` still OK
- Example idle check time: `2026-05-15T06:29:46.984Z`

### `context-mode@1.0.132`

- `ctx_doctor`: OK
- immediate `ctx_execute`: OK at `2026-05-15T06:32:11.581Z`
- after idle: `ctx_execute` fails at `2026-05-15T06:34:58Z`
- error returned to Pi: `MCP server has exited`

### `context-mode@1.0.133`

- Same stale-child behavior was observed with the default 15 minute idle window.

## Additional observations

A short idle repro using `CONTEXT_MODE_IDLE_TIMEOUT_MS=1000` showed the MCP server exits cleanly:

- exit code: `0`
- signal: `null`
- stderr: empty

So this does not look like a server crash. It looks like a normal idle self-shutdown that leaves the Pi adapter's already-registered tools pointing at a dead child process.

## Suspected regression range

- `1.0.131` published: `2026-05-14T03:05:05.014Z`
- `1.0.132` published: `2026-05-14T17:49:56.634Z`

Between those npm tarballs, key changed files include:

- `build/lifecycle.js`
[...]

PR #568, opened 2026-05-14
fix(lifecycle): idle self-shutdown + generalized sibling sweep (#565)

omercnet:
## Summary

Fixes #565. OpenCode/KiloCode open one MCP child per session and per subagent task and never tear them down for the host's lifetime. On the reproducer in #565 a single `opencode serve` host live for 1d 18h accumulated **26 idle `context-mode` children, all parented to the same opencode PID, totalling ~1.6 GB RSS**.

This PR adds two complementary defences:

1. **Idle self-shutdown timer** in `src/lifecycle.ts` — quiescent MCP servers exit themselves.
2. **Generalized sibling discovery + startup sweep** in `src/util/sibling-mcp.ts` — a fresh MCP child reclaims stale siblings parented to the same host on boot.

Each is independently useful; together they keep the per-host MCP child count bounded at 1.

## Why none of the existing exit paths catch this

<details>
<summary>Walk-through</summary>

OpenCode integrates context-mode through two surfaces (per shipped `configs/opencode/opencode.json`): a TS plugin running in-process AND a stdio MCP server spawned per MCP client. Opencode opens a fresh MCP client per session/subagent, and:

- `lifecycle.ts` ppid poll: parent never dies, opencode stays alive → "alive forever".
- `lifecycle.ts` grandparent reparent (#311): targets `npm exec` wrapper; not present here (chain is direct `.opencode serve → node bin/context-mode`).
- `lifecycle.ts` stdin EOF: removed in #471; opencode keeps stdio open anyway.
- `lifecycle.ts` signals: only fire on opencode shutdown.
- `sibling-mcp.ts` (#559) regex: hard-coded to `~/.claude/plugins/{cache,marketplaces}/...start.mjs`. OpenCode's child argv is `node /home/USER/.npm-global/bin/context-mode` — never matched. And the killer is only invoked from `/ctx-upgrade`, not at MCP startup.

So idle MCP children accumulate linearly with sessions until the opencode host dies.
</details>

### 1. `src/lifecycle.ts` — idle self-shutdown (#565)

- New `idleTimeoutForEnv()` reads `CONTEXT_MODE_IDLE_TIMEOUT_MS`, defaults to **15 min**, accepts `0` to disable. Malformed values fall back to the default.
- `startLifecycleGuard()` now tracks `lastActivity` (injectable `now` for tests). The existing periodic timer additionally checks idle. Skipped when `process.stdin.isTTY` (interactive dev).
[...]

Issue #565, opened 2026-05-14
[Bug]: OpenCode accumulates one context-mode MCP child per session — never reaped (26 idle children, 1.6 GB RSS in repro)


omercnet:
## Summary

A single long-lived `opencode serve` host accumulates one `context-mode` MCP child process per session/subagent it opens. Children are never reaped while the host stays alive. On this machine, a single opencode host (uptime 1d 18h) has **26 idle `context-mode` children, all parented to the same opencode PID, totalling ~1.6 GB RSS**.

This is a new class of leak, distinct from #103 (orphan-on-host-death), #311 (npm-exec wrapper reparent), #534 (nested `pi --help`), #559 (upgrade-time leftover), and #562 (Pi subagent DB lock). None of those code paths fires here — the parent is alive, the chain is direct (no npm-exec wrapper), and sibling discovery is hard-coded to Claude plugin paths.

## Observed (live snapshot)

<details>
<summary><strong>26 <code>context-mode</code> children, single opencode parent (PID 23986), ~1.6 GB total RSS</strong></summary>

```
ps --ppid 23986 -o pid,ppid,etime,pcpu,rss,command
    PID    PPID     ELAPSED %CPU   RSS COMMAND
  25257   23986  1-18:51:26  0.1 64752 node /home/USER/.npm-global/bin/context-mode
  25273   23986  1-18:51:26  0.1 62708 node /home/USER/.npm-global/bin/context-mode
  25301   23986  1-18:51:26  0.1 62768 node /home/USER/.npm-global/bin/context-mode
  25412   23986  1-18:51:25  0.1 63248 node /home/USER/.npm-global/bin/context-mode
  25477   23986  1-18:51:25  0.1 61112 node /home/USER/.npm-global/bin/context-mode
  27197   23986  1-18:51:00  0.1 61492 node /home/USER/.npm-global/bin/context-mode
  27686   23986  1-18:50:52  0.1 63648 node /home/USER/.npm-global/bin/context-mode
  28601   23986  1-18:50:34  0.1 62584 node /home/USER/.npm-global/bin/context-mode
  28810   23986  1-18:50:33  0.1 61740 node /home/USER/.npm-global/bin/context-mode
  28993   23986  1-18:50:32  0.1 63092 node /home/USER/.npm-global/bin/context-mode
  29006   23986  1-18:50:32  0.1 62152 node /home/USER/.npm-global/bin/context-mode
[...]

PR #596, opened 2026-05-17
fix(lifecycle): runtime OpenCode auto-detect + gate sibling-sweep (#592 follow-up to #595)

ousamabenyounes:
## Summary

Flip the idle self-shutdown introduced in #568 to off-by-default for adapters
without auto-respawn (Claude Code, Codex, vanilla MCP clients) and keep it on
only for OpenCode-class hosts. Fixes the regression reported in #592 while
preserving the OpenCode/KiloCode memory protection from #565.

## Why

#568 enabled a 15-minute idle MCP self-shutdown for every adapter to bound
the OpenCode/KiloCode leak documented in #565 (1.6 GB RSS / 26 children
under a single `opencode serve` host). The side effect, surfaced by #592
two days later and confirmed publicly by both omercnet and mksglu:

- On **Claude Code**, **Codex**, and any host that does not auto-respawn
  its MCP child, after 15 minutes of idle the server exits. The host then
  surfaces `MCP server has exited` for every subsequent `ctx_*` call until
  the user manually restarts the session.
- Pi already has a respawn path (#583 + #585 follow-up); the other adapters
  do not, and the Pi respawn lives in our in-process bridge so it cannot
  be ported to host-owned MCP clients without an upstream change.

Both maintainers' positions on 2026-05-17:

- omercnet (PR #568 author): _\"please revert this, it's too complex for
  other adapters, and there's a better approach for opencode. If this is
  not an issue for other agents I'll focus on an opencode solution.\"_
- mksglu (owner): _\"yea lets revert for now and I'll rewrite it. Let's
  disable the idle-shutdown by default for adapters known not to
  auto-respawn — it's just only OpenCode I guess.\"_

This PR is the literal implementation of the second sentence, with all
the structural code from #568 kept in place so omercnet's planned
MCPless-OpenCode rewrite can flip the env on its own terms later.

## Coverage added

- `idleTimeoutForEnv()`:
  - returns `0` (off) for `{}` and any non-OpenCode env.
  - returns 15 min when `OPENCODE_PROJECT_DIR` OR `OPENCODE_DEBUG` is present.
  - `CONTEXT_MODE_IDLE_TIMEOUT_MS` env still wins over auto-detection in
    both directions (explicit `0` disables even on OpenCode; explicit
    positive value enables on any host).
  - malformed env still falls back to the host-aware default.
- `startupSiblingSweep()`:
  - default-off for `{}` (returns empty report immediately).
  - auto-on when `OPENCODE_PROJECT_DIR` / `OPENCODE_DEBUG` is set.
  - `CONTEXT_MODE_STARTUP_SWEEP=1` still force-enables on any host.
  - `CONTEXT_MODE_STARTUP_SWEEP=0|false` still force-disables.

New assertions added to the existing domain files (`tests/lifecycle.test.ts`,
`tests/sibling-mcp.test.ts`) — no new test files.

## Test plan

- [x] `npm run build` — clean (bundles regenerated but reverted before
      commit per repo convention)
- [x] `npm test` — 145 / 148 test files pass, 3347 / 3394 tests pass
      (47 skipped). Baseline on clean `origin/next` is 144 / 148 files
      pass with 4 vitest-pool worker timeouts on
      `statusline-sqlite.test.ts` and `kiro-hooks.test.ts`; after this
      patch the same two flakes remain (2 errors total) — they exist on
      baseline and are unchanged by this PR.
- [x] Targeted runs verifying the host-detection behavior:
      `npx vitest run tests/lifecycle.test.ts tests/sibling-mcp.test.ts tests/lifecycle-cross-adapter.test.ts tests/server-stdin-eof-exit.test.ts tests/adapters/pi-bridge-parent-death.test.ts`
      -> 82 pass / 1 skipped / 0 fail.

## Notes

- No structural code from #568 is removed — only the two default values
  are flipped and gated on the OpenCode env markers. This keeps the diff
  to 4 files / +82 / -39 lines and leaves room for the future OpenCode
  rewrite to land on top without rebuilding the idle/sweep machinery.
- No doc surface enumerates the previous 15-minute default; nothing to
  update in `README.md` / `docs/platform-support.md`. The env vars
  `CONTEXT_MODE_IDLE_TIMEOUT_MS` and `CONTEXT_MODE_STARTUP_SWEEP` are
  not currently listed in any public env-var table.

### DIFF
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -52,23 +52,42 @@ export interface LifecycleGuardHandle {
   stop: () => void;
 }
 
+/** OpenCode/Kilo idle-shutdown threshold — 15 min, same value as their shipped configs. */
+const OPENCODE_IDLE_TIMEOUT_MS = 15 * 60 * 1000;
+
 /**
- * Resolve the idle-shutdown threshold (#565).
+ * Detect a host that originated the #565 child-accumulation case.
  *
- * Idle shutdown is OFF by default (#592) because most hosts (Claude
- * Code, Codex, editor MCP clients) keep registered tool handles after a
- * clean MCP child exit and do NOT transparently respawn on the next call.
- * The global 15 min default introduced in #568 solved OpenCode's child
- * accumulation, but stranded ctx_* tools in Claude Code/Codex-style
- * hosts once the MCP server exited cleanly while the editor stayed alive.
+ * #595 ships `CONTEXT_MODE_IDLE_TIMEOUT_MS=900000` in `configs/opencode/opencode.json`
+ * and `configs/kilo/kilo.json` so fresh installs are protected. Existing
+ * users with a customised / pinned MCP config will NOT pick up that env
+ * unless they re-pull the shipped config. We additionally key off env
+ * vars OpenCode itself always injects into spawned MCP children
+ * (`OPENCODE_PROJECT_DIR`, `OPENCODE_DEBUG`) so the protection engages
+ * for them too — no user action required.
  *
- * Hosts that are known to benefit from idle shutdown MUST opt in via
- * CONTEXT_MODE_IDLE_TIMEOUT_MS in their MCP config. Today that is
- * OpenCode/KiloCode (their configs set 900000 = 15 min). Users and test
- * harnesses can also opt in explicitly with any positive integer.
+ * Other hosts (Claude Code, Codex, editor MCP clients) stay default-off
+ * because they cannot transparently respawn the MCP child on next call
+ * (#592 root cause).
+ */
+export function hostNeedsIdleShutdown(
+  env: NodeJS.ProcessEnv = process.env,
+): boolean {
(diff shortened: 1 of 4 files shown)


ousamabenyounes:
Rebased on `next` (post-#595) and slimmed to the two incremental additions #595 leaves open:

1. Runtime `OPENCODE_PROJECT_DIR` / `OPENCODE_DEBUG` auto-detect so existing OpenCode users with a customised/pinned `opencode.json` still get the protection without re-pulling the shipped config.
2. `startupSiblingSweep` gated on the same OpenCode check — same #592 root cause exists on the sweep side (reaping a sibling on a host that cannot respawn its MCP client surfaces as `MCP server has exited` too).

No structural change to the #568 machinery. 4 files, +94/-28. Full suite green, net +3 test files pass vs baseline `next`.

Happy to close this if you'd rather land these two in a fresh PR — let me know.

Issue #598, opened 2026-05-17
vLLM prefix caching invalidated by behavioral_directive in system prompt (Pi adapter)


AntonProkopyev:
## vLLM prefix caching invalidated by behavioral_directive in system prompt (Pi adapter)

**Component:** `build/adapters/pi/extension.js` → `before_agent_start` hook

### Problem

`behavioral_directive` injected by `buildAutoInjection()` into `<session_state>` changes the system prompt on every turn, invalidating vLLM's prefix cache.

```xml
<session_state source="compaction">
<behavioral_directive>[user's latest message]</behavioral_directive>
<session_mode>[mode]</session_mode>
</session_state>
```

This is appended to `systemPrompt` via `parts.join("\n\n")`. Since `behavioral_directive` mirrors the user's latest message, the system prompt hash changes every turn → full re-prefill → minute-long wait.

### Fix

Extract `behavioral_directive` from `memoryContext` and append it at the END of the system prompt:

```javascript
const bdMatch = memoryContext.match(/(<behavioral_directive>\n[^<]*\n<\/behavioral_directive>)/);
if (bdMatch) {
    behavioralDirective = bdMatch[1];
    memoryContext = memoryContext.replace(bdMatch[1], '');
    if (memoryContext.match(/^<session_state[^>]*>\s*<\/session_state>\s*$/)) {
        memoryContext = '';
    }
}
// later: if (behavioralDirective) parts.push(behavioralDirective);
```

Result: stable prefix is identical across all requests. Only `behavioral_directive` at the very end changes → vLLM caches the prefix, only re-prefills the tail.

mksglu:
Let's create a PR instead of issue! @AntonProkopyev Thx!

PR #599, opened 2026-05-17
fix(pi): move behavioral_directive to end of system prompt

AntonProkopyev:
vLLM prefix caching invalidated by behavioral_directive in the middle of system prompt. Move it to the end so the stable prefix stays intact.

**Problem:** behavioral_directive injected into session_state changes the system prompt on every turn, invalidating vLLM prefix cache.

**Fix:** Extract behavioral_directive from memoryContext and append it at the end of the system prompt.

Ref: #598

### DIFF
--- a/stats.json
+++ b/stats.json
@@ -1,7 +1,7 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "195.1k+",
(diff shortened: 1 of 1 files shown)


mksglu:
The branch should be next

Issue #531, opened 2026-05-11
`.mcp.json` ships `./start.mjs` (relative) — Claude Code spawns from session CWD, MODULE_NOT_FOUND


Aragorn2046:
## Summary

The repo-root `.mcp.json` declares the MCP server as:

```json
{
  "mcpServers": {
    "context-mode": {
      "command": "node",
      "args": ["./start.mjs"]
    }
  }
}
```

That relative path has no companion `cwd` field. When Claude Code spawns the MCP server, the child inherits Claude Code's session CWD (whichever directory the user ran `claude` from), not the plugin install directory. Result: `node` resolves `./start.mjs` against the user's project dir, the file isn't there, and the spawn fails with `MODULE_NOT_FOUND` / `ENOENT` before initialize.

This is the same **class** of bug as #411 and #523 — but the missing-fix is in the third manifest, `.mcp.json` itself.

## Symptom

`claude mcp list` reports:
```
plugin:context-mode:context-mode: node ./start.mjs - ✗ Failed to connect
```

All `ctx_*` tools become unavailable; users land in "is the plugin broken?" territory.

## Environment

- OS: macOS 26 (Darwin 25.4.0)
- Node: v22 (system + bun runtime detected)
- Claude Code: current (Opus 4.7 session)
- context-mode: 1.0.111 (also confirmed in `main` at 1.0.121)
- Trigger: launching `claude` from any directory other than `~/.claude/plugins/cache/context-mode/context-mode/<v>/`

## Root cause

The two sibling manifests disagree:

| File | `mcpServers.context-mode.args[0]` |
|---|---|
| `.claude-plugin/plugin.json` | `${CLAUDE_PLUGIN_ROOT}/start.mjs` ✅ (correct since #411) |
| `.mcp.json` | `./start.mjs` ❌ |

Claude Code reads `.mcp.json` and launches with that command — confirmed by the literal `node ./start.mjs` shown in `claude mcp list` output (vs. the resolved absolute path that would appear if it had read `plugin.json > mcpServers`).

`start.mjs` itself does `process.chdir(__dirname)` on line 10, but that never runs because Node never finds the file.

## Repro

1. Install context-mode 1.0.111 (or any current version) on Claude Code
2. Launch `claude` from any directory other than the plugin cache (e.g. `~/workspace`)
3. `claude mcp list` → context-mode shows `✗ Failed to connect`
[...]

Issue #523, opened 2026-05-11
Upgrade leaves stale Temp directory path in .claude-plugin/plugin.json mcpServers.args — MCP fails to spawn with ENOENT (related to #411)


zkzi3254:
## Summary

After `/context-mode:ctx-upgrade`, the plugin manifest at `<HOME>/.claude/plugins/cache/context-mode/context-mode/<version>/.claude-plugin/plugin.json` contains a **stale temp-directory path** in `mcpServers.<name>.args`, causing the MCP server to fail spawning with `ENOENT` on every subsequent Claude Code session start.

This appears to be the same **class** of bug as #411, but affecting a different manifest file (`.claude-plugin/plugin.json` instead of `.mcp.json`). The fix landed for #411 evidently did not cover this second manifest.

## Environment

- OS: Windows 11
- Node: 22.x (Volta-managed)
- Host: Claude Code (running as VS Code extension)
- context-mode: v1.0.118 (upgraded from v1.0.103)
- Trigger: `/context-mode:ctx-upgrade` slash command

## Root cause

After upgrade, `plugin.json` `mcpServers` looks like:

```json
"mcpServers": {
  "context-mode": {
    "command": "<NODE_PATH>",
    "args": [
      "<HOME>/AppData/Local/Temp/context-mode-upgrade-<EPOCH>/start.mjs"
    ]
  }
}
```

The path `Temp/context-mode-upgrade-<EPOCH>/start.mjs` is the **upgrade script's own working directory** — in my case `<EPOCH>` was `1778466358412` which decodes to `2026-05-11T02:25:58Z`, the exact moment of my upgrade. That directory is cleaned up when the upgrade completes, so Claude Code's subsequent `spawn` attempts hit `ENOENT` and the MCP initialize handshake never completes.

## Reproduction

1. Have any prior context-mode version installed (e.g. v1.0.103)
2. Run `/context-mode:ctx-upgrade`
3. Wait for the upgrade to complete (the in-memory MCP server is still alive at this point, so doctor passes — the breakage only manifests on the next restart)
4. **Restart the Claude Code session**
5. Observe: context-mode MCP tools become unavailable. `/reload-plugins` does not recover (the MCP child process exits with ENOENT before initialize completes)
[...]

Issue #411, opened 2026-05-04
Cached .mcp.json hardcodes path to PREVIOUS version start.mjs — breaks reconnect after cleanup


mateuszruszkowski:
## Environment

- **OS**: Linux (Fedora 44)
- **Node.js**: v24+
- **Claude Code**: 1.0.107
- **context-mode**: 1.0.103 → 1.0.107 (auto-upgraded mid-session)

## Symptom

After a session-mid auto-upgrade, `/mcp` reports:

```
Failed to reconnect to plugin:context-mode:context-mode.
```

## Root cause

The cached `.mcp.json` at `~/.claude/plugins/cache/context-mode/context-mode/<VERSION>/.mcp.json` contains a **hardcoded absolute path that points to the PREVIOUS version's `start.mjs`**, not the current one and not `${CLAUDE_PLUGIN_ROOT}`.

Observed on my machine after upgrade chain `1.0.99 → 1.0.103 → 1.0.107`:

| Cache dir | `.mcp.json` `args[0]` | Problem |
|---|---|---|
| `1.0.103/.mcp.json` | `…/1.0.99/start.mjs` | `1.0.99/` already cleaned up → path broken |
| `1.0.107/.mcp.json` | `…/1.0.103/start.mjs` | works only until `1.0.103/` is cleaned |

Compare with marketplace template which is correct:

```json
// ~/.claude/plugins/marketplaces/context-mode/.mcp.json
{ "mcpServers": { "context-mode": { "command": "node",
    "args": ["${CLAUDE_PLUGIN_ROOT}/start.mjs"] } } }
```

So somewhere in the install/upgrade pipeline, `${CLAUDE_PLUGIN_ROOT}` is being resolved at install time and **the resolved value is the path of the previously-installed version**, not the version being installed.

## Repro

1. Install context-mode at version N
2. Auto-upgrade to N+1 mid-session
3. Inspect `~/.claude/plugins/cache/context-mode/context-mode/<N+1>/.mcp.json`
4. Observe `args[0]` points to `<N>/start.mjs`, not `<N+1>/start.mjs` and not `${CLAUDE_PLUGIN_ROOT}/start.mjs`
5. Wait until age-gated cleanup (#181 fix) removes `<N>/`
6. Try `/mcp` reconnect → fails

## Proposed fix

In the install/upgrade script, write the cached `.mcp.json` **verbatim** with the literal `${CLAUDE_PLUGIN_ROOT}` placeholder:

```json
{
[...]

PR #616, opened 2026-05-18
fix(install): ship .mcp.json + heal-create when missing (#609)

### LINKED ISSUE #609: .mcp.json missing from new version dir after auto-update — MCP server boots from stale dir, fails MODULE_NOT_FOUND
## Summary

After Claude Code's native plugin manager auto-updates context-mode, `.mcp.json` is only present in the **previous** version directory, not the new one. Claude Code finds the stale `.mcp.json` and launches the MCP server with `CLAUDE_PLUGIN_ROOT` pointing at the old version — where `start.mjs` does not exist — producing `MODULE_NOT_FOUND` on every boot.

This is the same class of bug as #411 (same file, different variant) and #604 (`hooks.json`). The existing `context-mode-cache-heal.mjs` fix does **not** catch this case.

## Environment

- OS: Linux (WSL2, kernel 6.6.87.2-microsoft-standard-WSL2)
- Node: system node
- Host: Claude Code
- context-mode: 1.0.135 → 1.0.136 (auto-updated by Claude Code plugin manager, mid-use)

## Symptom

Every hook fires `Failed with non-blocking status code: node:internal/modules/cjs/loader:1386`. MCP tools unavailable. `/plugin` returns `(no content)`.

```
Error: Cannot find module '/home/<user>/.claude/plugins/cache/context-mode/context-mode/1.0.135/start.mjs'
    at Function._resolveFilename (node:internal/modules/cjs/loader:1383:15)
    ...
```

## Root Cause

After auto-update:

| Path | `.mcp.json` | `start.mjs` |
|------|------------|------------|
| `cache/.../1.0.135/` | ✓ present | ✗ missing |
| `cache/.../1.0.136/` | ✗ missing | ✓ present |

Claude Code scans the cache tree, finds `1.0.135/.mcp.json`, and launches:

```
node "${CLAUDE_PLUGIN_ROOT}/start.mjs"   # CLAUDE_PLUGIN_ROOT = 1.0.135
```

`1.0.135/start.mjs` doesn't exist → MODULE_NOT_FOUND.

`context-mode-cache-heal.mjs` does not help: its trigger is `!existsSync(installPath)`. Since `installPath` (`1.0.136/`) **does** exist, the healer exits immediately without touching anything.

## Why #411's fix doesn't cover this

The cache-heal script only fixes a missing **`installPath` symlink** (the case where the entire previous version dir was deleted and `installed_plugins.json` pointed nowhere). It has no logic to detect or repair a missing `.mcp.json` in the current version dir.

## Reproduction

1. Boot once on `N` so `N/.mcp.json` is written/present.
2. Let Claude Code auto-update to `N+1` (do **not** use `/ctx-upgrade`).
3. `N+1/.mcp.json` will be absent; `N/.mcp.json` will remain.
4. Start a new session — MCP server fails to boot.

## Workaround

```bash
cp ~/.claude/plugins/cache/context-mode/context-mode/1.0.135/.mcp.json \
   ~/.claude/plugins/cache/context-mode/context-mode/1.0.136/.mcp.json
```

Then restart the session. Or run `/context-mode:ctx-upgrade` which rebuilds plugin wiring from scratch.

## Suggested Fix

Same three options as #604:

1. **`cache-heal.mjs`**: add a check — if current version dir exists but is missing `.mcp.json`, copy/regenerate it from the previous version dir or from the committed source.
2. **`normalize-hooks.mjs` / auto-update pipeline**: ensure `.mcp.json` is always written fresh (with placeholder intact) into the new version dir, not left absent.
3. **Upstream (Claude Code)**: atomically migrate or re-extract all per-version metadata files on update (anthropics/claude-code#46915).

## References

- #411 — same file, different variant (deleted previous dir vs. missing file in new dir)
- #604 — `hooks.json` variant of the same root cause
- #523 — `.claude-plugin/plugin.json` variant
- #528 — fixed `/ctx-upgrade` tmpdir variant

ousamabenyounes:
## Summary

Closes #609. Claude Code's native plugin auto-update doesn't ship `.mcp.json` into the new version directory and leaves a stale one behind in the previous version. CC's cache scanner picks the stale file, tries to launch the deleted previous-version `start.mjs` → `MODULE_NOT_FOUND` on every MCP boot, `/plugin` returns no content, every hook fails.

The fix that breaks the deadlock: ship `.mcp.json` (placeholder form) in every npm tarball so every cache version dir holds its own current copy. Plus a defensive heal that creates `.mcp.json` when missing (covers `/ctx-upgrade` tmpdir + pipelines that don't run prepack).

## Resolving #609

Claude Code's plugin scanner reads `.mcp.json` from the cache version dir when present, and only falls back to `.claude-plugin/plugin.json` when it isn't. Since `.mcp.json` isn't currently shipped in the npm tarball, fresh version dirs hold none — but legacy install chains can leave a stale one in the previous version dir from before #531 landed. The scanner picks the stale file, expands `${CLAUDE_PLUGIN_ROOT}` to the previous version's path, and tries to spawn its `start.mjs` — which the auto-update already removed.

Shipping `.mcp.json` (placeholder form) inside the tarball guarantees the scanner always finds the current version's file in the current version's dir. The `${CLAUDE_PLUGIN_ROOT}/start.mjs` form makes the path correct regardless of which version dir CC picks it from.

Contributor note: local dev (cloning the repo and running Claude Code against it) may surface CC's `Missing environment variable: CLAUDE_PLUGIN_ROOT` warning when `CLAUDE_PLUGIN_ROOT` isn't set. The warning is harmless and can be silenced via `export CLAUDE_PLUGIN_ROOT=$PWD` in the dev shell. The relative-path form (`./start.mjs`) is no longer supported in the shipped file — the placeholder is universal across contributor and end-user contexts. Documented in the `.gitignore` comment.

## Changes

- **`.gitignore`** — remove `.mcp.json` exclusion. Replace the comment with a `#609 — shipped placeholder form is universal` block.
- **`.mcp.json`** (new, committed) — placeholder form matching `.mcp.json.example` verbatim.
- **`package.json`** — add `.mcp.json` to `files` array so it ships in npm tarball.
- **`scripts/heal-installed-plugins.mjs`** — `healMcpJsonArgs` now creates a canonical `.mcp.json` when missing instead of skipping with `no-mcp-json`. Defensive layer for paths that don't run prepack.
- **`tests/core/cli.test.ts`** — assertion now requires `.mcp.json` to be in `files[]` and to use the placeholder form. Content shape pinned (drift guard).
- **`tests/util/heal-installed-plugins.test.ts`** — `healMcpJsonArgs` create-when-missing assertion replaces the previous skip assertion.
- **`tests/util/postinstall-heal.test.ts`** — `buildFakeHome` seeds `.mcp.json` so the heal-create branch is a no-op for tests that count stderr heal output.

The asymmetric-drift assert (`scripts/assert-asymmetric-drift.mjs` + its vitest sibling) already covered `.mcp.json` content when present — that drift guard is now unconditional since `.mcp.json` always exists in repo.

## Test plan

- [x] `npm run build` — assert-asymmetric-drift passes (`.mcp.json`, `.mcp.json.example`, `.claude-plugin/plugin.json` all pin args[0] to `${CLAUDE_PLUGIN_ROOT}/start.mjs`)
- [x] `npm test` — 145/145 test files pass, 3317/3360 tests pass (43 skipped, 0 fail). Same baseline as `origin/next` — no regression introduced.
- [x] `npm pack --dry-run` — `.mcp.json (129B)` included in tarball.

## Risk / compatibility

- Existing v1.0.137 users will still see #609 once when they upgrade to v1.0.138 (the v1.0.137 → v1.0.138 transition still goes through CC's auto-update which won't include `.mcp.json` from the v1.0.137 tarball). The next transition (v1.0.138 → v1.0.139+) will be clean because v1.0.138 ships `.mcp.json` → CC's auto-update preserves it.
- Defensive heal-create covers users who run `/ctx-upgrade` or `npm install -g context-mode` — both run paths that invoke `healMcpJsonArgs` via postinstall.
- No new env var, no schema migration, no CLI command change.

Refs: #609, #531, #411, #523, #253.

### DIFF
--- a/.mcp.json
+++ b/.mcp.json
@@ -1,0 +1,8 @@
+{
+  "mcpServers": {
+    "context-mode": {
+      "command": "node",
+      "args": ["${CLAUDE_PLUGIN_ROOT}/start.mjs"]
+    }
+  }
+}
--- a/.gitignore
+++ b/.gitignore
@@ -20,16 +20,16 @@ AGENTS.md
 context-mode-guidance-*/
 .claude/worktrees/
 
-# .mcp.json is local-only — purpose splits between contributor dev
-# (relative paths work because cwd = repo root) and end-user marketplace
-# install (placeholder ${CLAUDE_PLUGIN_ROOT}/start.mjs is required).
-# These two forms have flip-flopped 3 times in history (#253 #531 etc),
-# causing every fresh install to silently break. End users get MCP via
-# .claude-plugin/plugin.json (always placeholder) and /ctx-upgrade's
-# cli.ts write (always placeholder). The .mcp.json at repo root only
-# helps contributors during local dev. Untracked → can't regress.
-# Copy .mcp.json.example to .mcp.json for local dev.
-.mcp.json
+# .mcp.json is now tracked + shipped (#609 — Claude Code's native plugin
+# auto-update doesn't include .mcp.json from the previous version, leaves it
+# stranded, and the cache scanner picks up the stale file → MODULE_NOT_FOUND).
+# Canonical placeholder form (${CLAUDE_PLUGIN_ROOT}/start.mjs) is committed
+# at repo root so npm pack includes it in every version's cache dir. Drift is
+# guarded by asymmetric-drift assert + a unit test asserting the exact shape.
+# Contributors running local dev may see Claude Code's "Missing environment
+# variable: CLAUDE_PLUGIN_ROOT" warning — harmless, can be silenced by setting
+# CLAUDE_PLUGIN_ROOT=$PWD in the dev shell. The contributor-relative shape
+# (./start.mjs) is no longer supported — use the placeholder form universally.
 .claude-worktrees/
 .vibetree/
 .cw/
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
     ".codex-plugin",
     ".openclaw-plugin",
     "openclaw.plugin.json",
+    ".mcp.json",
     "start.mjs",
     "scripts/postinstall.mjs",
     "scripts/heal-better-sqlite3.mjs",
--- a/scripts/heal-installed-plugins.mjs
+++ b/scripts/heal-installed-plugins.mjs
@@ -341,7 +341,28 @@ export function healMcpJsonArgs({ pluginRoot, pluginCacheRoot, pluginKey }) {
   // `.mcp.json` lives at pluginRoot/.mcp.json (flat), NOT under .claude-plugin/.
   const mcpJsonPath = resolve(pluginRoot, ".mcp.json");
   if (!existsSync(mcpJsonPath)) {
-    return { healed: [], skipped: "no-mcp-json" };
+    // #609: Claude Code's native plugin auto-update doesn't ship a fresh
+    // `.mcp.json` into the new version dir and leaves a stale one behind in
+    // the previous version dir. CC's cache scanner then picks the stale file
+    // and tries to boot from the deleted previous-version `start.mjs`. The
+    // primary fix is shipping `.mcp.json` via the npm tarball; this branch is
+    // the defensive heal for the `/ctx-upgrade` tmpdir path + any other
+    // pipeline that lands files without invoking the prepack.
+    const ourServerName = pluginKey.split("@")[0];
+    const canonical = {
+      mcpServers: {
+        [ourServerName]: {
+          command: "node",
+          args: [PLACEHOLDER_ARG],
+        },
+      },
+    };
+    try {
+      writeFileSync(mcpJsonPath, JSON.stringify(canonical, null, 2) + "\n", "utf-8");
+      return { healed: ["mcp-json-created"] };
+    } catch (err) {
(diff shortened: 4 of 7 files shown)


.mcp.json
mksglu, line 1: We don't want that.

mksglu (changes requested):

ousamabenyounes:
Closing per @mksglu's review and the historical pattern.

After running the git archaeology, this PR re-introduces a file that has been added and removed multiple times for the same flip-flop reason that #531 closed architecturally. The proposed defensive heal-create on the function side is also insufficient on its own — Claude Code's native plugin auto-update path never invokes our postinstall or our `start.mjs` of the new version (chicken-and-egg: the new MCP child never boots), so any heal we add from inside context-mode cannot reach that path.

The root cause of #609 is upstream in Claude Code's plugin scanner / auto-update behavior, not in context-mode. Will file an upstream issue at `anthropics/claude-code` and post a manual workaround on #609 so affected users can recover via `/context-mode:ctx-upgrade` or a one-line `cp` until the upstream fix lands.

Sorry for the noise on this one.

Issue #482, opened 2026-05-08
[Feature]: Make normal-response communication style configurable


makoMakoGo:
## Feature description

Please add a documented switch for the response-style instructions that context-mode injects into the agent's final / normal user-facing output.

In `1.0.111`, context-mode injects this style rule:

```text
Terse like caveman. Technical substance exact. Only fluff die.
Use fragments when clear. Short synonyms...
```

It appears in `hooks/routing-block.mjs` under `<communication_style>`, and similar wording is repeated in MCP tool descriptions in `src/server.ts`:

```text
When reporting results — terse like caveman. Technical substance exact. Only fluff die...
```

The issue is that this does not only reduce raw tool output. It also pushes the model's final response to the user toward a specific terse/caveman style.

Context-mode is useful because it keeps large tool output out of context by running commands and file processing through sandboxed tools. That is separate from controlling how the model writes its final answer to the user. I want to use context-mode for tool-output filtering and token savings, but I do not want it to impose a global final-response style.

## Use case

When doing coding, debugging, migration, architecture, or security work, the final answer may need to include enough detail for the user to understand the result:

- assumptions
- caveats
- verification evidence
- failure modes
- security warnings
- what changed and why

A global rule like `Terse like caveman` can pressure the model to shorten the wrong thing: the final explanation / decision / warning, not just noisy tool output.

There is related evidence from OpenCode / Kimi that aggressive brevity instructions can be non-neutral for coding models:

- OpenCode issue: https://github.com/anomalyco/opencode/issues/20258
- OpenCode PR: https://github.com/anomalyco/opencode/pull/20259

Moonshot AI reported that OpenCode's default prompt degraded `kimi-k2.5` coding/reasoning benchmark performance. One cited cause was aggressive brevity constraints such as:

- `minimize output tokens as much as possible`
- `MUST answer concisely with fewer than 4 lines`
- `One word answers are best`

The wording is not identical, but the practical concern is similar: a framework/tool prompt should not force a global terse final-response style unless the user opts into it.

## Proposed solution

Add a config switch for the final-response communication style.

Example:

```json
{
  "context-mode": {
    "communicationStyle": "default"
  }
}
```

or:

```json
{
  "context-mode": {
    "injectCommunicationStyle": false
  }
}
```

Expected behavior when disabled:

1. Do not inject the `Terse like caveman...` communication-style block into SessionStart / routing instructions.
[...]

PR #617, opened 2026-05-18
feat: add runtime storage override

TyceHerrman:
## Problem

context-mode can run as an MCP server inside Codex Desktop where tool registration succeeds, but the first storage write can fail under a workspace-write sandbox, for example:

```text
EPERM: operation not permitted, mkdir '/Users/.../.codex/context-mode/content'
```

## Behavior Change

Before this PR, an unwritable platform-default storage path could throw from the first storage-writing `ctx_*` tool call, leaving the client with a transport-level failure or no useful tool result.

After this PR, the tool call returns an MCP tool result with `isError: true` and actionable text, for example:

```text
context-mode content directory is not writable: /Users/.../.codex/context-mode/content
Set CONTEXT_MODE_DIR to a writable absolute path.
```

## Change

Adds one public runtime override: `CONTEXT_MODE_DIR`.

- Derives sessions and stats from `<root>/sessions`.
- Derives indexed content from `<root>/content`.
- Treats empty or whitespace-only `CONTEXT_MODE_DIR` as unset, matching existing env-var conventions.
- Surfaces ignored empty overrides in `ctx_doctor` and storage errors when the adapter default path still fails.
- Rejects non-empty relative `CONTEXT_MODE_DIR` values with an actionable error.
- Converts storage setup failures into MCP `isError: true` tool results instead of allowing transport-breaking exceptions.
- Reports resolved storage roots and source markers in `ctx_doctor`.
- Keeps `CONTEXT_MODE_SESSION_DIR` as a deprecated statusline-only fallback.
- Memoizes writable-directory checks per resolved storage path.
- Composes storage-error handling into the existing `server.registerTool` wrapper.
- Consolidates shared default session-dir resolution for server, hooks, and statusline.

## Bundle policy

The storage resolver is folded into existing `src/session/db.ts`. Packaged hooks and statusline use the existing `hooks/session-db.bundle.mjs`, so this PR no longer adds a new generated `hooks/storage-paths.bundle.mjs` file or new package bundle-script entries. `server.bundle.mjs` and `cli.bundle.mjs` are left to the existing bundle workflow.

## Tests

- `npx vitest run tests/core/server.test.ts tests/statusline.test.ts tests/statusline-sqlite.test.ts -t "storage path resolution|ctx_doctor reports storage|ctx_index returns an actionable storage error|registerEmptyToolsListHandler|legacy CONTEXT_MODE_SESSION_DIR|CONTEXT_MODE_DIR" --reporter=dot`
- `npm run typecheck`
- `npm run build`
- `git diff --check`


sebastianbreguel:
Hi @mksglu @TyceHerrman — reviewing this from a scope/minimalism lens.

Real problem, right shape, but the surface area is wider than the bug requires.

## Scope Gate

| Gate | Result | Note |
|---|---|---|
| Real, specific problem | ✅ | Codex Desktop `EPERM ... mkdir '/.codex/context-mode/content'` — reproducible, transport-closing |
| Simpler than alternatives | ✅ | One env var, no config files, no platform-detection rework |
| Lightweight | ✅ | Startup-only `process.env` reads, no runtime cost |
| Sensible defaults | ✅ | Unset → adapter defaults unchanged |
| Tight, defensible scope | ⚠️ | Four env vars for a problem the root override alone solves |
| Safe by default | ✅ | Actionable errors + `doctor` writability check |
| Polished | ✅ | Statusline + session-helper paths honored |
| Maintainable long-term | ⚠️ | Every public env var is a forever commitment |

## Asks before merge

1. **Cut three env vars.** Keep `CONTEXT_MODE_DIR`. Drop `CONTEXT_MODE_SESSION_DIR`, `CONTEXT_MODE_CONTENT_DIR`, `CONTEXT_MODE_STATS_DIR`. The Codex case needs only the root override — the splits solve a problem no one has reported. Add them when someone actually asks. *"If you haven't used it in the past year, you probably won't use it in the future."*

2. **Rebase against `next`.** PR is currently `CONFLICTING`.

3. **Confirm `tests/storage-paths.test.ts`.** House rule here is no new test files without approval — new module → new test file is defensible, but worth an explicit yes rather than skipping the check. Otherwise fold cases into `tests/core/server.test.ts`.

## Verdict

Merge with changes. Trim to one env var + rebase → ship it. The core idea is right; the API surface is just too wide for the bug it fixes.

One line: solid problem, right shape, three env vars too many — cut to one and ship.

ousamabenyounes (commented):
Approach is correct — the bug is real, `CONTEXT_MODE_DIR` is the right shape, and @sebastianbreguel already covered the scope angle. Comments below focus on implementation drift and a stale base.

### Blockers

1. **#637 regression risk (rebase symptom).** The branch is missing `registerEmptyToolsListHandler` at `tests/core/server.test.ts:4828–4884`, present on `next` (added to close #637 — suppressed legacy MCP child returning JSON-RPC `-32601` on `tools/list`). Rebase resolves this mechanically, but explicit verification that the test survives the rebase is required before merge.

2. **Override logic duplicated across three files — drift already observable.** Same env-var precedence reimplemented in:
   - `src/storage-paths.ts` (canonical)
   - `hooks/session-helpers.mjs:268–276`
   - `bin/statusline.mjs:101–106`

   The header of `session-helpers.mjs` (lines 9–14) explicitly forbids exactly this pattern: *"PATH / HASH HELPERS ARE BOUND, NOT REIMPLEMENTED … the same drift that produced rounds 5 and 6 of case-fold fixes."* The divergence is already visible: `storage-paths.ts:envDir()` rejects whitespace via `.trim()`, `session-helpers.mjs:resolveSessionDir` only rejects empty string — `CONTEXT_MODE_DIR="  "` is accepted by hooks and rejected by the server. Right fix: expose `resolveSessionStorageDir` via a bundle (same pattern as `session-db.bundle.mjs`) and import from hooks + statusline.

### Major

3. **No tilde expansion on the new env vars.** `path.resolve("~/foo")` does not expand `~` — it becomes `<cwd>/~/foo`. `session-helpers.resolveConfigDir` (lines 193–194) already expands `~` for `CLAUDE_CONFIG_DIR` / `CODEX_HOME` / `GEMINI_CLI_HOME`. The new `CONTEXT_MODE_*` vars don't — `CONTEXT_MODE_DIR=~/cm` silently misroutes. Expand in `envDir()` or extend the error message to require an absolute path.

4. **No EPERM end-to-end test.** PR motivation is "EPERM closes transport → return actionable MCP error." `tests/storage-paths.test.ts` covers resolution + error formatting; nothing exercises the catch path at `src/server.ts:618–624` end-to-end. A test that points `CONTEXT_MODE_DIR` at a `chmod 0500` tmp dir, invokes `ctx_index`, and asserts the result has `isError: true` with the formatted override hint would lock the contract that justifies this PR.

5. **Split-only override** (only relevant if splits survive @sebastianbreguel's scope cut): the current model — each split overrides independently and content falls back via `join(dirname(session.path), "content")` — is too fragile for a forever public API. With `CONTEXT_MODE_SESSION_DIR=/foo` and no other override, content lands at `/content`. Two safer designs:
   - Require all three splits together when any split is set, **or**
   - Forbid mixing base + splits — either `CONTEXT_MODE_DIR` alone, or all three splits together, never partial.

### Minor

6. **Doc gap.** `CONTEXT_MODE_DIR` (and the splits) are not documented in `README.md` or surfaced in `--help` / `doctor`. The error message is a good last-resort fallback, but a user tuning their Codex Desktop setup before hitting the crash won't find the env var. Add to README env-var section and to `--help`.

7. **Double resolution in `getStatsFilePath`** (`src/server.ts:853`). `resolveStatsStorageDir(getSessionDir)` passes the wrapped `getSessionDir` (which itself runs `resolveSessionStorageDir` + `ensureWritableStorageDir`). When stats override is absent, the full session pipeline re-runs including a redundant `mkdirSync`/`accessSync`. Pass `getDefaultSessionDir` — the raw callback the parameter was designed for.

8. **`StorageDirectoryError` constructor** (`storage-paths.ts:42`): `(this as Error & { cause?: unknown }).cause = cause` — `engines.node = ">=22.5.0"`, so `super(message, { cause })` is supported natively.

9. **Constant extraction**: `"sessions"` and `"content"` recur as literals across `storage-paths.ts` (lines 51, 70, 75, 79, 85, 92). Module-local `SUBDIR_SESSIONS` / `SUBDIR_CONTENT` would centralize.

### Nits

- `doctor()` prints the path table (`p.note`, cli.ts:379–386) *before* the per-line writability check, so paths flash green before any FAIL line. Cosmetic.
- Error message could append the resolved adapter-default path so users see what *would* have been used.

mksglu:
Hi @TyceHerrman! Your finding PRs is so valuable. Let's reach out to me via DM (e-mail, Linkedin, X) and come to our private engineering group.

tests/storage-paths.test.ts
mksglu, line 1: No need seperate file for that. Please use existing one.

package.json
mksglu, line 1: No need seperate line for that. For files, please use existing one.

src/storage-paths.ts
mksglu, line 1: No need seperate file for that. Please use existing one.

hooks/storage-paths.bundle.mjs
mksglu, line 1: No need seperate file for that. Please use existing one.

mksglu (changes requested):

mksglu:
Deep Staff Engineer + Architect review against [CONTRIBUTING.md](https://github.com/mksglu/context-mode/blob/next/CONTRIBUTING.md). Verdict: **REQUEST CHANGES**, keep this as a single PR and address the items below.

## Scope note

15 files touched (`+808 −341`) for what is fundamentally **1 new module + 3 call-site swaps + 1 README row + 1 test**. The maintainer prefers single-PR delivery here — do **not** split, just address the blockers below in place. A follow-up cleanup for the three reimplementations of the default-session-dir callback (Architect A2 below) can land separately.

## Staff Engineer findings (blockers)

### S1 — Monkey-patch collision with #623 suppression wrap (`src/server.ts`)

You add a second `server.registerTool = ((origReg) => function (...) {...})(server.registerTool.bind(server))` at L670-683. PR #623 (merged into `origin/next`) already wraps `server.registerTool` at L267-280 to emit the suppression diagnostic. Two un-coordinated monkey-patches on the same SDK method = order-dependent behavior, hard to reason about, and the SDK isn't designed for this.

**Fix**: compose into a single wrap, or factor a shared `wrapToolHandler` helper that both call sites use.

### S2 — Hot-path perf regression in `ensureWritableStorageDir`

`getSessionDir()`, `getStorePath()`, `getStatsFilePath()` each call `ensureWritableStorageDir(...)` which runs `mkdirSync({recursive:true}) + accessSync(W_OK)` on every invocation. These getters fire on every `ctx_*` tool call. The PR description claims "negligible startup cost" but `accessSync` is a syscall on every call.

**Fix**: memoize per `(adapter, dir)` tuple. Run the writability check once per boot, cache the result, surface the cached error from subsequent callers.

### S3 — Silent breaking change: `CONTEXT_MODE_SESSION_DIR` removal

`bin/statusline.mjs:102-107` drops the `CONTEXT_MODE_SESSION_DIR` env var handling with no deprecation path. Users who set that env var lose their override silently — statusline starts pointing at the wrong DB and they have no signal why.

**Fix**: keep the env var as a back-compat fallback, OR emit a one-shot stderr deprecation notice on first read so the migration path is visible.

### S4 — `tests/storage-paths.test.ts` creates a new test file (CONTRIBUTING line 282)

> Do NOT create new test files. Add your tests to the existing file that covers the same domain.

The storage-path behavior covered here belongs in `tests/core/server.test.ts` (already opened by this PR for the EPERM e2e). Fold the 111 lines in.

### S5 — Stale bundle files committed

`server.bundle.mjs` (`+108/-106`) and `cli.bundle.mjs` (`+188/-184`) are committed manually. Per the repo's `project_ci_bundles.md` memory: **CI rebuilds bundles on main push only — never commit them manually**. After rebase, drop these from the PR and let CI handle them.

### S6 — Long-term sustainability of `CONTEXT_MODE_DIR`

Env vars rot fast. Two common failure modes we want to actively prevent for this one:

1. **Dead-code drift** — someone refactors the resolver, the env var read silently disappears from the cascade, and no test fails because no test pins "env var X overrides default Y". Six months later nobody remembers the contract.
2. **Documentation drift** — the env var is documented in README but the actual precedence (env > config > default? env > config? env only when config missing?) lives only in the implementation, and the two diverge.

**Asks for this PR to make `CONTEXT_MODE_DIR` long-term safe:**

- [ ] **Behavior-pin test** in `tests/core/server.test.ts` that exercises the env var end-to-end — set `CONTEXT_MODE_DIR=/tmp/x`, call `getSessionDir()` / `getStorePath()` / `getStatsFilePath()`, assert each returns the expected subpath of `/tmp/x`. Not a source-grep — actual behavioral assertion that breaks loud if a future refactor drops the read.
- [ ] **Precedence test** — explicit env var beats default; unset env var falls back to default; whitespace/empty env value is treated as unset (don't accidentally write to `/`).
- [ ] **`ctx_doctor` surface** — add a one-line check that prints the resolved storage root, marked `(via CONTEXT_MODE_DIR)` when the env var is the source. Doctor visibility is the strongest anti-dead-code signal we have — users see the var being read, contributors see it as a tested checkpoint.
- [ ] **README env vars table** — if no canonical env-vars section exists, add one with `CONTEXT_MODE_DIR` as the first entry. Document: scope (session+content+stats dirs), precedence (overrides platform default), tilde expansion, fallback (unset → platform default), since version.

These four together make it materially harder for the env var to silently disappear in future refactors.

### S7 — Per-adapter README install-section examples for users who actually hit this

The headline reason this PR exists is sandboxed / read-only-home environments where the platform-default storage path isn't writable (Codex Desktop is the trigger). Generic env-var documentation in a central table (S6) tells engineers the var exists; **adapter install-section examples tell affected users how to fix their broken install in 30 seconds**.

For each adapter whose users are known to hit this class (Codex Desktop primarily; VS Code Copilot and JetBrains Copilot are likely candidates too if sandboxed), add a short **"Custom storage location"** subsection in that adapter's install block in `README.md`, immediately after Verify:

```
### Codex
...existing install steps...
**Custom storage location** (sandboxed installs, read-only $HOME, multi-user machines):

If you see EPERM / permission errors on first ctx_* call, redirect storage with:

\`\`\`bash
export CONTEXT_MODE_DIR=/path/to/writable/dir
# or persist in your shell config
\`\`\`
\`ctx_doctor\` will show the resolved root with \`(via CONTEXT_MODE_DIR)\` once set.
```

**Asks:**

- [ ] **Codex install section** in `README.md` — add the "Custom storage location" snippet (this is the user cohort that drove the PR — they should see the fix without leaving the install page)
- [ ] **VS Code Copilot install section** — same snippet if sandboxed-config users hit this
- [ ] **JetBrains Copilot install section** — same snippet if sandboxed-config users hit this
- [ ] **Test each example as written** — actually run `export CONTEXT_MODE_DIR=/tmp/ctx-test-readme && /context-mode:ctx-doctor` on a clean adapter install for each, confirm the output matches what the README claims. README snippets that don't reproduce are worse than no snippet.

If only Codex Desktop is empirically confirmed, ship the Codex example and skip the others — false coverage is worse than honest "we don't know yet". Note in the PR description which adapters were tested vs which are inferred.

## Architect findings

### A1 — Sprawling scope (keep single PR, just constrain blast radius)

The 15-file diff makes review and revert risky. Stripping bundles (S5) + folding the test file (S4) brings the surface down meaningfully. The split into three PRs (originally suggested) is **not required** — maintainer's call is single-PR delivery.

### A2 — Three reimplementations of default-session-dir callback (defer to follow-up)

After this PR:

- `src/server.ts` `getDefaultSessionDir` callback
- `bin/statusline.mjs` `resolveSessionDir`
- `hooks/session-helpers.mjs` `resolveSessionDir`

…each compute the same default. Should consolidate into `src/storage-paths.ts` as `getDefaultSessionDirFor(adapter)` and have all three call it. This **does not block this PR** — track as a follow-up.

### A3 — ADR 0001 (multi-writer contract) is preserved

No `acquireDbLock`, no `EXCLUSIVE` pragma, no lock-file logic added to the SQLite path. Good.

## CONTRIBUTING.md rubric

- L264 (TDD) — passes: test added before implementation per commits
- L282 (no new test files) — **violates** (`tests/storage-paths.test.ts` is new)
- L283-310 (test file table) — should go in `tests/core/server.test.ts`
- L348-363 (prose-style policy #482) — passes: no caveman/brevity language
- L233-243 (rebuild policy) — **violates** (bundles committed manually)
- L85 (multi-writer contract) — passes
- L313-318 (output quality) — partial: PR shows before/after EPERM error but not behavior diff
- L375-384 (PR template) — passes: branched from `next`, tests added

## Stale PR body

The PR description still references the original `CONTEXT_MODE_SESSION_DIR / CONTENT_DIR / STATS_DIR` split-env-var design. The code, README, and tests all ship **only `CONTEXT_MODE_DIR`** (per @sebastianbreguel's scope cut — good!). Update the PR body to remove the stale split-var section, or add an explicit "deferred per scope review" note.

## Acknowledgements

- @sebastianbreguel correctly trimmed scope to single env var.
- @ousamabenyounes flagged helper drift, stale base, tilde, double-resolution, doc gaps, constant extraction.

This review extends those with the monkey-patch collision (S1), hot-path perf regression (S2), the silent `CONTEXT_MODE_SESSION_DIR` removal (S3), the long-term env-var sustainability concern (S6), and the per-adapter install-section UX asks (S7) — independent and not yet raised.

## Checklist (re-review trigger)

- [ ] Rebase on current `origin/next` (HEAD `3e51e53`)
- [ ] S1: compose registerTool wraps
- [ ] S2: memoize `ensureWritableStorageDir`
- [ ] S3: restore `CONTEXT_MODE_SESSION_DIR` (back-compat or deprecation notice)
- [ ] S4: fold `tests/storage-paths.test.ts` into `tests/core/server.test.ts`
- [ ] S5: drop bundle files from PR
- [ ] S6: behavior-pin test + precedence test + `ctx_doctor` surface + README env vars row
- [ ] S7: Codex install section "Custom storage location" snippet (+ VS Code / JetBrains if confirmed) + each example actually tested before commit
- [ ] Update PR body to reflect final scope

Re-ping when ready — happy to re-review.

mksglu (approved):

PR #632, opened 2026-05-19
fix(opencode): convert zod v3 schemas to v4 for kilo platform compatibility

mikij:
## What / Why / How

KiloCode's runtime bundles Zod v4 internally, but plugin tool definitions may contain Zod v3 schemas. This causes crashes with "undefined is not an object" when accessing ._zod.def.

Added a utility that converts Zod v3 schema shapes to v4 equivalents specifically when platform === "kilo". OpenCode uses Zod 3 natively and receives the original shapes unchanged.

The conversion handles all common Zod types including objects, arrays, unions, optionals, nullables, defaults, and effects. 

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [x] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

Includes comprehensive test coverage for all supported types.

## Checklist

- [x] `npm test` passes

<details>

</details>

### DIFF
--- a/src/adapters/opencode/plugin.ts
+++ b/src/adapters/opencode/plugin.ts
@@ -34,24 +34,6 @@ import { AdapterPlatformType, OpenCodeAdapter } from "./index.js";
 import { PLATFORM_ENV_VARS } from "../detect.js";
 import { zod3ShapeToV4 } from "./zod3tov4.js";
 
-// Read package.json version once at module load (not on every hook call).
-// Used in the resume-injection visible signal so users can confirm in
-// OPENCODE_DEBUG logs which plugin version actually injected.
-const VERSION: string = (() => {
-  try {
-    const pkgRoot = dirname(fileURLToPath(import.meta.url));
-    // Search both the legacy depths (when bundled flat under build/) and
-    // the post-refactor depths (when compiled to build/adapters/opencode/).
-    // `../../../package.json` is the canonical location after the
-    // `src/opencode-plugin.ts → src/adapters/opencode/plugin.ts` move.
-    for (const rel of ["../../../package.json", "../package.json", "./package.json"]) {
-      const p = resolve(pkgRoot, rel);
-      if (existsSync(p)) return JSON.parse(readFileSync(p, "utf8")).version ?? "unknown";
-    }
-  } catch { /* fall through */ }
-  return "unknown";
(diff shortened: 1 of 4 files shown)


ousamabenyounes:
Approach looks sound: platform-gated on `platform === "kilo"` so OpenCode/others are untouched, and the original Zod 3 schema is still what `execute()` parses against, so runtime validation is preserved on the plugin side. The dedicated `tests/adapters/zod3tov4.test.ts` plus the integration case in `tests/opencode-plugin.test.ts` is the right shape for this kind of converter.

Thanks for the clear writeup.

ousamabenyounes:
Follow-up: ran a real red→green against the live kilo CLI 7.3.1 binary (`@kilocode/cli@7.3.1`).

**Red (no PR)** — kilo's cached install of `context-mode@1.0.141` from npm:
```
TypeError: undefined is not an object (evaluating 'n._zod.def')
      at process (/$bunfs/root/chunk-3hsp8h8f.js:22:198266)
      at process (/$bunfs/root/chunk-3hsp8h8f.js:22:201152)
      at Bv (/$bunfs/root/chunk-3hsp8h8f.js:24:2057)
      ...
```

**Green (this PR)** — replaced `~/.cache/kilo/packages/context-mode@latest/node_modules/context-mode/build` with the PR build; same `kilo run` invocation: clean exit, no `_zod.def` crash (only an unrelated `AI_LoadAPIKeyError` because I didn't have GOOGLE_GENERATIVE_AI_API_KEY set).

Gotcha worth flagging for anyone else reproducing this: `kilo` does **not** load the plugin from the project's local `node_modules` — it npm-installs the spec into its own cache at `~/.cache/kilo/packages/<name>@latest/` and loads from there. So a local `npm install ./path/to/context-mode` in the working directory has no effect; you have to overwrite the cached install (or rely on `npm publish`).

The fix activates correctly on `platform === "kilo"` via `KILO_PID`, which kilo CLI 7.3.1 still sets.

Issue #559, opened 2026-05-13
ctx-upgrade leaves old server process running (zombie instance on upgrade)


ishabana:
## Bug

Running `ctx-upgrade` installs the new plugin version and starts a new server process, but never terminates the old server process. Both versions run concurrently after every upgrade.

## Reproduction

1. Note the running node processes: `ps aux | grep context-mode`
2. Run `ctx-upgrade`
3. Check processes again — old version (`1.0.x/start.mjs`) is still alive alongside the new one

## Impact

Each upgrade leaks a server process. If the user upgrades multiple times in a session (e.g. troubleshooting), multiple zombie instances accumulate. In combination with the WAL contention bug (see related issue), this caused a 238MB WAL file and every `ctx_search` call hanging indefinitely.

## Expected behaviour

`ctx-upgrade` should send SIGTERM (or equivalent on Windows) to the currently running server before starting the new one.

## Environment

- Platform: Windows 11
- Claude Code plugin
- context-mode v1.0.125 → v1.0.127

mksglu:
Fixed in v1.0.128, just published. Thank you @ishabana — your reproducer was clean (`ps aux | grep context-mode` before/after) and your impact analysis (5-instance accumulation → #560) made the architectural class visible. Both fixes shipped together since they compose.

## Verification

The leak you observed isn't Windows-specific. We verified empirically on macOS during diagnosis: 5 PIDs across 3 versions (1.0.113, 1.0.117, 1.0.126) running concurrently, oldest 8h22m alive. Same root cause: `cli.ts:746 upgrade()` ran `npm install -g` but never signaled sibling MCP processes; `lifecycle.ts:126 shutdown()` only fired on parent death / SIGTERM / stdin-EOF — none of which `npm install -g` triggers.

## Fix architecture (v1.0.128)

**New helper `src/util/sibling-mcp.ts`**:
- Cross-OS sibling MCP discovery — `pgrep -f` on POSIX, PowerShell `Get-CimInstance Win32_Process` on Windows (NOT deprecated wmic)
- Discovery regex catches both `~/.claude/plugins/cache/context-mode/context-mode/*/start.mjs` AND `~/.claude/plugins/marketplaces/context-mode/start.mjs` shapes
- Skips `process.pid` AND `process.ppid` (parent claude process — never kill the host)

**SIGTERM → 1500ms wait → SIGKILL escalation**:
[...]

Issue #514, opened 2026-05-10
Node 26 silently skips better-sqlite3; /ctx-upgrade does not recover


alaa-nassef:
## Summary

On Node.js v26, fresh installs of context-mode silently ship without `better-sqlite3`, leaving FTS5 disabled. `/ctx-upgrade` does not recover the installation — it runs `npm install` again, which keeps skipping the package for the same reason. The user is left with a permanent `[FAIL] FTS5 / SQLite: FAIL — Cannot find module 'better-sqlite3'` and no path forward via the supported tooling.

## Environment

- macOS (Darwin 25.4.0, arm64)
- Node v26.0.0 (NODE_MODULE_VERSION 147), npm v11.12.1
- bun v1.3.13 present
- context-mode v1.0.118 (installed via Claude Code marketplace)
- Xcode CLT and Python toolchain available (verified via `node-gyp` fetching headers and starting compile)

## Reproduction

1. Install context-mode through the Claude Code marketplace on a host running Node 26.
2. Run `/ctx-doctor`. Output includes:
   ```
   [FAIL] FTS5 / SQLite: FAIL — Cannot find module 'better-sqlite3'
   Require stack:
   - .../1.0.118/server.bundle.mjs
   ```
3. Run `/ctx-upgrade`. It reports success, but `/ctx-doctor` continues to report the same failure.
4. `node_modules/better-sqlite3` does not exist on disk; only `@types/better-sqlite3` is present.

### 1. better-sqlite3 declares engines that exclude Node 26

Every published 12.x (12.0.0 through 12.9.0) declares:

```json
"engines": { "node": "20.x || 22.x || 23.x || 24.x || 25.x" }
```

Node 26 is therefore "incompatible" from npm's perspective.

### 2. npm silently drops optional deps with engine mismatches

`package.json` lists better-sqlite3 under `optionalDependencies`. On npm 7+, an optional dependency whose `engines` field does not match the running Node is silently excluded from the resolved tree — no warning, no error. Concretely, the following all return "up to date" without writing the package on Node 26:

```bash
npm install
npm install better-sqlite3@12.6.2
npm install better-sqlite3@12.6.2 --include=optional --force
npm install https://registry.npmjs.org/better-sqlite3/-/better-sqlite3-12.6.2.tgz --force
```

`node_modules/.package-lock.json` reflects the omission — only `@types/better-sqlite3` is in the resolved tree.

[...]

PR #650, opened 2026-05-20
feat(opencode): streamline installation process and improve CLI upgrade

mikij:
## What / Why / How

Remove redundant global npm install step from installation instructions for both OpenCode and KiloCode integrations. Refactor CLI upgrade command to omit global package updates for these platforms.

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [x] OpenCode
- [x] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

Added test for regression detection

## Checklist

- [x] `npm test` passes

<details>

</details>

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -1402,20 +1402,20 @@ async function upgrade(opts?: { platform?: string }) {
               color.dim("\n  Try (fallback): /context-mode:ctx-doctor"),
           );
         }
-      }
-
-      // Update global npm
-      s.start("Updating npm global package");
-      try {
-        npmExecFile(["install", "-g", pluginRoot, "--no-audit", "--no-fund"], {
-          stdio: "pipe",
-          timeout: 30000,
-        });
-        s.stop(color.green("npm global updated"));
-        changes.push("Updated npm global package");
-      } catch {
-        s.stop(color.yellow("npm global update skipped"));
-        p.log.info(color.dim("  Could not update global npm — may need sudo or standalone install"));
+        
+        // Update global npm
+        s.start("Updating npm global package");
+        try {
+          npmExecFile(["install", "-g", pluginRoot, "--no-audit", "--no-fund"], {
+            stdio: "pipe",
+            timeout: 30000,
+          });
+          s.stop(color.green("npm global updated"));
+          changes.push("Updated npm global package");
+        } catch {
+          s.stop(color.yellow("npm global update skipped"));
+          p.log.info(color.dim("  Could not update global npm — may need sudo or standalone install"));
+        }
       }
 
       // Cleanup
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -139,6 +139,9 @@ async function hookDispatch(platform: string, event: string): Promise<void> {
  * Entry point
  * ------------------------------------------------------- */
 
+const IN_PROCESS_PLUGIN_PLATFORMS = new Set(["opencode", "kilo"]);
+const isInProcessPluginPlatform = (p: string | undefined) =>
+  p ? IN_PROCESS_PLUGIN_PLATFORMS.has(p) : false;
 const args = process.argv.slice(2);
 
 if (args[0] === "doctor") {
@@ -278,7 +281,7 @@ function cachePluginRoot(platform: string): string {
 
 function getPluginRoot(): string {
   const platform = detectPlatform().platform;
-  if (platform === 'opencode' || platform === 'kilo') {
+  if (isInProcessPluginPlatform(platform)) {
     return cachePluginRoot(platform);
   }
   return defaultPluginRoot();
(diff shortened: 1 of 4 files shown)


src/cli.ts
mksglu, line 1: Why that file changed?

mksglu (changes requested):

src/cli.ts
mikij: because I want to omit CLI upgrade installs or updates context-mode globally as npm package when we are upgrading OpenCode or KiloCode

mikij: otherwise CLI silently installs it. it will install/update only when/if other platform is upgraded for the same user

ousamabenyounes (commented):
Hey @mikij — reviewed per @mksglu's CR ask. Verified the change with a static RED/GREEN against `next` (pre-PR) vs this branch:

| branch | platform | runs `npm install -g`? |
|---|---|---|
| next | opencode | yes ← the bug |
| next | kilo | yes ← the bug |
| next | claude-code | yes (correct) |
| this PR | opencode | no ← fixed |
| this PR | kilo | no ← fixed |
| this PR | claude-code | yes (non-regression) |

The brace move is correct, and the README install-step removal honestly reflects what `cachePluginRoot()` does on these platforms — nice catch on the `changes.push("Updated npm global package")` line too, that one was lying to opencode/kilo users every upgrade.

One ask before merge, plus a couple of small follow-ups.

### Blocker — regression test

This PR changes a load-bearing branch in `upgrade()`: "never run `npm install -g context-mode` when `detection.platform` is `opencode` or `kilo`." Right now nothing pins that contract. The upgrade flow has been touched a lot recently (#514, #559, #564), and a future refactor can re-leak the global install without any visible signal — the symptom is just silent npm-tree pollution on users' machines, which we'd never catch in CI.

The PR's own checklist box `Tests added/updated (TDD: red → green)` is unchecked. The repo standard for behavior changes is a test that fails when the patch is reverted; this one has none.

The repo already has the right pattern for testing `upgrade()` — source-text inspection of `src/cli.ts` (see `tests/util/cli-upgrade-verification.test.ts:39` and `tests/util/ctx-upgrade-platform-threading.test.ts:77`, both slice the upgrade body out of `cliSrc` and assert on substrings). A test that fits that pattern is short:

```ts
test("upgrade() gates 'npm install -g' behind the opencode/kilo exclusion (PR #650)", () => {
  const upgradeIdx = cliSrc.indexOf("async function upgrade");
  const upgradeBody = cliSrc.slice(upgradeIdx, upgradeIdx + 30000);

  const gateIdx = upgradeBody.indexOf("detection.platform !== 'opencode'");
  const npmGIdx = upgradeBody.indexOf("'install', '-g'");

  // Gate must exist, and the '-g' call must sit AFTER it (i.e. inside the block).
  expect(gateIdx).toBeGreaterThan(0);
  expect(npmGIdx).toBeGreaterThan(gateIdx);

  // The closing brace of the gate must come AFTER the '-g' call,
  // not between the ABI verifier and the global install (the pre-PR shape).
  const closeBefore = upgradeBody.lastIndexOf("      }", upgradeBody.indexOf("// Cleanup"));
  expect(closeBefore).toBeGreaterThan(npmGIdx);
});
```

That mirrors how `cli-upgrade-verification.test.ts` already locks in the v1.0.114 hotfix assertions. Reverting your brace move makes this test fail; keeping it keeps it green. Happy to pair if helpful.

### Minor — would be nice, not blocking

1. **Platform check repeated.** `detection.platform !== 'opencode' && detection.platform !== 'kilo'` at `src/cli.ts:1323` now gates two unrelated blocks, and the inverse appears at `getPluginRoot()` L281. A small helper would future-proof this — when the next in-process plugin platform lands (Antigravity/Zed have been mentioned), it's one line instead of three:
   ```ts
   const IN_PROCESS_PLUGIN_PLATFORMS = new Set(['opencode', 'kilo']);
   const isInProcessPluginPlatform = (p: string) => IN_PROCESS_PLUGIN_PLATFORMS.has(p);
   ```

2. **Doctor SIGSEGV fix-it is now inconsistent** (`src/cli.ts:390`). The Linux/Node unsafe-version error still recommends `nvm install 22.5 && nvm use 22.5 && npm install -g context-mode`. For an opencode/kilo user hitting that branch, `-g` now contradicts the rest of the upgrade flow. Pre-existing, not introduced here — flagging for a separate follow-up rather than asking you to fix in this PR.

### Pre-existing wart (separate follow-up — I'll file an issue)

3. **README `cp node_modules/context-mode/configs/opencode/AGENTS.md`** — broken pre-PR too: the old `npm install -g` step wrote to the global node_modules tree, never the user's CWD, and OpenCode/Kilo install into `~/.cache/<platform>/packages/...`, also outside CWD. So `cp` from local `node_modules/context-mode/` has never resolved without an unstated `npm install context-mode` somewhere. Same pattern in Gemini-CLI, Cursor, VS-Code-Copilot, JetBrains sections. Will open a separate issue — completely out of scope for this PR.

Once the test lands, this is ready to merge from my side. Thanks for cleaning up the in-process plugin story — it was overdue.

PR #654, opened 2026-05-20
fix(hooks): reword WebFetch deny so agents don't read it as network block

kerneltoast:
## Summary

- Reword the WebFetch / curl-wget / inline-HTTP redirect messages so the leading word is "redirected" (not "blocked"), and explicitly state the redirect is a context-window optimization, **NOT a network restriction**. Newer models tolerate "blocked" fine, but on Opus 4.6 the wording trips the network-error-handling pattern and the agent falls back to training knowledge instead of trying `ctx_fetch_and_index`.
- Append a sandbox hint to `ssrfGuard`'s DNS error when the failure code looks like the resolver itself is unreachable (ETIMEOUT/ETIMEDOUT/EAI_AGAIN/ENETUNREACH/EPERM) so users learn that the MCP host inherits its sandbox state at spawn and runtime sandbox toggles on the host won't reach it -- the actionable fix is to restart the host with network access enabled.
- The bug is most lethal when both fire in sequence (`ctx_fetch_and_index` ETIMEOUT, then WebFetch "blocked"). On the same `parallelize-skill` session, here's the before/after — same model class (Opus 4.6), same compound-failure conditions:

| | Pre-fix verdict (old "blocked") | Post-fix verdict (new "redirected") |
|---|---|---|
| Agent's verbatim decision | *"Network access to arxiv.org is blocked by the sandbox. Let me work with what I can access and rely on my training knowledge of these papers."* | *"WebFetch is being redirected to context-mode. Let me use the context-mode fetch tool for all of these, plus run more targeted searches."* |
| Outcome | Capitulated to training | Pivoted to `ctx_fetch_and_index`, hit the (still-failing) DNS issue, fell back to WebSearch, produced a 5,172-char synthesis with cited findings |
| Fallback phrases (`"Network access blocked"`, `"fall back to training"`) | Present | Absent everywhere in the transcript |

## Test plan

- [x] Update hook tests for the new substrings: `tests/hooks/core-routing.test.ts`, `tests/hooks/cursor-hooks.test.ts`, `tests/hooks/tool-naming.test.ts`, `tests/hooks/integration.test.ts`.
- [x] Hook + session-DB suites green: `bun x vitest run tests/hooks/ tests/session/session-db.test.ts` -> 521 passed, 1 skipped.
- [x] Pre-existing failures across `tests/statusline*`, `tests/util/project-dir.test.ts`, `tests/security.test.ts`, `tests/core/server.test.ts` reproduce on bare `main` (10 failed / 3307 passed) and are unrelated to this change.
- [x] `bun run assert-bundle` and `bun run assert-asymmetric-drift` clean. Bundles rebuilt so the DNS hint ships via both `server.bundle.mjs` and `cli.bundle.mjs`.
- [x] **End-to-end verified on the original failure scenario.** Post-fix Opus 4.6 worker on the same `parallelize-skill` session hit the WebFetch redirect on four URLs and pivoted as quoted in the table above; reasoning stayed constructive through the subsequent DNS failure on `ctx_fetch_and_index` instead of capitulating.
- [x] Synthetic A/B with Sonnet and Opus 4.7 subagents calling WebFetch directly under both wordings: both models quote the new wording verbatim and classify it as a tool redirect.

## Known follow-up

The DNS hint in `ssrfGuard` only lands on the single-URL error return path. `ctx_fetch_and_index`'s batch wrapper formats each DNS ETIMEOUT through its own per-request error line, which bypasses `ssrfGuard`'s return — so in the post-fix Opus 4.6 transcript the hint didn't actually surface (the agent saw the bare `DNS lookup failed for "arxiv.org": getaddrinfo ETIMEOUT` line). A follow-up needs to plumb the sandbox hint through the batch formatter so the common multi-URL case gets the same diagnostic. The wording-only fix still resolves the agent-behavior part of the bug independently, as the table above shows.

## Notes

Drafted with the help of Claude (Opus 4.7) under direct human review (bug report, JSONL transcripts, and test scenarios supplied by the author; PR text reviewed before submission).


mksglu:
Did you test that manually?

kerneltoast:
> Did you test that manually?

Yes. I had an Opus 4.6 session that reproduced this issue 100% of the time trying to fetch an arxiv.org page due to this issue

I started a separate Opus 4.7 session on the context mode tree to diagnose the problem, and then restarted the MCP in the 4.6 session. The fix worked and my 4.6 session could finally fetch from arxiv.org.

Anecdotally, I have actually run into this issue several times over the past couple months but only debugged it today because I saw the failure happened 100% of the time with arxiv in my 4.6 session.

ousamabenyounes:
Hi @kerneltoast — thanks for this fix, the wording change itself is solid. I reproduced the bug locally and confirmed the fix works (red/green proof below), then took a closer look at the `ssrfGuard` half. A few things I'd like to flag, with a patch you're welcome to take, adapt, or drop.

## Red/green verification

Checked out the PR, reverted only `src/server.ts` + `hooks/core/routing.mjs` to `next` while keeping the new tests → **14 failures** (`Expected: "WebFetch redirected" / Received: "...WebFetch blocked..."`). Restored the PR sources → all hook-routing tests pass. The 2 pre-existing failures in `tests/core/server.test.ts` (tsc binary + projectRoot resolution) reproduce identically on `next` — not caused by this PR.

## Findings on the `ssrfGuard` half

1. **No test coverage for the new code.** The existing SSRF suite (`tests/core/server.test.ts:3528`) never exercises a `dns.lookup` that throws `ETIMEOUT`/`ETIMEDOUT`/`EAI_AGAIN`/`ENETUNREACH`/`EPERM`, so `looksLikeSandbox` + the hint message land untested. Repo policy is "every code change ships with a test that fails when the patch is reverted".

2. **The hint doesn't surface on the path most users will hit.** You called this out yourself in the "Known follow-up" — `ctx_fetch_and_index` runs each URL in a worker subprocess (`server.ts:2721`, `reason: "exit"`), so DNS failures arrive as captured stderr from the child and bypass `ssrfGuard`'s `catch` entirely. As-is, the hint is wired but inert for the batch flow.

3. **Hardcoded literals.** The five DNS codes + the multi-line hint string live inline in the `catch`. Repo's "no hardcoded values" rule + the "same literal 2+ times → constant" rule both apply once we want this in two paths.

## Proposed patch

Hoists the codes and the hint to module-level exports, adds a small `looksLikeSandboxError(message)` helper that matches against the codes in either an `errno` or the formatted message, and plumbs the hint through the two `fetchOneUrl` error branches (subprocess `exit` and `throw`). 13 new tests cover each code, two negative cases (`ENOTFOUND`, `ECONNREFUSED`), the constants surface, and a source-grep check that both `fetchOneUrl` branches reference the helper + constant.

<details>
<summary><b>diff — src/server.ts</b></summary>

```diff
diff --git a/src/server.ts b/src/server.ts
index ef39dc7..0ebe127 100644
--- a/src/server.ts
+++ b/src/server.ts
@@ -2525,6 +2525,34 @@ main();
 const FETCH_TTL_MS = 24 * 60 * 60 * 1000; // 24 hours
 const FETCH_PREVIEW_LIMIT = 3072;
 
+// libuv / getaddrinfo error codes that typically indicate the resolver itself
+// can't reach a nameserver — common when the MCP host process is running under
+// a sandbox that blocks outbound network. Surfaced as a hint so the user knows
+// the failure isn't a bad URL; the MCP server inherits its sandbox state at
+// spawn time and can't be re-permissioned at runtime by toggling the host's
+// sandbox (e.g. Claude Code's /sandbox).
+export const SANDBOX_DNS_CODES: ReadonlySet<string> = new Set([
+  "ETIMEOUT",
+  "ETIMEDOUT",
+  "EAI_AGAIN",
+  "ENETUNREACH",
+  "EPERM",
+]);
+
+export const SANDBOX_HINT =
+  " — looks like the MCP host process can't reach a DNS resolver. If your" +
+  " editor/CLI runs MCP servers under a network sandbox, the server inherits" +
+  " that state at spawn and a runtime sandbox toggle won't reach it. Restart" +
+  " the MCP host with network access enabled.";
+
+/** True if `s` mentions a libuv DNS code that suggests a sandboxed resolver. */
+export function looksLikeSandboxError(s: string): boolean {
+  for (const code of SANDBOX_DNS_CODES) {
+    if (s.includes(code)) return true;
+  }
+  return false;
+}
+
 type FetchOneResult =
   | { kind: "cached"; label: string; chunkCount: number; estimatedBytes: number; ageStr: string }
   | { kind: "fetched"; url: string; source?: string; markdown: string; header: string }
@@ -2605,18 +2633,10 @@ async function ssrfGuard(rawUrl: string): Promise<FetchOneResult | null> {
       }
     }
   } catch (err) {
-    // libuv DNS error codes that typically indicate the resolver itself can't
-    // reach a nameserver — common when the MCP host process is running under
-    // a sandbox that blocks outbound network. Surface a sandbox hint so the
-    // user knows the failure isn't a bad URL; the MCP server inherits its
-    // sandbox state at spawn time and can't be re-permissioned at runtime by
-    // toggling the host's sandbox (e.g. Claude Code's /sandbox).
     const errCode = (err as NodeJS.ErrnoException | undefined)?.code ?? "";
-    const looksLikeSandbox = errCode === "ETIMEOUT" || errCode === "ETIMEDOUT" ||
-      errCode === "EAI_AGAIN" || errCode === "ENETUNREACH" || errCode === "EPERM";
     const baseMsg = err instanceof Error ? err.message : String(err);
-    const hint = looksLikeSandbox
-      ? " — looks like the MCP host process can't reach a DNS resolver. If your editor/CLI runs MCP servers under a network sandbox, the server inherits that state at spawn and a runtime sandbox toggle won't reach it. Restart the MCP host with network access enabled."
+    const hint = SANDBOX_DNS_CODES.has(errCode) || looksLikeSandboxError(baseMsg)
+      ? SANDBOX_HINT
       : "";
     return {
       kind: "fetch_error",
@@ -2718,7 +2738,9 @@ async function fetchOneUrl(url: string, source: string | undefined, force: boole
       timeout: 30_000,
     });
     if (result.exitCode !== 0) {
-      return { kind: "fetch_error", url, error: result.stderr || result.stdout || "unknown error", reason: "exit" };
+      const baseErr = result.stderr || result.stdout || "unknown error";
+      const hint = looksLikeSandboxError(baseErr) ? SANDBOX_HINT : "";
+      return { kind: "fetch_error", url, error: `${baseErr}${hint}`, reason: "exit" };
     }
     const header = (result.stdout || "").trim();
     let markdown: string;
@@ -2732,10 +2754,15 @@ async function fetchOneUrl(url: string, source: string | undefined, force: boole
     }
     return { kind: "fetched", url, source, markdown, header };
   } catch (err: unknown) {
+    const errCode = (err as NodeJS.ErrnoException | undefined)?.code ?? "";
+    const baseErr = err instanceof Error ? err.message : String(err);
+    const hint = SANDBOX_DNS_CODES.has(errCode) || looksLikeSandboxError(baseErr)
+      ? SANDBOX_HINT
+      : "";
     return {
       kind: "fetch_error",
       url,
-      error: err instanceof Error ? err.message : String(err),
+      error: `${baseErr}${hint}`,
       reason: "throw",
     };
   } finally {
```

</details>

<details>
<summary><b>new file — tests/core/sandbox-hint.test.ts</b></summary>

```ts
import { describe, test, expect } from "vitest";
import { readFileSync } from "node:fs";
import { resolve, dirname } from "node:path";
import { fileURLToPath } from "node:url";
import {
  SANDBOX_DNS_CODES,
  SANDBOX_HINT,
  looksLikeSandboxError,
} from "../../src/server.js";

const __dirname = dirname(fileURLToPath(import.meta.url));

describe("sandbox-hint: DNS code detection", () => {
  test.each([
    ["getaddrinfo ETIMEOUT arxiv.org"],
    ["connect ETIMEDOUT 1.2.3.4:443"],
    ["getaddrinfo EAI_AGAIN example.com"],
    ["connect ENETUNREACH 1.2.3.4"],
    ["EPERM: operation not permitted"],
  ])("flags %s as sandbox-style failure", (msg) => {
    expect(looksLikeSandboxError(msg)).toBe(true);
  });

  test("does NOT flag a normal NXDOMAIN (ENOTFOUND)", () => {
    expect(looksLikeSandboxError("getaddrinfo ENOTFOUND no-such-domain.example"))
      .toBe(false);
  });

  test("does NOT flag a refused connection (ECONNREFUSED)", () => {
    expect(looksLikeSandboxError("connect ECONNREFUSED 1.2.3.4")).toBe(false);
  });

  test("does NOT flag empty / unrelated strings", () => {
    expect(looksLikeSandboxError("")).toBe(false);
    expect(looksLikeSandboxError("Some other error")).toBe(false);
  });
});

describe("sandbox-hint: constants surface", () => {
  test("SANDBOX_DNS_CODES exposes the documented libuv codes", () => {
    expect(SANDBOX_DNS_CODES.size).toBe(5);
    for (const c of ["ETIMEOUT", "ETIMEDOUT", "EAI_AGAIN", "ENETUNREACH", "EPERM"]) {
      expect(SANDBOX_DNS_CODES.has(c)).toBe(true);
    }
  });

  test("SANDBOX_HINT mentions sandbox + restart guidance", () => {
    expect(SANDBOX_HINT).toMatch(/sandbox/i);
    expect(SANDBOX_HINT).toMatch(/restart/i);
    expect(SANDBOX_HINT.startsWith(" — ")).toBe(true);
  });
});

describe("sandbox-hint: fetchOneUrl plumbs hint through every error path", () => {
  const serverSrc = readFileSync(
    resolve(__dirname, "../../src/server.ts"),
    "utf-8",
  );
  const fetchOneSrc = serverSrc.match(/async function fetchOneUrl\([\s\S]+?^}/m);
  if (!fetchOneSrc) throw new Error("fetchOneUrl source not found");
  const block = fetchOneSrc[0];

  test("subprocess stderr branch (exitCode !== 0) appends SANDBOX_HINT", () => {
    const exitBranch = block.match(/if \(result\.exitCode !== 0\)[\s\S]+?\}/);
    expect(exitBranch).not.toBeNull();
    expect(exitBranch![0]).toContain("looksLikeSandboxError");
    expect(exitBranch![0]).toContain("SANDBOX_HINT");
  });

  test("throw catch branch appends SANDBOX_HINT", () => {
    const catchBranch = block.match(/catch \(err: unknown\)[\s\S]+?reason: "throw",[\s\S]+?\}/);
    expect(catchBranch).not.toBeNull();
    expect(catchBranch![0]).toContain("looksLikeSandboxError");
    expect(catchBranch![0]).toContain("SANDBOX_HINT");
  });

  test("ssrfGuard catch uses the shared SANDBOX_HINT constant (no inline literal)", () => {
    const ssrfSrc = serverSrc.match(/async function ssrfGuard\([\s\S]+?^}/m);
    expect(ssrfSrc).not.toBeNull();
    expect(ssrfSrc![0]).toContain("SANDBOX_HINT");
    expect(ssrfSrc![0]).not.toContain('"ETIMEOUT"');
  });
});
```

</details>

## Local test logs

Same workspace as the red/green check above, against `fix-webfetch-redirect-wording` with the patch + new test file applied.

```text
# RED — stash the src/server.ts patch, keep the new test file
$ git stash push -- src/server.ts && bun x vitest run tests/core/sandbox-hint.test.ts
 Test Files  1 failed (1)
      Tests  13 failed (13)
   Duration  1.18s

# GREEN — restore patch
$ git stash pop && bun x vitest run tests/core/sandbox-hint.test.ts
 Test Files  1 passed (1)
      Tests  13 passed (13)
   Duration  1.06s

# Full hooks + sandbox-hint regression with patch applied
$ bun x vitest run tests/core/sandbox-hint.test.ts tests/hooks/
 Test Files  22 passed (22)
      Tests  451 passed | 1 skipped (452)
   Duration  10.92s
```

After applying, `server.bundle.mjs` / `cli.bundle.mjs` / `stats.json` need regeneration (`bun run build` + `bun run assert-bundle`).

Happy to open a follow-up PR on top of this branch if that's easier — whatever you prefer. The wording fix itself stands on its own either way.

murataslan1:
Manual behavior test verdict: mergeable.

I tested this as a model-behavior change, not as a normal string/code review. Full evidence and raw probe outputs are here:

https://gist.github.com/murataslan1/befe0628e4caa4ebff8fbc97e2b5e2b7

The core issue is real: the old `WebFetch blocked` framing can be interpreted by the model as a network restriction after a preceding `ctx_fetch_and_index` DNS failure. In a compound-failure probe, the old wording produced `network_is_blocked: true` and the next action was to stop network attempts / search existing context. The PR wording produced `network_is_blocked: false` and kept the next action on `ctx_fetch_and_index`, with `ctx_execute` as fallback.

I also verified the PR hook path directly for WebFetch, `mcp_web_fetch`, `mcp_fetch_tool`, curl, inline HTTP, and the silent file-output curl passthrough. Focused tests passed: `521 passed | 1 skipped`, plus `assert-bundle` and `assert-asymmetric-drift`.

Limitations: I could not run the exact Opus 4.6 reproduction locally because my local Claude CLI is not logged in. The behavior probe still reproduced the critical WebFetch compound-failure distinction under Codex. A parallel Claude Code pass reported current Sonnet handled both old and new wording correctly, which matches the PR's claim that newer models tolerate `blocked` better.

One optional tightening: bind the `full network access` claim even more explicitly to the context-mode tool, e.g. `ctx_fetch_and_index has full network access - it can reach the same URL. WebFetch/curl/wget are disabled here regardless; do not retry them.` That would reduce the chance that a weaker model reads `full network access` as permission to retry curl/WebFetch.

mksglu:
Hi @kerneltoast! It's Mert. Firstly, thanks a lot. A straightforward code review is not enough here, because even a single-word change in the tool instructions can alter model behavior.

This is a critical PR, so it needs to be manually tested, and the actual problem must be clearly understood before moving forward.

Please give me time to test that on manually. I'd like to understand the root cause here actually.

Share with me like "Try it" prompts for real-world testing.

kerneltoast:
> Hi @kerneltoast! It's Mert. Firstly, thanks a lot. A straightforward code review is not enough here, because even a single-word change in the tool instructions can alter model behavior.
> 
> This is a critical PR, so it needs to be manually tested, and the actual problem must be clearly understood before moving forward.
> 
> Please give me time to test that on manually. I'd like to understand the root cause here actually.
> 
> Share with me like "Try it" prompts for real-world testing.

Gotcha, no problem. I'll work on creating a minimal reproducer prompt today.

mksglu:
> > Hi @kerneltoast! It's Mert. Firstly, thanks a lot. A straightforward code review is not enough here, because even a single-word change in the tool instructions can alter model behavior.
> > This is a critical PR, so it needs to be manually tested, and the actual problem must be clearly understood before moving forward.
> > Please give me time to test that on manually. I'd like to understand the root cause here actually.
> > Share with me like "Try it" prompts for real-world testing.
> 
> Gotcha, no problem. I'll work on creating a minimal reproducer prompt today.

That's nice. Let me know us when you're ready. We're going to test that case to the different LLM's. Firstly, we should repro that issue. @murataslan1 @ousamabenyounes So please give us instruction for the repro.

kerneltoast:
@mksglu Here is a minimal reproducer prompt: [webfetch-deny-misread-repro.txt](https://github.com/user-attachments/files/28132523/webfetch-deny-misread-repro.txt)

To run it:
```bash
claude -p --model "claude-opus-4-6[1m]" --output-format json < webfetch-deny-misread-repro.txt
```

The bug reproduces sporadically on Opus 4.6 about 60% of the time from several runs I did, so run it ~10 times to be certain. Look for `"concludes_no_network": true|false` in the `result` field of the returned JSON (`true` = bug hit).

And here is Claude's analysis for the reproducer: [webfetch-deny-misread-analysis.md](https://github.com/user-attachments/files/28132636/webfetch-deny-misread-analysis.md)

The analysis explains what this is actually testing and why it's an artificial reproducer instead of a natural prompt (because creating a "natural" way to consistently reproduce this is tricky).

murataslan1:
Manual behavior validation from my side:

I tested the old and new hook wording with the same prompt/task.

Old wording:
`WebFetch blocked`

Observed behavior:
The agent tended to interpret this as a real network/DNS failure and stopped instead of choosing the next context-mode route.

New wording:
`WebFetch redirected ... NOT a network restriction`

Observed behavior:
The agent correctly treated this as a context-window routing redirect and selected `ctx_fetch_and_index` / `ctx_execute` as the next step.

Detailed notes:
https://gist.github.com/murataslan1/800422c44f9df4da2abf6029518fc2e3

Screen recording:
https://x.com/iammurataslan/status/2057844261303783846?s=20

This is not a deterministic unit test, but it is a useful manual behavior check. From an agent-behavior perspective, this PR looks directionally correct to me.

mksglu:
@kerneltoast — first and most importantly: **thank you**. Your reproduction on Opus 4.6 and the precise diagnosis (the `"blocked"` token tripping the network-error-handling pattern and the agent falling back to training knowledge) was the single observation that turned a one-line wording tweak into a corpus-wide audit of how every prompt surface we ship interacts with cross-LLM safety priors. The rewrite landed in #683 and shipped in **v1.0.147**, but the credit for noticing the problem and giving us a reliable repro is yours.

I'm closing this PR in favor of #683 because the surface area outgrew a single deny-reason fix — but everything you pointed at is now codified as policy, not just patched. Closing carries no "your PR was wrong" signal; the opposite. You can ignore the section below if you just want the summary, but I want to be transparent about what landed so this isn't a black-box close.

### What landed in v1.0.147 (substitutes #654)

**1. The original WebFetch / curl / wget / inline-HTTP fix.** All four CASE A redirect deny reasons in `hooks/core/routing.mjs` now open with the affirmative verb `"redirected"`, name the alternative `ctx_*` tool by an imperative call, affirm capability (`"<tool> has full network access"`), and end with a positive transient-DNS retry hint (`EAI_AGAIN | ETIMEDOUT | ENETUNREACH`). The `"blocked"` token is gone from CASE A entirely.

**2. ADR-0002 (tool description style) — new, canonical and contract-tested.** Documents the cross-LLM bias rubric (Constitutional AI priors, bare-NOT negation, emoji-bullet leakage, off-spec UPPERCASE sections, RFC 2119 imperative hierarchy). Locks a canonical structure for every routing-target ctx_* tool: `WHEN: → WHEN NOT: → RETURNS: → EXAMPLE:`, markdown `- ` bullets only, header-on-own-line RETURNS form. Enforced by a contract test that scans `src/server.ts` on every commit (131 forbidden-token assertions + 7 canonical-structure assertions + per-tool guards). Future contributors either adhere to it or amend the ADR.

**3. ADR-0003 (routing deny reasons: redirect ≠ restriction) — also new.** Distinguishes CASE A (routing redirect to an alternative tool) from CASE B (true security/policy denial). CASE A MUST open with `"redirected"`, MUST NOT contain bare `BLOCKED`, MUST NOT contain bare-NOT negations (`NOT a network restriction`, `Do NOT retry with curl`), MUST NOT contain org-rationale prefaces (`for context-window efficiency`). Three contract-test guards lock each rule. The §Amendment + §Second amendment sections document the empirical reasoning chain that led from your original observation to the final policy.

**4. Two LLM-behavioral findings the rewrite was calibrated against.**
- **Negation framing primes the forbidden frame (ironic process theory).** Tri-LLM probe measured Haiku capitulation regressing from 0/6 → 2/6 when the original `"blocked"` was replaced with `"(context-window optimization, NOT a network restriction)"` — the bare-NOT construct anchored attention on the very frame it tried to deny. Resolution: affirmative `"<tool> has full network access"` carries the same signal positively.
- **Threshold-based heuristics are unactionable at tool-selection time.** Phrases like `"Bash with output >20 lines → use ctx_execute"` were rewritten around INTENT (`"when you intend to PROCESS the output (filter, count, parse, aggregate)"`) because the LLM cannot predict output size before running — only its own intent.

**5. Description corpus rewrite — every ctx_* tool.** Eleven tool descriptions in `src/server.ts` rewritten to surface load-bearing mechanism in LLM-actionable terms (the actual ranking pipeline, custom TTL cache, Think-in-Code principle, unified knowledge base reaching auto-captured session memory, etc.). The `hooks/routing-block.mjs` system-prompt injection got the same treatment — `"context window floods"` → `"every byte enters your conversation memory and costs reasoning capacity"`. Five-language-aware tokenizer thinking throughout (Claude / GPT / Gemini / Llama / Qwen).

**6. Behavior preserved.** No tool surface area changed, no schema breaking change, no behavior regression. Description-and-deny-reason rewrite only. 635 tests pass.

### Please test v1.0.147

If you have the repro from your original report (Opus 4.6 + WebFetch fetch attempt), `npm i -g @mksglu/context-mode@1.0.147` or `/context-mode:ctx-upgrade` and verify the agent now follows the redirect to `ctx_fetch_and_index` cleanly. If it still falls back to training knowledge, that's a real regression and I want to know about it — open an issue and tag this PR.

### Invitation

Your diagnostic discipline (reproduction → token-level hypothesis → repro on a specific model version) is the exact bar I want collaborators at. If you'd like to be part of the private engineering channel where the deeper architectural reviews happen (cross-LLM probe design, ADR drafting, contract-test extensions), reply here or open a discussion and I'll send you the invite. No pressure, no commitment expected — the door is open.

Thanks again. This release exists in this shape because of your report.

— Mert

mksglu:
Closing in favor of #683 (now merged into next and shipped in v1.0.147). The credit + technical breakdown is in the comment above — thank you again @kerneltoast.

kerneltoast:
> If you have the repro from your original report (Opus 4.6 + WebFetch fetch attempt), `npm i -g @mksglu/context-mode@1.0.147` or `/context-mode:ctx-upgrade` and verify the agent now follows the redirect to `ctx_fetch_and_index` cleanly. If it still falls back to training knowledge, that's a real regression and I want to know about it — open an issue and tag this PR.

@mksglu Just tested 1.0.150 on the original session where I ran into this and it is indeed fixed. Thanks for the thorough audit and fixing this pattern across context-mode! I didn't catch the negative framing issue in my PR with smaller models like Haiku fixating on "NOT", and I hadn't tested Haiku. The positive framing is indeed the correct solution for better compliance.

> If you'd like to be part of the private engineering channel where the deeper architectural reviews happen (cross-LLM probe design, ADR drafting, contract-test extensions), reply here or open a discussion and I'll send you the invite. No pressure, no commitment expected — the door is open.

Sure, I'd love to. I rely on context-mode quite heavily :slightly_smiling_face:

mksglu:
> > If you have the repro from your original report (Opus 4.6 + WebFetch fetch attempt), `npm i -g @mksglu/context-mode@1.0.147` or `/context-mode:ctx-upgrade` and verify the agent now follows the redirect to `ctx_fetch_and_index` cleanly. If it still falls back to training knowledge, that's a real regression and I want to know about it — open an issue and tag this PR.
> 
> @mksglu Just tested 1.0.150 on the original session where I ran into this and it is indeed fixed. Thanks for the thorough audit and fixing this pattern across context-mode! I didn't catch the negative framing issue in my PR with smaller models like Haiku fixating on "NOT", and I hadn't tested Haiku. The positive framing is indeed the correct solution for better compliance.
> 
> > If you'd like to be part of the private engineering channel where the deeper architectural reviews happen (cross-LLM probe design, ADR drafting, contract-test extensions), reply here or open a discussion and I'll send you the invite. No pressure, no commitment expected — the door is open.
> 
> Sure, I'd love to. I rely on context-mode quite heavily 🙂

Reach out me via DM

PR #657, opened 2026-05-21
fix(batch_execute): preserve heredoc commands

### LINKED ISSUE #656: ctx_batch_execute corrupts heredoc commands by appending stderr redirection
## Summary

`ctx_batch_execute` corrupts shell commands that use heredocs because `runBatchCommands()` appends `2>&1` directly to each command string before passing it to the shell executor.

For a command like:

```sh
node - <<'NODE'
console.log('heredoc works')
NODE
```

`ctx_batch_execute` effectively executes:

```sh
node - <<'NODE'
console.log('heredoc works')
NODE 2>&1
```

That changes the heredoc terminator from `NODE` to `NODE 2>&1`, so the shell does not terminate the heredoc where intended. The batch result can then show `(no output)` or otherwise misleading indexed/search output even though the command itself is valid.

## Reproduction

Call `ctx_batch_execute` with a heredoc command:

```json
{
  "commands": [
    {
      "label": "heredoc node simple",
      "command": "node - <<'NODE'\nconsole.log('heredoc works');\nconsole.log(process.version);\nNODE"
    }
  ],
  "queries": ["heredoc works node version"],
  "timeout": 5000
}
```

Observed output includes an indexed section for the command, but the section body is empty / `(no output)`.

For comparison, the same command works through `ctx_execute(language: "shell", code: ...)`.

## Expected behavior

`ctx_batch_execute` should execute the user's command string as provided. It should not append shell syntax after the command body because that can alter multiline shell constructs such as heredocs.

Stderr should still be captured and indexed alongside stdout.

## Likely cause

Current upstream `main` still appears to append redirection in `src/server.ts`:

```ts
code: `${nodeOptsPrefix}${cmd.command} 2>&1`,
```

This appears in both the serial and parallel `runBatchCommands()` paths.

## Suggested fix

Do not mutate the command with `2>&1`. Instead:

1. Execute the command string as-is:

```ts
code: `${nodeOptsPrefix}${cmd.command}`,
```

2. Combine the executor's captured streams before formatting/indexing:

```ts
const combinedOutput = result.stderr
  ? `${result.stdout}${result.stdout && !result.stdout.endsWith("\n") ? "\n" : ""}${result.stderr}`
  : result.stdout;

formatCommandOutput(cmd.label, combinedOutput, onFsBytes)
```

Apply the same pattern in both serial and parallel paths.

## Environment

- `context-mode`: `1.0.146`
- npm latest checked: `1.0.146`
- Platform observed: macOS / Pi coding agent

## Impact

This is easy for agents to hit because `ctx_batch_execute` is recommended as the primary data-gathering tool, and heredoc-based Node/Python snippets are a common way to keep complex analysis code readable inside shell commands.

Noctivoro:
## Summary

- stop appending `2>&1` directly to `ctx_batch_execute` command strings
- combine executor-captured stdout/stderr before formatting and indexing batch output
- add serial and parallel `runBatchCommands()` regression tests for heredoc commands with stderr

## Why

Fixes #656.

Appending `2>&1` after the user command mutates multiline shell constructs. For heredocs, it changes the terminator from e.g. `NODE` to `NODE 2>&1`, which can make valid commands produce empty/misleading batch output.

The executor already captures stdout and stderr separately, so this keeps the command string intact and merges the captured streams afterward.

## Verification

- `npm run typecheck`
- `npx vitest run tests/core/server.test.ts -t runBatchCommands`
- `CONTEXT_MODE_BRIDGE_DEPTH= npm test`

Note: a first full `npm test` run from inside context-mode failed unrelated Pi bridge tests because the parent environment had `CONTEXT_MODE_BRIDGE_DEPTH=1`; rerunning with that variable unset passed all tests.


ousamabenyounes:
Thanks for the fix — the source change in `src/server.ts` is correct and the new tests cover both serial and parallel heredoc paths. Reproduced the bug locally first (`bash: here-document delimited by end-of-file (wanted 'NODE')`), then confirmed the patched src makes it green. `npm run typecheck` clean, 22/22 `runBatchCommands` tests passing.

One blocker before merge, plus two suggested test additions:

### 🟥 Blocker — `cli.bundle.mjs` not regenerated

The PR updates `src/server.ts` and `server.bundle.mjs`, but `cli.bundle.mjs` still reflects the pre-fix version at lines 561 and 564:

```js
code:`${s}${h.command} 2>&1`
```

This matters because `package.json` line 57 sets:

```json
"bin": { "context-mode": "./cli.bundle.mjs" }
```

So users running the package via the CLI (`npx context-mode`, global install, etc.) would still take the previous code path — the `src/server.ts` fix would not reach them. `src/cli.ts` does `import("./server.js")` and esbuild inlines `runBatchCommands` into `cli.bundle.mjs`, which is why the function lives in both bundles.

Running `npm run bundle` locally regenerates both bundles cleanly:

```
$ npm run bundle
$ grep -c 'command} 2>&1' cli.bundle.mjs
0
$ grep -c 'combineExec\|stdout.*stderr.*endsWith' cli.bundle.mjs
1   # the new combineExecOutput is inlined
```

Could you run `npm run bundle` and commit the refreshed `cli.bundle.mjs` (and `server.bundle.mjs` too, it shifts slightly under minification)?

### 🟧 Suggested — two uncovered branches in `combineExecOutput`

The new helper has a branch that no test currently hits:

```ts
function combineExecOutput(result) {
  ...
  if (!stdout) return stderr;   // ← stderr-only path, untested
  ...
}
```

Both serial and parallel tests assert `stdout + stderr` merging, but never the stderr-only case. Suggested addition:

```ts
test("preserves stderr when command writes nothing to stdout", async () => {
  const exec = mkMockExecutor(() => ({ stdout: "", stderr: "boom\n" }));
  const { outputs } = await runBatchCommands(
    [{ label: "stderr-only", command: "node -e 'process.stderr.write(\"boom\\n\")'" }],
    { timeout: 5000, concurrency: 1, nodeOptsPrefix: NOOP_PREFIX },
    exec,
  );
  expect(outputs[0]).toContain("boom");
});
```

Similarly, the parallel `run` path at `src/server.ts:1343` calls `combineExecOutput(result)` even when `result.timedOut === true`, so stderr should still survive a timeout — also worth a regression test:

```ts
test("merges stderr into timed-out output (parallel)", async () => {
  const exec = mkMockExecutor(() => ({ stdout: "S", stderr: "E", timedOut: true }));
  const { outputs, timedOut } = await runBatchCommands(
    [{ label: "slow", command: "sleep 999" }],
    { timeout: 10, concurrency: 2, nodeOptsPrefix: NOOP_PREFIX },
    exec,
  );
  expect(timedOut).toBe(true);
  expect(outputs[0]).toContain("E");
  expect(outputs[0]).toContain("(timed out after");
});
```

Not blocking, but they'd lock down both halves of the new helper and the timeout-merge contract.

### What's already clean ✅

- `combineExecOutput` is called in both code paths (no asymmetric drift between serial and parallel).
- Existing edge-case test at `tests/core/server.test.ts:3140` correctly updated (`'echo hi 2>&1'` → `'echo hi'`); no silent skip/xfail introduced.
- Comment at `src/server.ts:1341` updated (`partial stdout` → `partial output`) to match the new combined-stream semantics.
- No dead code introduced; `formatCommandOutput` still has its single responsibility.

Once `cli.bundle.mjs` is regenerated this is good to go from my side. Thanks again 🙏

ousamabenyounes:
@mksglu fair point on manual testing — I ran an end-to-end check with a real `PolyglotExecutor` (no mocks) against the PR build, three heredoc shapes, serial + parallel. Sharing the logs so the verification is auditable.

### 🔴 RED — live `v1.0.143` (current published, before the fix)

Reproduced just now via the live `ctx_batch_execute` MCP tool:

```
# node heredoc        → (no output)
# python heredoc      → (no output)
# cat heredoc         → LINE_A
                        LINE_B
                        EOF2 2>&1     ← terminator + redirection leaking as literal text
```

The `2>&1` suffix concatenated to the heredoc terminator is exactly what the patch removes.

### 🟢 GREEN — PR build (`24bceba`), real executor

```
$ node pr-657-e2e.mjs
=== SERIAL ===
# node heredoc        → STDOUT_OK / STDERR_OK
# python heredoc      → PY_STDOUT_OK / PY_STDERR_OK
# cat heredoc         → LINE_A / LINE_B
=== PARALLEL ===
(same outputs, concurrency: 3)
=== ASSERTS ===
PASS STDOUT_OK
PASS STDERR_OK
PASS PY_STDOUT_OK
PASS PY_STDERR_OK
PASS LINE_A
PASS LINE_B
PASS no EOF2 2>&1 leak     ← 7/7
```

Plus `vitest tests/core/server.test.ts` → 22/22 pass, `tsc --noEmit` clean.

### On Murat's tool-instruction concern

Valid concern in general. For this specific PR, the diff in `src/server.ts` only touches `runBatchCommands` and adds the `combineExecOutput` helper — no MCP tool description string is modified. Model-behavior surface is untouched.

### One remaining item before merge

`cli.bundle.mjs` still reflects the pre-fix path at lines 561 and 564 (`code:\`\${s}\${h.command} 2>&1\``). Since `package.json` `bin: { "context-mode": "./cli.bundle.mjs" }`, that's the binary `npm install -g`/`npx` users execute — so the source fix won't reach them until both bundles are regenerated. Running `npm run bundle` locally produces a clean `cli.bundle.mjs` with the new `combineExecOutput` helper inlined. Once committed, I'm confident this is good to ship.

PR #664, opened 2026-05-22
fix(auto-memory): scope memory dir by projectDir to stop cross-project leak (#663)

### LINKED ISSUE #663: Feature: ctx_search with sort:timeline should scope to current projectDir
## Summary

`ctx_search` with `sort: "timeline"` returns results from all projects in the knowledge base, not just the current project. This causes cross-contamination when working in multiple projects in parallel.

## Problem

When working in two projects simultaneously (e.g., `project-a` and `project-b` in separate terminals), data indexed from `project-a` (npm outdated, package.json, etc.) appears in `ctx_search` results inside `project-b` sessions.

The content DBs are already isolated by `projectDir` hash (`Cr(projectDir)` → separate `.db` files). However, `ctx_search` with `sort: "timeline"` uses `auto-memory` and cross-session search that is not scoped to the current project.

## Expected behavior

`ctx_search` should only return results from:
1. The current session
2. Previous sessions in the **same project** (same `projectDir` hash)
3. Explicitly shared sources (if any)

It should **never** return indexed content from a different project's working directory.

## Root cause

Looking at the server bundle, `searchEvents` in the session DB queries by `session_id` but `ctx_search` with timeline sort appears to join across sessions without filtering by `project_dir`.

## Environment

- context-mode: installed via pi (latest)
- Working in two project directories simultaneously
- Both projects share the same `~/.claude/` config directory

## Workaround

We are now setting `CONTEXT_MODE_DATA_DIR` per-project in `.claude/settings.local.json` to fully separate the data directories. This works but requires manual configuration for every project.

sebastianbreguel:
## Problem (#663)

`getMemoryDir()` ignored `projectDir`. Every adapter except OpenClaw returned a path shared across all projects on the machine. Two terminals open in different repos read each other's auto-memory `.md` files via `searchAutoMemory()`, then those notes contaminated `ctx_search` timeline results.

Reported by @raremalo with a concrete repro: project-A's npm-outdated audit notes appeared in `ctx_search` from inside project-B. Verified architecturally — auto-memory was the only leak path. `ContentStore` and `searchEvents` SQL are already project-scoped via `hashProjectDirCanonical`.

## Fix

`HookAdapter.getMemoryDir(projectDir?)` now scopes by `hashProjectDirCanonical(projectDir)` when supplied. Same canonicalization as `ContentStore` and `searchEvents` — all three project-scoped surfaces share one identity. `searchAutoMemory` passes `projectDir` through; the adapterless legacy fallback applies the same hash directly so the contract holds at both call sites.

Backwards compat: `getMemoryDir()` without `projectDir` keeps the unscoped path so external adapter consumers don't break.

## Scope

5 src files, 4 test files, 291 insertions. Deliberately NOT included:
- Migration fallback that kept reading the legacy unscoped path (would void the fix).
- Claude Code re-route to `<configDir>/projects/<encoded>/memory` (couples us to runtime path ajeno).

## Tests

- `tests/adapters/base-adapter-memory.test.ts` — 5 tests pinning the scoping contract (deterministic, distinct, env-aware, backwards-compat unscoped).
- `tests/core/auto-memory-adapter.test.ts` — leak canary (projectA marker invisible from projectB) + positive control.
- `tests/core/auto-memory-config-dir.test.ts` — pins that the old unscoped `CLAUDE_CONFIG_DIR` path is no longer surfaced.
- `tests/core/search.test.ts` — fixture writes to both scoped + unscoped paths to cover both call shapes.

Targeted suite: 181/181 green. Typecheck clean. Pre-existing failures verified against pristine `next`.

Closes #663.

### DIFF
--- a/src/adapters/base.ts
+++ b/src/adapters/base.ts
@@ -38,6 +38,7 @@
 import { join, resolve } from "node:path";
 import { accessSync, copyFileSync, constants, mkdirSync } from "node:fs";
 import { homedir } from "node:os";
+import { hashProjectDirCanonical } from "../session/db.js";
 
 /**
  * Universal storage-root override. Returns the resolved absolute path when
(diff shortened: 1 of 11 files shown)


mksglu:
Hi @sebastianbreguel CI has error.

mksglu:
Hi @sebastianbreguel. Let's resolve conflict please.

Issue #648, opened 2026-05-20
[FEATURE]: `ctx_fetch_and_index` cache TTL is global — add per-call `ttl` parameter


jetnet:
## Summary

`ctx_fetch_and_index` caches fetched content for 24 hours (`FETCH_TTL_MS = 24 * 60 * 60 * 1000`, `server.ts:2475`). This TTL is global and hardcoded. Different sources have fundamentally different freshness needs:

- A static API reference or README: fine to cache for days.
- A CI dashboard, a changelog, a live issue tracker: stale within an hour.
- An offline / slow-network setup: should survive multiple session restarts without re-fetching.

The tool already has a per-call escape hatch — `force: true` bypasses the cache entirely. A `ttl` parameter is the natural companion: override the cache window for *this* source without affecting others.

## Proposed API

```
ctx_fetch_and_index({
  url: "https://example.com/changelog",
  ttl: 3_600_000   // 1 hour for this source; omit to use the 24 h default
})
```

- `ttl: 0` means never cache (equivalent to `force: true`).
- `ttl` absent means use the global default (24 h, unchanged).
- `ttl` is stored alongside the source metadata so re-fetches in the same session respect it.

## Implementation sketch

1. Add optional `ttl?: number` to the `ctx_fetch_and_index` input schema.
2. Pass it into `fetchOne` / the cache-check path in `server.ts` alongside `force`.
3. Replace the single `ageMs < FETCH_TTL_MS` guard with `ageMs < (opts.ttl ?? FETCH_TTL_MS)`.

No change to the default behaviour. `force: true` continues to work as today.

## Verified against

`mksglu/context-mode` `main` at `1.0.142`, `src/server.ts:2475, 2635-2641`.

baqdauletd:
Hi, I’d like to start working on this issue as suggested above.

Will implement TTL support in both forms:

1. A top-level `ttl` option for the whole `ctx_fetch_and_index` call:

```ts
ctx_fetch_and_index({
  requests: [
    { url: "https://example.com/docs" },
    { url: "https://example.com/changelog" }
  ],
  ttl: 60 * 60 * 1000
})
```

2. A per-request `ttl` option inside `requests`, so different URLs in the same batch can have different cache freshness rules:

```ts
ctx_fetch_and_index({
  requests: [
    {
      url: "https://github.com/mksglu/context-mode/issues/xxx",
      ttl: 5 * 60 * 1000
    },
    {
      url: "https://docs.example.com/api",
[...]

PR #666, opened 2026-05-22
feat(fetch): add per-call cache ttl

NgoQuocViet2001:
## Summary
- Add an optional `ttl` parameter to `ctx_fetch_and_index` so callers can override the cache freshness window per call.
- Treat `ttl: 0` as a cache bypass, matching the existing `force: true` behavior without changing the global default.
- Surface the effective TTL in cached responses and add regression coverage for the schema + cache guard.

## Why
`ctx_fetch_and_index` currently uses the global 24h freshness window for every source. That makes live sources such as changelogs or dashboards stale unless callers always force a refetch, while static docs may still benefit from caching. This keeps the existing default but lets callers tune freshness when needed.

Refs #648.

## Validation
- `pnpm exec vitest run tests/core/server.test.ts -t "ctx_fetch_and_index batch refactor"`
- `pnpm run typecheck`
- `git diff --check`

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2525,8 +2525,19 @@ main();
 const FETCH_TTL_MS = 24 * 60 * 60 * 1000; // 24 hours
 const FETCH_PREVIEW_LIMIT = 3072;
 
+function formatFetchTtl(ttlMs: number): string {
+  if (ttlMs === 0) return "0ms";
+  const day = 24 * 60 * 60 * 1000;
+  const hour = 60 * 60 * 1000;
+  const minute = 60 * 1000;
+  if (ttlMs % day === 0) return `${ttlMs / day}d`;
+  if (ttlMs % hour === 0) return `${ttlMs / hour}h`;
+  if (ttlMs % minute === 0) return `${ttlMs / minute}m`;
+  return `${ttlMs}ms`;
+}
+
 type FetchOneResult =
-  | { kind: "cached"; label: string; chunkCount: number; estimatedBytes: number; ageStr: string }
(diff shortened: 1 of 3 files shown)


mksglu:
@NgoQuocViet2001 Ci has error

NgoQuocViet2001:
Thanks for the heads up. I pushed `6262d49` to fix the Windows CI failure.

Root cause: the failing test used `spawnSync("npm", ...)`; on Windows CI `npm` is a `.cmd` shim, so the spawn returned `ENOENT`/`status=null`. The test now invokes `npm pack --dry-run --json` through `cmd.exe` on Windows and keeps the direct `npm` invocation on Unix.

Validation:
- `pnpm exec vitest run tests/scripts/asymmetric-drift-assert.test.ts -t "npm pack dry-run contains"`
- `pnpm exec vitest run tests/scripts/asymmetric-drift-assert.test.ts`
- `pnpm run typecheck`
- `git diff --check`

The refreshed PR checks are now green, including `test (windows-latest)`.

NgoQuocViet2001:
I resolved the merge conflict by merging the latest `upstream/next` (`dee946c`) into this branch and keeping the Windows-safe `npm pack --dry-run --json` test path.

Local validation after resolving:
- `pnpm exec vitest run tests/scripts/asymmetric-drift-assert.test.ts`
- `pnpm exec vitest run tests/core/server.test.ts -t "ctx_fetch_and_index batch refactor"`
- `pnpm run typecheck`
- `git diff --check`

The refreshed CI is running now. Could you take another look and let me know if this looks OK?

NgoQuocViet2001:
Update: the refreshed checks are all green now (5/5), including `test (windows-latest)`.

PR #686, opened 2026-05-24
fix(codex): report standalone doctor status accurately

NgoQuocViet2001:
## What / Why / How

`context-mode doctor` currently hard-codes the Codex adapter's installed version as `not installed`, so Codex users see `Codex CLI: not installed — using standalone MCP mode` even when `codex --version` resolves successfully. I reproduced this on Windows with `CONTEXT_MODE_PLATFORM=codex`: the machine had `codex-cli 0.132.0`, but the doctor version section still looked like the Codex CLI itself was missing.

This PR separates the two states:

- adds a Codex diagnostic that probes `codex --version` (`cmd.exe /c` on Windows so npm `.cmd` shims resolve, `execFileSync` on POSIX)
- returns a `standalone` sentinel for Codex because it has no platform-owned plugin version to compare against the `context-mode` npm package
- teaches `doctor` to render that sentinel as `standalone MCP mode — no platform plugin version to compare`
- adds adapter/CLI tests so this does not regress back to the misleading `not installed` path

This is intentionally diagnostic-only: no hook config, MCP registration, bundle, or install flow changes.

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [x] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

- [x] `pnpm exec vitest run tests/adapters/codex.test.ts tests/core/cli.test.ts` — 221 tests passed
- [x] `pnpm run typecheck`
- [x] `git diff --check`
- [x] `CONTEXT_MODE_PLATFORM=codex pnpm exec tsx src/cli.ts doctor` on Windows confirmed:
  - `Codex CLI binary: PASS — codex --version resolved to codex-cli 0.132.0`
  - `Codex CLI: standalone MCP mode — no platform plugin version to compare`

## Duplicate check

- Open PRs checked: only #685 is open and it touches stats aggregation, not Codex doctor/version diagnostics.
- Searched prior PRs/issues for `Codex CLI: not installed`, `Codex CLI binary`, `standalone MCP mode`, and `codex --version doctor`; no active or merged PR appears to cover this exact diagnostic split.

## Checklist

- [ ] `npm test` passes

### DIFF
--- a/src/adapters/codex/index.ts
+++ b/src/adapters/codex/index.ts
@@ -14,6 +14,7 @@
  * Track: https://github.com/openai/codex/issues/18491
  */
 
+import { execFileSync } from "node:child_process";
 import {
   readFileSync,
   writeFileSync,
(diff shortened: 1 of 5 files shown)


mksglu:
Hi @NgoQuocViet2001 CI has error

NgoQuocViet2001:
Thanks for catching and fixing the CI issue. I missed that the new runtime `Codex CLI binary` probe would be included in the existing all-pass hook-config assertion on runners without `codex` installed. Your fix to keep that probe separate from the hook-config checks is exactly right.

I'll be more careful to run the broader adapter suite and watch the first CI pass next time. Appreciate the quick merge.

NgoQuocViet2001:
@mksglu quick follow-up: thanks again for catching and fixing the CI issue. I missed the CI-runner path where `codex` is not installed, so the new runtime probe could warn while the hook-config test expected every diagnostic to pass. Your separation of the runtime probe from the hook-config assertion is the right fix.

PR #689, opened 2026-05-24
fix(doctor): surface missing launch files when the integrity helper itself is absent

### LINKED ISSUE #688: ctx_upgrade can leave an incomplete plugin-cache install (missing start.mjs / server bundle); ctx_doctor masks it as "integrity helper unavailable"
## Summary
A `ctx_upgrade` can leave the Claude Code plugin cache for the new version **partially populated** — including the MCP **launch entrypoint** `start.mjs` and `server.bundle.mjs`. Claude Code's launch command is `node ${CLAUDE_PLUGIN_ROOT}/start.mjs`, so a missing `start.mjs` makes the MCP server fail to start (persisting across restarts). Worse, `ctx_doctor` **masks** this: because the integrity helper `scripts/plugin-cache-integrity.mjs` is itself shipped in `package.json files[]`, an interrupted swap can drop it too — and `checkPluginCacheIntegritySync` then reports only `integrity helper unavailable: Cannot find module …`, hiding the missing launch files.

## Environment
- Claude Code plugin install, macOS arm64, Node 26 (Bun runtime)
- Upgrade path: v1.0.121 → v1.0.148 via the `ctx_upgrade` MCP tool

## What happened
1. `ctx_upgrade` ran; the MCP server disconnected and stayed down across Claude Code restarts.
2. The active plugin cache `~/.claude/plugins/cache/context-mode/context-mode/1.0.148/` was missing: `start.mjs`, `server.bundle.mjs`, `scripts/postinstall.mjs`, `scripts/heal-better-sqlite3.mjs`, `scripts/heal-installed-plugins.mjs`, `scripts/plugin-cache-integrity.mjs`, and several `build/` files. (The old `…/1.0.121/` dir was marked `.orphaned_at` but complete.)
3. `ctx_doctor` reported `Plugin cache integrity: FAIL — integrity helper unavailable: Cannot find module …/scripts/plugin-cache-integrity.mjs`. After restoring just the helper, it then said `missing: server.bundle.mjs`. So the failure mode masked its own scope.

## Root cause
The atomic swap populated the active version dir only partially, omitting the launch entrypoint. Because the integrity helper was among the missing files, the doctor couldn't run the full `files[]`-derived check and fell back to an opaque "helper unavailable" — never naming the missing launch files that actually stop the server.

## Secondary (separate, not the disconnect cause)
The npm global symlink `/opt/homebrew/lib/node_modules/context-mode` was left pointing at the orphaned `…/1.0.121` dir. This affects only the standalone `context-mode` CLI, not the Claude Code MCP launch (which uses the plugin-cache `start.mjs`). `ctx_doctor` prints `npm (MCP): PASS — v1.0.148` from the running CLI's own version and never validates the symlink target.

## Workaround
- Restored the missing runtime files into the plugin cache from a verified-complete same-version install → MCP server launches; `ctx_doctor` all-OK.
- `npm install -g context-mode@latest` to replace the stale CLI symlink.

## Suggested fixes
- **Make plugin-cache population atomic + verified against `package.json files[]` before switching the active version / killing the old server.** A partial copy that omits `start.mjs` or the server bundle should fail and roll back, never leave a half-populated active dir.
- **Don't let a missing integrity helper mask missing core files.** The doctor should fall back to a dependency-free check of the launch entrypoint (`start.mjs`, `server.bundle.mjs`/`build/server.js`) when the helper itself can't load. → addressed by the linked PR.
- `ctx_doctor` should also resolve+validate the npm global symlink target (flag when it points at an orphaned/non-current version).

PR (against `next`): #<PR> — adds the dependency-free launch-file fallback so the doctor surfaces the real breakage.

ken-jo:
## What
When an interrupted `/ctx-upgrade` leaves the plugin cache partially populated, `scripts/plugin-cache-integrity.mjs` (the integrity helper, shipped in `package.json files[]`) can itself be among the missing files. `checkPluginCacheIntegritySync` then returns only `integrity helper unavailable: …`, **masking** that the MCP launch entrypoint (`start.mjs` / server bundle) is also gone — the actual reason `node ${CLAUDE_PLUGIN_ROOT}/start.mjs` fails and the MCP server never starts.

## Change
- Add `findMissingLaunchFiles(pluginRoot)` — a dependency-free (fs-only) check of:
  - `start.mjs` — the plugin `command` target (`.claude-plugin/plugin.json`), **no fallback**.
  - the server — `server.bundle.mjs`, falling back to `build/server.js` (so no false positive); reported only when **both** are absent.
- In `checkPluginCacheIntegritySync`, when the integrity helper can't load, call `findMissingLaunchFiles` and surface the missing launch files + the reinstall command, instead of stopping at the opaque "helper unavailable".

No behavior change when the helper loads normally.

## Reproduce + verify
Before vs after, on a simulated partial-install (empty) cache dir:
```
BEFORE → [FAIL] Plugin cache integrity: integrity helper unavailable: Cannot find module …/plugin-cache-integrity.mjs
AFTER  → [FAIL] Plugin cache integrity: partial install — critical launch files missing: start.mjs, server.bundle.mjs (or build/server.js); the MCP server cannot start. Reinstall: npm install -g context-mode@latest
```

- New `tests/util/plugin-cache-launch-files.test.ts` (5 cases): present / missing start.mjs / build/server.js fallback / both-server-absent / fully-empty.
- `npm test` (pretest build = tsc + bundle + assert-bundle + assert-asymmetric-drift): **148 files, 3663 passed, 24 skipped**. `tsc --noEmit` clean.

## Notes
- Bundles (`server.bundle.mjs`/`cli.bundle.mjs`) intentionally **not** committed — regenerated by the `bundle` CI workflow.
- Targets `next`.

Closes #688


mksglu:
Thanks @ken-jo — well-diagnosed bug (real partial-install masking case I could reproduce locally) and a minimal fs-only fix that closes it cleanly. Pushed one small follow-up commit (`db57551`) folding the 5 `findMissingLaunchFiles` tests from `tests/util/plugin-cache-launch-files.test.ts` into the existing `tests/core/cli.test.ts` plugin-cache-integrity describe block per CONTRIBUTING.md L282 (no new test files — extend the file that covers the same domain). All 5 cases still GREEN. The production fix is yours — only the test home moved. Merging to `next` for v1.0.149.

PR #695, opened 2026-05-24
feat(session): persist /goal directives across compaction and resume

### LINKED ISSUE #694: Persist /goal directives across compaction and resume (only a coarse intent mode is kept today)
## Summary
A user's stated **session goal is not preserved** across compaction or resume. When a `/goal <text>` directive (or a `goal:` / `objective:` statement) is submitted, `extractIntent` captures only a **coarse mode** — `"investigate"` or `"implement"` — and **discards the goal text**. The resume snapshot then restores only `<intent mode="implement"/>`. So after `/compact` or `--resume`, the model no longer knows *what* it was working toward; only *that* it was implementing something.

## Repro
1. In a session, state a goal, e.g. `/goal migrate the auth module to OAuth2 (PKCE) and keep tests green`.
2. Search the captured session memory for that text (`ctx_search`). The directive text is not found as a retrievable goal.
3. Inspect a resume snapshot: it contains `<intent mode="implement"/>` but no record of the actual objective.

Root: `src/session/extract.ts::extractIntent` returns `{ category: "intent", data: mode }` (mode only); `src/session/snapshot.ts::buildIntentSection` renders only the mode. There is no first-class "goal" event, and the goal text never reaches the snapshot.

## Why it matters
The whole value of session continuity is that the resuming model can pick up where it left off. Restoring "mode = implement" without the objective means the model has to ask the user to restate the goal — exactly the re-explaining that context-mode exists to avoid.

## Proposed fix (PR against `next`: #<PR>)
- Capture the goal **text** as a first-class `goal` event (priority 1) when the message is a `/goal <text>` command or an explicit `goal:` / `objective:` marker.
- Restore the most recent goal verbatim in a `<session_goal>` section placed **first** in the resume snapshot (right after `how_to_search`), with an instruction to keep working toward it and not ask the user to restate it.
- `intent` stays unchanged (still the coarse mode); its label is corrected to "Session intent" since it never held the goal text.

End-to-end verification in the PR: a fresh agent given **only** the restored snapshot correctly identified the goal and derived a sensible next step toward it, with no prompting.

## Note
This captures from whatever prompt text reaches the `UserPromptSubmit` hook. If a host intercepts `/goal` without firing `UserPromptSubmit`, the `goal:` / `objective:` markers still work; worth confirming hook visibility per host.

ken-jo:
## What
A `/goal <text>` directive, or explicit `goal:` / `objective:` marker, was not preserved across compaction or resume. `extractIntent` kept only the coarse mode (`investigate` / `implement`) and discarded the actual objective, so restored context could say only `<intent mode="implement"/>` without saying what to implement.

This PR makes the stated session goal first-class session memory so a resumed model can continue toward the user's objective without asking the user to restate it.

## Change
- **Goal extraction** (`src/session/extract.ts`): add `extractGoal()` and wire it into `extractUserEvents()`. It captures `/goal <text>`, `goal: <text>`, and `objective: <text>` as a `goal` event.
- **Durability under the session cap**: store `goal` at `priority: 4`, matching the existing `SessionDB` eviction contract where higher numeric priority survives before lower-priority events.
- **Snapshot restore** (`src/session/snapshot.ts`): render the latest goal in a `<session_goal>` section immediately after `<how_to_search>`, before files, tasks, and `<intent>`.
- **Analytics copy** (`src/session/analytics.ts`): label `goal` as "Session goal" and relabel `intent` as "Session intent".
- **Codex compact coverage** (`tests/hooks/codex-goal-compact.test.ts`): run the actual Codex hook sequence `UserPromptSubmit -> PreCompact -> SessionStart(source=compact)` and assert the slash `/goal` objective is restored in `hookSpecificOutput.additionalContext`.
- **Bundles**: regenerate affected runtime bundles (`server.bundle.mjs`, `cli.bundle.mjs`, `hooks/session-extract.bundle.mjs`, `hooks/session-snapshot.bundle.mjs`).

No behavior change to the coarse `intent` mode.

## Codex Review Follow-Up
Codex review found that the original `priority: 1` goal event would be evicted first at the per-session 1000-event cap because `SessionDB` removes the lowest numeric priority first. That would break the core continuity claim for long Codex sessions before compaction.

Follow-up changes:
- changed goal priority to `4`;
- added an event-cap regression proving a goal survives after 1000 lower-priority events;
- added Codex hook lifecycle coverage proving slash `/goal` survives through compact restore when the prompt text reaches `UserPromptSubmit`.

### Local
- `npx vitest run tests/hooks/codex-goal-compact.test.ts tests/session/session-goal.test.ts tests/session/session-db.test.ts tests/session/session-snapshot.test.ts --reporter=dot` — 4 files / 132 tests passed.
- `npx tsc --noEmit` — passed.
- `npm run build` — passed (`tsc`, `bundle`, `assert-bundle`, `assert-asymmetric-drift`).
- `git diff --check` — passed.

### Independent Codex Subagent Verification
A separate Codex subagent verified the goal behavior without editing files:
- session goal/unit + DB/snapshot suite: 4 files / 295 tests passed;
- bundle probes confirmed `/goal`, `goal:`, and `Objective:` create `goal` events with `priority: 4`;
- bundle DB probe confirmed a goal survives the 1000-event cap;
- bundle snapshot probe confirmed latest goal wins and `<session_goal>` is ordered before files/intent.

A second Codex subagent independently exercised the Codex hook compact path:
- ran `tests/hooks/codex-goal-compact.test.ts` successfully;
- manually simulated `hooks/codex/userpromptsubmit.mjs -> hooks/codex/precompact.mjs -> hooks/codex/sessionstart.mjs(source=compact)` with a fake `CODEX_HOME`;
- confirmed a stored goal row with `{ category: "goal", priority: 4 }`;
- confirmed restored `additionalContext` contains `<session_goal>`, contains the objective, and orders goal before `<intent>`.

### Remote CI
- `test (ubuntu-latest)` — success.
- `test (macos-latest)` — success.
- `test (windows-latest)` — success.
- `openclaw-e2e (ubuntu-latest)` — success.
- `openclaw-e2e (macos-latest)` — success.

## Constraints / Assumptions
- This preserves `/goal ...` only when the host passes the submitted slash-command text to `UserPromptSubmit`. If a host consumes the slash command before hooks see it, context-mode cannot infer that exact text from this path.
- `goal:` and `objective:` remain explicit marker forms for hosts or workflows where slash-command text is not available to hooks.
- The PR stores the latest stated goal. It does not implement goal completion semantics, multiple named goals, or goal history UX.
- `priority: 4` intentionally follows today's `SessionDB` eviction contract. A future cleanup can introduce named priority constants to remove numeric ambiguity.
- The PR targets `next` and remains a draft while maintainers review the new behavior and test shape.

## Future Work
- Consider extending context-mode with a more explicit goal feature surface, instead of treating goals only as parsed prompt text. For example, a future context-mode command/API could set, inspect, clear, or update the current session goal directly while reusing the same `<session_goal>` restore path.
- Consider replacing numeric priorities with named constants (`CRITICAL`, `HIGH`, etc.) so future continuity-critical events cannot accidentally use the wrong eviction priority.

Closes #694


mksglu:
Reach out to me via DM I would like to invite you our private engineering group on WhatsApp. Send me e-mail. @ken-jo

ken-jo:
Hi @mksglu, Claude Code and Codex verification are both complete on this PR.

src/session/extract.ts
```diff
@@ -1052,6 +1052,35 @@ function extractIntent(message: string): SessionEvent[] {
...
+ * Captures the user's stated objective so it survives compaction and resume —
+ * unlike `intent`, which stores only the coarse mode (investigate/implement)
+ * and discards the goal text. Triggered by the `/goal <text>` command or an
+ * explicit `goal:` / `objective:` marker, so the FULL goal text is preserved
+ * (priority 4 = critical in the DB eviction contract) and restored at the top
+ * of the resume snapshot.
+ * Without this, a `/goal` directive is lost across compaction/resume.
+ */
+const GOAL_DIRECTIVE_PATTERN =
+  /^(?:\/goal\s+|(?:goal|objective)\s*:\s*)(.+)$/is;
```
mksglu, line 1067: What about Chinese/Japan lang?

mksglu (changes requested):

src/session/extract.ts
ken-jo: Good question — let me clarify the design intent and what I measured.

**Marker keywords (`goal` / `objective`):** these mirror only the `goal` keyword convention that Codex and Claude Code already support. I deliberately did *not* add per-language translations (目标 / 目標 / 목표 / …) as marker keywords — that felt like overfitting (which languages? simplified vs. traditional? where do we stop?).

**Goal body (the text after the marker):** this already supports any language — the capture group `(.+)` with the `s` flag is Unicode-agnostic. I verified it:

| Input | Result |
|---|---|
| `/goal 中文目标内容` (slash + Chinese text) | ✅ match, captured `中文目标内容` |
| `goal: 日本語の目標` (English keyword + Japanese text) | ✅ match |
| `objective: 한국어 목표` (English keyword + Korean text) | ✅ match |
| `目标: 提高性能` (Chinese marker keyword) | ❌ no match |
| `目標：…` (Japanese keyword + fullwidth colon) | ❌ no match |

So CJK **content** is fully supported today; only the **marker keyword** is English-only. (Note: CJK input also tends to use the fullwidth colon `：` U+FF1A, which the current ASCII-`:` regex wouldn't match either — so marker i18n would mean both keywords *and* colon variants.)

Given that the body is already language-agnostic, do you still want the marker *keywords* themselves to be multilingual?

My suggestion: support only the language-neutral `/goal` slash command for now, and split per-language marker matching into a separate follow-up — implementing it later once the feature has stabilized and seen enough real-world usage, rather than guessing the language set up front. Happy to go either way — just let me know your preference.

PR #698, opened 2026-05-24
fix: query_scope for batch search, throttle visibility, cache observability, insight clarity

### LINKED ISSUE #696: [Bug]: ctx_batch_execute.queries silently searches only batch output, not global index — API parameter name misleads
### Platform

Claude Code

### context-mode version

1.0.151

```json
{
  "context_mode": {
    "version": "1.0.151",
    "binary_path": "C:\\Users\\<user>\\AppData\\Roaming\\npm\\node_modules\\context-mode\\",
    "available_commands": ["doctor", "upgrade", "hook", "statusline"]
  },
  "runtime": {
    "node": "v25.9.0",
    "bun": "1.3.14",
    "npm": "11.12.1"
  },
  "os": {
    "platform": "Windows 11 Pro",
    "version": "10.0.26200",
    "arch": "x64"
  },
  "doctor_output": {
    "platform_detection": "Claude Code (high confidence)",
    "storage_sessions": "C:\\Users\\<user>\\.claude\\context-mode\\sessions",
    "performance": "FAST (Bun detected)",
    "language_coverage": "5/11 (45%) — javascript, shell, typescript, python, perl",
    "server_test": "PASS",
    "hooks_configured": ["PreToolUse", "SessionStart"],
    "fts5_sqlite": "PASS",
    "plugin_cache_integrity": "PASS"
  }
}
```

```text
Pre-condition: at least one document is already indexed in the global
session store via ctx_index (any file or directory containing the word
"frontmatter" somewhere in its text).

Step 1 — confirm the global index has matches via ctx_search:
   ctx_search(query: "validação de frontmatter")
   → expected: ≥1 result returned

Step 2 — issue an identical search via ctx_batch_execute.queries:
   ctx_batch_execute(
     commands: [{ label: "noop", command: "echo none" }],
     queries: ["validação de frontmatter"]
   )
   → observed: "No matching sections found"
   → expected (based on parameter name): same hit as Step 1
```

```text
Tool: ctx_batch_execute
Input:
  {
    "commands": [{ "label": "noop", "command": "echo none" }],
    "queries": ["validação de frontmatter"]
  }
Output:
  Executed 1 commands (1 lines, 0.0KB). Indexed 1 sections. Searched 1 queries.

  ## validação de frontmatter → No matching sections found.

  > Tip: Results are scoped to this batch only. To search across all
  > indexed sources, use ctx_search(queries: [...]).

By contrast, the same query via ctx_search:
  Tool: ctx_search
  Input: { "query": "validação de frontmatter" }
  Output: 1+ matching sections from the global indexed corpus

The contradiction is not a runtime error — it's a silent semantic mismatch.
The tool's own tip confirms the scope: `queries` inside ctx_batch_execute
searches ONLY the output of the commands in that same batch, never the
persistent global index that ctx_index/ctx_search operate on.

Reproduced across 4 independent client/model combinations during a
multi-IDE comparison test on v1.0.146 AND v1.0.148 (Windows 11):

┌────────────────────────────────────┬──────────────────────────────────┐
│ Client                             │ batch hits vs sequential hits    │
├────────────────────────────────────┼──────────────────────────────────┤
│ Claude Code (Opus 4.7)             │ batch: 0 / sequential: 8         │
│ Codex CLI (GPT-5)                  │ batch: 0 / sequential: 8         │
│ Antigravity (Gemini 3.5 Flash)     │ batch: 0 / sequential: 8         │
│ Antigravity (Gemini 3.1 Pro)       │ batch: 0 / sequential: 8         │
└────────────────────────────────────┴──────────────────────────────────┘

Same 5 queries, same indexed corpus, identical mismatch in 4/4 runs.
v1.0.149 → v1.0.151 changelog confirms this code path was not touched,
so the behavior persists on the current latest.
```

### Steps to reproduce (REQUIRED)

1. Install / upgrade:
   npm install -g context-mode@1.0.151

2. Configure MCP in any client (Claude Code example):
   claude mcp add --transport stdio --scope user context-mode -- bun x context-mode

3. From the agent session, seed the global index with any file:
   ctx_index(path: "<path-to-any-markdown-file-mentioning-'frontmatter'>")

4. Verify the global index responds:
   ctx_search(query: "validação de frontmatter")
   → expect ≥1 result

5. Issue the same search inside ctx_batch_execute:
   ctx_batch_execute(
     commands: [{ label: "noop", command: "echo none" }],
     queries: ["validação de frontmatter"]
   )
   → observe "No matching sections found"
   → tip in output explicitly confirms: "Results are scoped to this batch only"

### What have you tried to fix it?

1. Verified across 4 client/model combinations on Windows 11 with v1.0.146,
   v1.0.148, and confirmed unchanged in the v1.0.149 → v1.0.151 changelog.
   Identical 0-hit result in all 4 batch runs vs 8-hit baseline via
   ctx_search. The behavior is server-side and consistent — not a client
   adapter issue.

2. Read the tool's own response tip carefully:
     "Results are scoped to this batch only. To search across all indexed
      sources, use ctx_search(queries: [...])."
   This confirms the maintainer is aware the parameter has a counter-
   intuitive scope. The issue is that the parameter name (`queries`)
   gives no signal that its scope differs from the top-level ctx_search.

3. Workaround currently in use: instead of ctx_batch_execute, call
   ctx_search(queries: [...]) which accepts an array and runs all queries
   against the global index in a single round-trip. Measured ~7× faster
   than sequential ctx_search calls (the original motivation for trying
   ctx_batch_execute). This works perfectly — the ctx_batch_execute path
   is functionally redundant for the array-of-queries use case.

4. Investigation hints (from external reading):
   - ctx_batch_execute is genuinely useful for the OTHER half of its
     contract: batch SHELL command execution with shared timing/output
     aggregation.
   - The `queries` parameter inside it appears designed to search the
     just-produced command output (e.g. grep-after-shell pattern).
   - The naming collision between top-level `queries` (global) and
     batch-internal `queries` (local) is what trips users up.

5. Impact observed in our multi-IDE benchmark: the 4 LLMs we tested all
   reached for ctx_batch_execute first when asked to "run the same 5
   searches via batch instead of sequentially" — none anticipated the
   scope shift. All 4 only recovered after reading the tip in the tool's
   response. The naming is materially confusing.

Expected behavior:

   Either rename the parameter to signal local scope, OR add a `scope`
   flag, OR add a schema-level note. Any of:

   Option A — rename:
     ctx_batch_execute({
       commands: [...],
       batch_local_queries: [...]   // or `output_queries` / `local_queries`
     })

   Option B — explicit scope flag:
     ctx_batch_execute({
       commands: [...],
       queries: [...],
       query_scope: "batch" | "global"   // default "batch" preserves
                                          // current behavior
     })

   Option C — schema description fix only (lowest effort):
     Update the tool's JSON-Schema `description` for `queries` to:
     "Search queries executed against the output produced by THIS batch's
      commands only. To search the global indexed corpus, use ctx_search."

   Bonus consideration: the tip currently appears only AFTER a 0-hit
   response. Surfacing the same warning in the schema description
   prevents the agent from making the wrong call in the first place.

Proposed PR shape (happy to implement if approach is approved):

   If Option C (schema-only) is preferred — trivial: edit the tool
   definition string in the MCP server registration. ~5 LOC.

   If Option B (scope flag) is preferred — opt-in `query_scope` param
   defaulting to "batch" preserves backwards-compat. Branch in the
   handler routes "global" to the same code path ctx_search uses. ~30 LOC.

   Tests: add cases for (a) batch scope returns 0 on globally-indexed-only
   content (current behavior), (b) global scope returns same as ctx_search
   for identical queries, (c) schema description includes the scope warning.

Happy to PR against `next` — let me know which option fits your roadmap.

### Pre-submission checklist

### Operating System

macOS (Apple Silicon)

### JS Runtime

Bun 1.3.14

### LINKED ISSUE #697: [Bug]: 3 remaining issues from cross-platform audit — ctx_insight Q&A, ctx_stats observability, ctx_search throttling (companion to #687)
### Platform

Claude Code

### context-mode version

1.0.151

```json
{
  "context_mode": {
    "version": "1.0.151",
    "binary_path": "C:\\Users\\<user>\\AppData\\Roaming\\npm\\node_modules\\context-mode\\",
    "available_commands": ["doctor", "upgrade", "hook", "statusline"]
  },
  "runtime": {
    "node": "v25.9.0",
    "bun": "1.3.14",
    "npm": "11.12.1"
  },
  "os": {
    "platform": "Windows 11 Pro",
    "version": "10.0.26200",
    "arch": "x64"
  },
  "doctor_output": {
    "platform_detection": "Claude Code (high confidence)",
    "performance": "FAST (Bun detected)",
    "fts5_sqlite": "PASS",
    "plugin_cache_integrity": "PASS"
  }
}
```

```text
This is a bundled report covering 3 distinct issues discovered during the
same cross-platform audit that produced #687 (resolved) and the
ctx_batch_execute scope issue (opened separately). Each issue below has
its own reproducer prompt and was verified across 4 client/model
combinations on v1.0.146, v1.0.148, and v1.0.151 (current).

────────────────────────────────────────────────────────────────────────
ISSUE A — ctx_insight ignores the `query` parameter (dashboard launcher only)
────────────────────────────────────────────────────────────────────────

Prompt that triggered:
   ctx_insight(query: "What is the practical difference between skill A
                       and skill B in this codebase?")

────────────────────────────────────────────────────────────────────────
ISSUE B — ctx_stats omits cache hit rate, chunk count, last_indexed_at
────────────────────────────────────────────────────────────────────────

Prompt that triggered:
   ctx_stats()
   (run twice — once cold, once after ~6 ctx_search calls — to observe
    what the second call reports vs what's available to compute.)

────────────────────────────────────────────────────────────────────────
ISSUE C — ctx_search applies undocumented progressive throttling (3→2→1)
────────────────────────────────────────────────────────────────────────

Prompt that triggered:
   Issue ctx_search sequentially 5+ times within ~60 seconds, each call
   passing `limit: 3`. Observe the response `limit` shrinking after the
   2nd call.
```

```text
═══════════════════════════════════════════════════════════════════════
CONTEXT: This bundles 3 issues from an audit that produced 5 findings.
#687 (ctx_index directory)        — RESOLVED in v1.0.149, thank you
#<bug2 number> (ctx_batch_execute) — opened separately yesterday
THIS issue                         — the 3 remaining items below

Bundling these instead of opening 3 separate issues to respect your
time as a solo maintainer. Happy to split into 3 if you prefer — just
say the word.
═══════════════════════════════════════════════════════════════════════

──────────────────────────────────────────────
ISSUE A — ctx_insight ignores `query` parameter
──────────────────────────────────────────────

Tool: ctx_insight
Input: { "query": "<any natural language question>" }
Output:
  - Copying source files...
  - Source files copied.
  - Installing dependencies (first run, ~30s)...
  - Dependencies installed.
  - Building dashboard...
  - Dashboard running at http://localhost:4747
  PID: <pid>

The `query` parameter is accepted by the JSON-Schema but completely
ignored at runtime. The tool only launches a React analytics dashboard
on port 4747. Naming + parameter shape strongly imply Q&A / RAG-style
synthesis over the indexed corpus — that expectation is wrong.

Cross-platform verification (4/4 ignored the parameter identically):
  Claude Code (Opus 4.7)         — dashboard launched, query ignored
  Codex CLI (GPT-5)              — dashboard launched, query ignored
  Antigravity (Gemini 3.5 Flash) — dashboard launched, query ignored
  Antigravity (Gemini 3.1 Pro)   — dashboard launched, query ignored

All 4 LLM agents reached for ctx_insight when asked to "synthesize
an answer about X from the indexed corpus". All 4 had to fall back
to ctx_search after seeing the dashboard URL.

──────────────────────────────────────────────
ISSUE B — ctx_stats observability gap
──────────────────────────────────────────────

Tool: ctx_stats
Input: {} (no args)
Output (current):
  - events count
  - sessions count
  - tokens saved (lifetime + session)
  - % reduction
  - $ saved estimate

What's NOT exposed:
  - cache_hit_rate (nominal hit ratio per query)
  - cache_hits_count / cache_miss_count breakdown
  - total_chunks_indexed (chunks, not MB)
  - last_indexed_at timestamp
  - per-query cache state on subsequent identical queries

Effect: optimizing usage is guesswork. We can only INFER cache
behavior by diffing `tokens saved` between two ctx_stats calls,
which is indirect and noisy. During our v1.0.151 verification of
#687, all 3 IDEs (Opus, GPT-5, Gemini) independently noted: when
calling ctx_index on the same path twice, the response does not
indicate whether the second call hit cache or re-indexed (Codex
measured a ~3% time delta, Opus and Gemini saw none).

Cross-platform verification — all 4 clients had to invent the same
workaround (subtract Fase 1 stats from Fase 6 stats to estimate
session savings, with no way to know hit rate per query).

──────────────────────────────────────────────
ISSUE C — ctx_search progressive throttling (undocumented)
──────────────────────────────────────────────

Tool: ctx_search
Pattern: 5+ sequential calls within a short window (~60s), each
with `limit: 3`.

Observed warning (literal, surfaces in response from call #3 onwards):
  ⚠ search call #N/8 in this window. Results limited to X/query.

Progression we measured:
  Call 1: returns up to 3 results (limit honored)
  Call 2: returns up to 3 results (limit honored)
  Call 3: returns up to 2 results (limit silently cut)
  Call 4: returns up to 1 result (limit silently cut)
  Call 5+: returns up to 1 result (limit silently cut)

This is reasonable rate-limiting behavior, but:
  1. It's nowhere in the tool's JSON-Schema description.
  2. Agents have no signal until the warning appears in a response
     (already too late — the truncated result already shipped).
  3. The exact window size (8 calls) and reset period are not
     documented anywhere we could find.
  4. Workaround (ctx_search(queries: [...])) is excellent but only
     discoverable by reading the warning text mid-session.

Cross-platform verification (3/4 hit the throttle on the same
test):
  Claude Code (Opus 4.7)         — throttle observed at call #3
  Antigravity (Gemini 3.1 Pro)   — throttle observed at call #3
  Codex CLI (GPT-5)              — did NOT hit throttle (suspected
                                   internal cache absorbed repeats)
  Antigravity (Gemini 3.5 Flash) — warning surfaced but reported
                                   less explicitly than the others

The v1.0.149 → v1.0.151 changelog confirms none of these 3 code
paths were touched, so all behaviors persist on current latest.
```

### Steps to reproduce (REQUIRED)

Prerequisites (same for all 3 issues):
  npm install -g context-mode@1.0.151
  Configure MCP in any client (Claude Code example):
    claude mcp add --transport stdio --scope user context-mode -- bun x context-mode

──── ISSUE A — ctx_insight ignores query ────
1. From an agent session, call:
     ctx_insight(query: "any natural language question")
2. Observe: dashboard URL is returned; `query` is ignored.
3. Confirm by re-running with no `query` param — identical output:
     ctx_insight()

──── ISSUE B — ctx_stats missing fields ────
1. Call ctx_stats() — record output JSON shape.
2. Run any 5 ctx_search calls.
3. Call ctx_stats() again — note that the only things that changed
   are `events`, `tokens saved`, `% reduction`, `$ saved`. There is
   no `cache_hit_rate`, `total_chunks`, or `last_indexed_at` field
   in either snapshot.

──── ISSUE C — ctx_search throttling ────
1. Call ctx_search 5 times in quick succession, each with
   different queries, each passing `limit: 3`:
     ctx_search(query: "alpha", limit: 3)
     ctx_search(query: "beta",  limit: 3)
     ctx_search(query: "gamma", limit: 3)
     ctx_search(query: "delta", limit: 3)
     ctx_search(query: "epsilon", limit: 3)
2. Observe response from call #3 onwards:
     ⚠ search call #N/8 in this window. Results limited to X/query.
3. Confirm the array-form bypasses the throttle:
     ctx_search(queries: ["alpha", "beta", "gamma", "delta", "epsilon"], limit: 3)
   → returns 3 per query, no throttle warning.

### What have you tried to fix it?

Context: these 3 issues were uncovered during the same cross-platform
audit that produced #687 (ctx_index directory, RESOLVED in v1.0.149)
and the ctx_batch_execute scope issue opened yesterday. The full audit
ran the same 6-phase test prompt across Claude Code (Opus 4.7), Codex
CLI (GPT-5), and Antigravity (Gemini 3.5 Flash + Gemini 3.1 Pro) on
v1.0.146 and v1.0.148, then re-verified bug #687 on v1.0.151. Total
of 5 structural findings; first 4 in flight, this issue covers the
remaining 3. We bundled them to avoid spamming the queue.

──── ISSUE A — ctx_insight ────

Investigation:
- Tool surface (`context-mode --help`) shows no `--mode` flag or
  subcommand for Q&A vs dashboard distinction.
- Output text suggests this tool was designed exclusively as a
  dashboard launcher; the `query` slot may be an artifact of an
  earlier design or a placeholder.
- The dashboard at :4747 is genuinely useful and works perfectly —
  the bug is purely about the MCP contract.

Proposed fixes (in order of effort):

  Option A1 — Tool description fix (5 LOC):
    Update the JSON-Schema description of `ctx_insight.query` to:
    "(Currently unused — reserved for future Q&A synthesis.) Calling
     ctx_insight launches the analytics dashboard at localhost:4747.
     For Q&A over the indexed corpus, use ctx_search."

  Option A2 — Rename for clarity (deprecate gracefully):
    Add ctx_dashboard() as an alias, keep ctx_insight as a deprecated
    alias for one minor version, then remove the `query` param.

  Option A3 — Implement Q&A (larger scope):
    Wire `query` to a RAG-over-indexed-content path. Probably too
    ambitious for a single PR; not recommended unless you already
    have plans here.

──── ISSUE B — ctx_stats observability ────

Investigation:
- The data needed to compute cache_hit_rate clearly exists internally
  (since `tokens saved` and `% reduction` are accurate end-to-end).
- The chunk count and last_indexed_at presumably live in the SQLite
  store already.

Proposed fix (one PR, additive — no breaking changes):

  Extend ctx_stats() output with new fields:
    {
      // ... existing fields ...
      "cache_hit_rate": 0.74,           // nominal: hits / (hits + miss)
      "cache_hits_count": 14,
      "cache_miss_count": 5,
      "total_chunks_indexed": 4216,
      "last_indexed_at": "2026-05-24T14:16:40Z",
      "throttle_window_remaining": 5    // ← also covers Issue C below
    }

  All existing fields preserved. Clients can fall back gracefully if
  fields are absent (the new fields are purely additive observability).

  Bonus: surface the same data on `ctx_doctor` for one-shot inspection
  without a dedicated stats call.

──── ISSUE C — ctx_search throttling ────

Investigation:
- The throttle is clearly intentional (rate-limit / abuse protection).
- The warning string proves the server already knows: "call #N/8 in
  this window". That counter could be exposed to clients upfront.
- The workaround (ctx_search with array of queries) is excellent and
  should be the documented happy path for multi-query workloads.

Proposed fixes (in order of effort):

  Option C1 — Schema description fix (5 LOC):
    Update `ctx_search` JSON-Schema description to:
    "Searches the global indexed corpus. Note: this server throttles
     sequential calls — after the 2nd call in a short window, the
     effective limit is progressively reduced (3 → 2 → 1 per query).
     For multiple searches, prefer passing `queries: [...]` as an
     array, which bypasses the throttle counter."

  Option C2 — Expose counter in every response (low effort):
    Add `throttle_remaining: N` to every ctx_search response so the
    agent can pace itself. Pairs naturally with Issue B's exposed
    fields.

  Option C3 — Configurable throttle (later):
    Env var `CONTEXT_MODE_THROTTLE_WINDOW=8` or similar. Lowest
    priority — most users will be fine with the default once it's
    documented.

──── Combined PR offer ────

If approach is approved, I can PR all 3 fixes in a single branch:
  - Option A1 (insight description fix): ~5 LOC
  - Issue B observability fields:        ~40 LOC + tests
  - Option C1 + C2 (search description + throttle counter): ~20 LOC

Total: ~65 LOC + tests, single PR against `next`, fully backwards-
compatible (everything is additive or descriptive). Happy to
implement if you confirm the approach.

Apologies for the volume of issues this week — these all surfaced
in the same audit and we batched them as much as we could (one
PR, three logical fixes). Let me know if you'd prefer them split.

### Pre-submission checklist

### Operating System

macOS (Apple Silicon)

### JS Runtime

Bun 1.3.14

matiasduartee:
## What / Why / How

Four follow-up fixes surfaced during cross-platform validation (Claude Code, Codex CLI, Antigravity/Gemini).

**Fixes #696 — `ctx_batch_execute`: search scope leaks across batch items**
Added `query_scope: "batch" | "global"` param (default `"batch"`). When `"batch"`, each item searches only sources indexed in that call — existing behaviour. When `"global"`, searches the full store, enabling mixed pipelines where one batch item needs cross-project context.

**Fixes #697 (1/3) — `ctx_insight` description implies Q&A capability it doesn't have**
Added a one-line disclaimer to the tool description clarifying it launches the dashboard rather than answering natural-language queries. Prevents misrouted calls that silently return nothing.

**Fixes #697 (2/3) — `ctx_search` throttle counter invisible until soft cap**
Throttle status now surfaces on every response (not only after the limit fires), so callers can pace themselves proactively. Throttle policy and all three env-var overrides (`CONTEXT_MODE_SEARCH_WINDOW_MS`, `CONTEXT_MODE_SEARCH_MAX_RESULTS_AFTER`, `CONTEXT_MODE_SEARCH_BLOCK_AFTER`) are documented in the tool description under a `THROTTLING:` section.

**Fixes #697 (3/3) — `ctx_stats` missing cache miss rate and index state**
`AnalyticsEngine` now tracks `cacheMisses`, computes `hit_rate`, and accepts `IndexState` (total chunks, sources, last indexed). `ContentStore.getIndexState()` provides the aggregate via a single SQL query. `ctx_stats` handler wires these together; `formatReport` renders them in an **Observability** section that is omitted entirely when no data is present (no output regression for clean sessions).

---

## Affected platforms

- [x] Claude Code (macOS)
- [x] Claude Code (Windows)
- [x] Claude Code (Linux)
- [x] Codex CLI (macOS)
- [x] Codex CLI (Windows)
- [x] Codex CLI (Linux)
- [x] Antigravity (macOS)
- [x] Antigravity (Windows)
- [x] Antigravity (Linux)
- [x] Other MCP clients (macOS)
- [x] Other MCP clients (Windows)
- [x] Other MCP clients (Linux)
- [x] Cursor (macOS / Windows / Linux)
- [x] Windsurf (macOS / Windows / Linux)

> `query_scope` and throttle visibility affect all platforms equally. Cache/index observability is additive to `ctx_stats` output.

---

## Test plan

13 new Vitest cases added inside the existing `tests/core/server.test.ts` (no new test files — per CONTRIBUTING.md):

| Group | Cases | What is verified |
|---|---|---|
| `ctx_batch_execute query_scope` | 3 | `"batch"` default, `"global"` override, schema rejects invalid values |
| `ctx_insight description disclaimer` | 1 | Source-grep confirms disclaimer text present |
| `ctx_search throttle visibility` | 3 | Counter present at call #1, soft-cap warning at threshold, THROTTLING section in description |
| `ctx_stats cache + index observability` | 3 | `cacheMisses` increments, `hit_rate` computed correctly, `getIndexState()` aggregate |
| `AnalyticsEngine.queryAll cache` | 3 | `cache.misses`, `cache.hit_rate`, Observability section omitted when empty |

All 13 pass. `npm test` green. `npx tsc --noEmit` clean.

---

## Checklist

- [x] Tests added for all changes
- [x] `npm test` passes
- [x] TypeScript type-check passes (`npx tsc --noEmit`)
- [x] Docs / description strings updated where relevant
- [x] No Windows path regressions
- [x] Targets `next` branch

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -690,6 +690,7 @@ const sessionStats = {
   bytesIndexed: 0,
   bytesSandboxed: 0, // network I/O consumed inside sandbox (never enters context)
   cacheHits: 0,
+  cacheMisses: 0, // ctx_fetch_and_index calls that bypassed the TTL cache
   cacheBytesSaved: 0, // bytes avoided by TTL cache hits
   sessionStart: Date.now(),
 };
@@ -1235,14 +1236,23 @@ export function extractSnippet(
   return parts.join("\n\n");
 }
 
+export type BatchQueryScope = "batch" | "global";
+
 export function formatBatchQueryResults(
   store: ContentStore,
   queries: string[],
   source: string,
   maxOutput = 80 * 1024,
+  scope: BatchQueryScope = "batch",
 ): string[] {
   const sections: string[] = [];
   let outputSize = 0;
+
+  // When scope is "global", searchWithFallback receives `undefined` for the
+  // source filter, which makes it query the entire persistent index instead
+  // of only the chunks just produced by this batch's commands. Default
+  // remains "batch" to preserve the historical behavior.
+  const searchSource = scope === "global" ? undefined : source;
 
   for (const query of queries) {
     if (outputSize > maxOutput) {
@@ -1250,7 +1260,7 @@ export function formatBatchQueryResults(
       continue;
     }
 
-    const results = store.searchWithFallback(query, 3, source, undefined, "exact");
+    const results = store.searchWithFallback(query, 3, searchSource, undefined, "exact");
     sections.push(`## ${query}`);
     sections.push("");
     if (results.length > 0) {
@@ -1268,7 +1278,11 @@ export function formatBatchQueryResults(
     sections.push("");
   }
 
-  sections.push(`\n> **Tip:** Results are scoped to this batch only. To search across all indexed sources, use \`ctx_search(queries: [...])\`.`);
+  if (scope === "global") {
+    sections.push(`\n> **Scope:** Queries searched the entire persistent index (query_scope: "global").`);
+  } else {
+    sections.push(`\n> **Tip:** Results are scoped to this batch only. To search across all indexed sources, use \`ctx_search(queries: [...])\` or call ctx_batch_execute with \`query_scope: "global"\`.`);
+  }
 
   return sections;
 }
@@ -2166,12 +2180,24 @@ EXAMPLE: ctx_index(path: "/path/to/large-spec.md", source: "openapi-v2-spec")`,
 // Tool: search — progressive throttling
 // ─────────────────────────────────────────────────────────
 
-// Track search calls per 60-second window for progressive throttling
+// Track search calls per N-second window for progressive throttling.
+// Defaults preserve the historical behavior (60s window, soft-cap at 3
+// calls, hard-block at 8). All three thresholds are overridable via env
+// vars so users can loosen or tighten the policy without forking. Invalid
+// values (non-positive numbers, NaN) fall back to the default to avoid
+// silently disabling the protection.
+function readPositiveEnv(name: string, defaultValue: number): number {
+  const raw = process.env[name];
+  if (!raw) return defaultValue;
+  const parsed = Number(raw);
+  return Number.isFinite(parsed) && parsed > 0 ? parsed : defaultValue;
+}
+
 let searchCallCount = 0;
 let searchWindowStart = Date.now();
-const SEARCH_WINDOW_MS = 60_000;
-const SEARCH_MAX_RESULTS_AFTER = 3; // after 3 calls: 1 result per query
-const SEARCH_BLOCK_AFTER = 8; // after 8 calls: refuse, demand batching
+const SEARCH_WINDOW_MS = readPositiveEnv("CONTEXT_MODE_SEARCH_WINDOW_MS", 60_000);
+const SEARCH_MAX_RESULTS_AFTER = readPositiveEnv("CONTEXT_MODE_SEARCH_MAX_RESULTS_AFTER", 3); // after N calls: 1 result per query
+const SEARCH_BLOCK_AFTER = readPositiveEnv("CONTEXT_MODE_SEARCH_BLOCK_AFTER", 8); // after N calls: refuse, demand batching
 
 /**
  * Defensive coercion: parse stringified JSON arrays, AND lift a bare
@@ -2259,7 +2285,7 @@ WHEN NOT:
   - You have one ad-hoc question against data that is not in the knowledge base — answer it inline by running code in the sandbox tool; one round-trip instead of capture-then-query
 
 RETURNS:
-  Per-query ranked sections with window-extracted snippets. Use 2-4 specific technical terms per query. Common session-memory source labels: \`decision\` (user corrections / preferences), \`error\` and \`error-resolution\` (past failures + their fixes), \`blocker\`, \`plan\`, \`user-prompt\`, \`rejected-approach\`, \`compaction\` (post-compact session guide). See ctx_stats for live category counts.
+  Per-query ranked sections with window-extracted snippets. Use 2-4 specific technical terms per query. Common session-memory source labels: \`decision\` (user corrections / preferences), \`error\` and \`error-resolution\` (past failures + their fixes), \`blocker\`, \`plan\`, \`user-prompt\`, \`rejected-approach\`, \`compaction\` (post-compact session guide). See ctx_stats for live category counts. Each response carries a throttle counter (call #N/M in the rolling time window); results taper toward the soft cap and calls block after the hard cap. Tune via CONTEXT_MODE_SEARCH_WINDOW_MS, CONTEXT_MODE_SEARCH_MAX_RESULTS_AFTER, CONTEXT_MODE_SEARCH_BLOCK_AFTER.
 
 EXAMPLE: ctx_search(queries: ["root cause", "proposed fix", "test coverage"], source: "issue-#683")
 EXAMPLE: ctx_search(queries: ["what did we decide about caching"], source: "decision", sort: "timeline")
@@ -2443,11 +2469,20 @@ EXAMPLE: ctx_search(queries: ["last user prompt", "active skills", "open blocker
         output = `> Auto-refreshed ${store.lastRefreshCount} stale source${store.lastRefreshCount > 1 ? "s" : ""} (file changed since indexing).\n\n` + output;
       }
 
-      // Add throttle warning after threshold
+      // Throttle counter — always surfaced so agents can pace themselves
+      // proactively instead of discovering the limit only after results are
+      // already truncated. Soft warning after SEARCH_MAX_RESULTS_AFTER calls;
+      // gentle informational line before that.
+      const throttleRemaining = Math.max(0, SEARCH_BLOCK_AFTER - searchCallCount);
+      const softCapRemaining = Math.max(0, SEARCH_MAX_RESULTS_AFTER - searchCallCount);
       if (searchCallCount >= SEARCH_MAX_RESULTS_AFTER) {
         output += `\n\n⚠ search call #${searchCallCount}/${SEARCH_BLOCK_AFTER} in this window. ` +
-          `Results limited to ${effectiveLimit}/query. ` +
+          `Results limited to ${effectiveLimit}/query. ${throttleRemaining} call(s) remaining before block. ` +
           `Batch queries: ctx_search(queries: ["q1","q2","q3"]) or use ctx_batch_execute.`;
+      } else {
+        output += `\n\n> Throttle: call #${searchCallCount}/${SEARCH_BLOCK_AFTER} in this window. ` +
+          `${softCapRemaining} call(s) before soft cap. ` +
+          `Prefer ctx_search(queries: [...]) array form for multi-query workloads — it counts as a single call.`;
       }
 
       if (output.trim().length === 0) {
@@ -3221,6 +3256,10 @@ EXAMPLE: ctx_fetch_and_index(
         finalized.push({ kind: "fetch_error", url: v.url, error: v.error, reason: v.reason });
       } else {
         // Serial FTS5 write here — no parallel store.index calls.
+        // Cache miss: the URL was not in the TTL window so we paid the
+        // network round-trip + re-indexed. Counted here so ctx_stats can
+        // report nominal cache_hit_rate alongside the existing hit metrics.
+        sessionStats.cacheMisses++;
         finalized.push({ kind: "fetched", indexed: indexFetched(v) });
       }
     }
@@ -3407,9 +3446,21 @@ EXAMPLE: ctx_batch_execute(
           ">1 switches to per-command timeouts (no shared budget) and " +
           "individual `(timed out)` blocks instead of cascading skip.",
         ),
+      query_scope: z
+        .enum(["batch", "global"])
+        .optional()
+        .default("batch")
+        .describe(
+          "Scope for `queries` (default: `batch`). " +
+          "`batch` searches ONLY the chunks produced by this batch's commands " +
+          "— useful when you want answers about the just-fetched output. " +
+          "`global` searches the entire persistent index (same scope as ctx_search) " +
+          "— useful when you want the batch commands to enrich context and " +
+          "the queries to also surface related prior knowledge in one round trip.",
+        ),
     }),
   },
-  async ({ commands, queries, timeout, concurrency }) => {
+  async ({ commands, queries, timeout, concurrency, query_scope }) => {
     // Security: check each command against deny patterns
     for (const cmd of commands) {
       const denied = checkDenyPolicy(cmd.command, "batch_execute");
@@ -3472,9 +3523,11 @@ EXAMPLE: ctx_batch_execute(
         sectionTitles.push(s.title);
       }
 
-      // Run all search queries — source scoped only.
-      // Cross-source search remains available via explicit ctx_search().
-      const queryResults = formatBatchQueryResults(store, queries, source);
+      // Run all search queries — default scope is batch-local (legacy behavior).
+      // When the caller passes query_scope: "global", searches reach the entire
+      // persistent index in the same round trip. Cross-source search remains
+      // available via explicit ctx_search() as well.
+      const queryResults = formatBatchQueryResults(store, queries, source, undefined, query_scope);
 
       // Get searchable terms for edge cases where follow-up is needed
       const distinctiveTerms = store.getDistinctiveTerms
@@ -3692,7 +3745,13 @@ server.registerTool(
           }
           // v1.0.117: pass projectDir as cwd so the narrative renderer's
           // "started in <path>" line matches the user's actual project.
-          text = formatReport(report, VERSION, _latestVersion, { lifetime, mcpUsage, multiAdapter, conversation, realBytes, cwd: projectDir });
+          // Snapshot the persistent store so the renderer can show
+          // total_chunks / last_indexed_at without callers having to query
+          // separately. Best-effort — getStore() is process-local and may
+          // be unavailable on cold paths; failures are absorbed.
+          let indexState;
+          try { indexState = getStore().getIndexState(); } catch { /* never block ctx_stats */ }
+          text = formatReport(report, VERSION, _latestVersion, { lifetime, mcpUsage, multiAdapter, conversation, realBytes, indexState, cwd: projectDir });
         } finally {
           sdb.close();
         }
@@ -3707,7 +3766,9 @@ server.registerTool(
         }
         let multiAdapter;
         try { multiAdapter = getMultiAdapterLifetimeStats(); } catch { /* never block ctx_stats */ }
-        text = formatReport(report, VERSION, _latestVersion, { lifetime, multiAdapter });
+        let indexState;
+        try { indexState = getStore().getIndexState(); } catch { /* never block ctx_stats */ }
+        text = formatReport(report, VERSION, _latestVersion, { lifetime, multiAdapter, indexState });
       }
     } catch {
       // Session DB not available or incompatible — build minimal report from runtime stats
@@ -4367,7 +4428,7 @@ server.registerTool(
   {
     title: "Open Insight Dashboard",
     description:
-      "Opens the context-mode Insight dashboard in the browser. " +
+      "Opens the context-mode Insight dashboard in the browser — a dashboard launcher for session analytics; for natural-language queries over indexed content, use ctx_search. " +
       "Shows personal analytics: session activity, tool usage, error rate, " +
       "parallel work patterns, project focus, and actionable insights. " +
       "First run installs dependencies (~30s). Subsequent runs open instantly. " +
--- a/src/session/analytics.ts
+++ b/src/session/analytics.ts
@@ -136,7 +136,19 @@ export interface RuntimeStats {
   calls: Record<string, number>;
   sessionStart: number;
   cacheHits: number;
+  cacheMisses?: number;
   cacheBytesSaved: number;
+}
+
+/**
+ * Index observability snapshot — point-in-time view of the persistent
+ * content store. Optional input to `formatReport` so callers that don't
+ * have store access (or don't want the extra DB hit) can omit it.
+ */
+export interface IndexState {
+  totalChunks: number;
+  totalSources: number;
+  lastIndexedAt?: string; // ISO-8601 when available
 }
 
 // ─────────────────────────────────────────────────────────
@@ -160,6 +172,8 @@ export interface FullReport {
   };
   cache?: {
     hits: number;
+    misses: number;
+    hit_rate: number; // hits / (hits + misses); 0 when both are zero
     bytes_saved: number;
     ttl_hours_left: number;
     total_with_cache: number;
@@ -450,12 +464,21 @@ export class AnalyticsEngine {
 
     // ── Cache ──
     let cache: FullReport["cache"];
-    if (runtimeStats.cacheHits > 0 || runtimeStats.cacheBytesSaved > 0) {
+    const cacheMisses = runtimeStats.cacheMisses ?? 0;
+    if (runtimeStats.cacheHits > 0 || runtimeStats.cacheBytesSaved > 0 || cacheMisses > 0) {
       const totalWithCache = totalProcessed + runtimeStats.cacheBytesSaved;
       const totalSavingsRatio = totalWithCache / Math.max(totalBytesReturned, 1);
       const ttlHoursLeft = Math.max(0, 24 - Math.floor((Date.now() - runtimeStats.sessionStart) / (60 * 60 * 1000)));
+      // hit_rate is the nominal cache effectiveness — the metric ctx_stats
+      // historically inferred-only by diffing tokens_saved snapshots. When
+      // there is no activity we report 0 instead of NaN/undefined so the
+      // renderer stays JSON-safe.
+      const totalLookups = runtimeStats.cacheHits + cacheMisses;
+      const hitRate = totalLookups > 0 ? runtimeStats.cacheHits / totalLookups : 0;
       cache = {
         hits: runtimeStats.cacheHits,
+        misses: cacheMisses,
+        hit_rate: hitRate,
         bytes_saved: runtimeStats.cacheBytesSaved,
         ttl_hours_left: ttlHoursLeft,
         total_with_cache: totalWithCache,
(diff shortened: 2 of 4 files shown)


mksglu:
Hi @matiasduartee Let's follow ADR 
- https://github.com/mksglu/context-mode/blob/main/docs/adr/0002-tool-description-style.md
- https://github.com/mksglu/context-mode/blob/main/docs/adr/0003-routing-deny-reasons.md

matiasduartee:
Hi @mksglu 

Done — conformance fixes pushed in the latest commit:

- ctx_insight: removed the NOT-negation, rewrote as a positive one-liner per the exempt-tool rule
- ctx_search: folded THROTTLING: into RETURNS: (no carve-out approved for that section), removed it from ALLOWED_EXTRA_SECTIONS in the contract test

ADR-0003 — no changes needed on our side; we don't touch routing.mjs.

mksglu:
> * THROTTLING:

Why we need THROTTLING @matiasduartee

matiasduartee:
> > * THROTTLING:
> 
> Why we need THROTTLING @matiasduartee

@mksglu The runtime counter messages (added per the original bug report) already surface the throttle state on every response. The mention in RETURNS: was a secondary heads-up so agents could opt for the array-query form from the first call rather than discovering the limit reactively.

If you prefer the runtime messages carry the full signal and the description stays clean, happy to drop it from RETURNS: entirely.

mksglu:
> > > * THROTTLING:
> > 
> > 
> > Why we need THROTTLING @matiasduartee
> 
> @mksglu The runtime counter messages (added per the original bug report) already surface the throttle state on every response. The mention in RETURNS: was a secondary heads-up so agents could opt for the array-query form from the first call rather than discovering the limit reactively.
> 
> If you prefer the runtime messages carry the full signal and the description stays clean, happy to drop it from RETURNS: entirely.

We need test that manually

matiasduartee:
@mksglu Tested manually with the swapped bundle (7 consecutive ctx_search calls):

- Calls ＃1 and ＃2 (before soft cap): counter text is appended to output and returned, but the search results were long enough that the counter at the end was cut off in the MCP display. The logic is confirmed correct in the bundle — both branches do h +=.
- Call ＃3 onward: ⚠ search call ＃3/8 in this window. Results limited to 2/query. — appeared consistently.
- The gentle counter for calls 1-2 is present in the response payload but not visible when results are long. I can move the counter to the top of the output instead of the bottom — that would make it always visible regardless of result length, and would be a one-line fix.

Alternatively, if you'd prefer to drop the pre-cap counter entirely and keep only the existing soft-cap ⚠ warning, I'm happy to revert just that part.

src/server.ts
```diff
@@ -2401,11 +2427,19 @@ EXAMPLE: ctx_search(queries: ["last user prompt", "active skills", "open blocker
...
+      // gentle informational line before that.
+      const throttleRemaining = Math.max(0, SEARCH_BLOCK_AFTER - searchCallCount);
       if (searchCallCount >= SEARCH_MAX_RESULTS_AFTER) {
         output += `\n\n⚠ search call #${searchCallCount}/${SEARCH_BLOCK_AFTER} in this window. ` +
-          `Results limited to ${effectiveLimit}/query. ` +
+          `Results limited to ${effectiveLimit}/query. ${throttleRemaining} call(s) remaining before block. ` +
           `Batch queries: ctx_search(queries: ["q1","q2","q3"]) or use ctx_batch_execute.`;
+      } else {
+        output += `\n\n> Throttle: call #${searchCallCount}/${SEARCH_BLOCK_AFTER} in this window. ` +
+          `${throttleRemaining} call(s) before soft cap. ` +
```
NgoQuocViet2001: Manual test on Windows + Codex confirmed the counter is returned, but this line is counting from `SEARCH_BLOCK_AFTER` while saying `before soft cap`. With the defaults, call #1 prints `7 call(s) before soft cap` even though the soft cap is call #3. Could we either compute a separate `softCapRemaining` for this pre-cap branch, or change the wording to `before block` if the hard-block counter is intentional?

NgoQuocViet2001 (commented):
Manual test notes from Windows + Codex:

- `query_scope` default stays batch-local: a pre-indexed global-only token did not appear in the default batch query result.
- `query_scope: "global"` did find the pre-indexed global content and returned the global scope note.
- `ctx_search` returns the gentle throttle counter for calls #1/#2 and the warning branch from call #3 onward.

One small wording issue inline before merge.

src/server.ts
matiasduartee: Hi @NgoQuocViet2001

Fixed — added softCapRemaining = Math.max(0, SEARCH_MAX_RESULTS_AFTER - searchCallCount) for the pre-cap branch. Call #1 now prints 2 call(s) before soft cap instead of 7. The hard-block counter (throttleRemaining) stays unchanged in the ⚠ branch.

NgoQuocViet2001 (approved):
@matiasduartee Re-tested after the fix to my soft-cap wording note. Looks good from my side.

Checked on Windows:
- The pre-cap throttle counter now reports the soft cap correctly: call #1 shows `2 call(s) before soft cap`, call #2 shows `1 call(s) before soft cap`, and call #3 uses the warning branch.
- `ctx_batch_execute` default scope stays batch-local: it did not return pre-indexed global content and kept the batch-only tip.
- `query_scope: "global"` returned the global indexed content and the global-scope note.

Ran:
- `npm run build`
- `node node_modules/vitest/vitest.mjs run tests/core/server.test.ts -t "ctx_batch_execute query_scope|ctx_insight schema description|ctx_search progressive throttle|ctx_stats cache|AnalyticsEngine.queryAll cache"`
- Full `tests/core/server.test.ts`: 465 passed, 3 local Windows shell/projectRoot failures that also reproduce on `upstream/next`, so I don't think they are from this PR.

matiasduartee:
Rebased onto latest `next` and resolved the conflicts.

- The conflict was a clean 3-way merge: your search/analytics perf work
  (in-place sorts in `analytics.ts` / `store.ts`) and this PR's cache
  observability additions (`cacheMisses`, `hit_rate`, `IndexState`) live in
  separate hunks, so both are preserved.
- Bundles (`server.bundle.mjs`, `cli.bundle.mjs`) rebuilt fresh from the
  merged source; `assert-bundle` and `assert-asymmetric-drift` pass.
- Squashed to 2 commits (feat + bundle rebuild) for a clean diff.
- Test suite: 3695 passing. The 2 failures I see locally are unrelated to
  this PR — `codex/marketplace-layout` needs the generated `plugins/` tree
  (absent in my dev checkout), and `ctx_index #365` depends on
  `IDEA_INITIAL_DIRECTORY` (JetBrains env). Happy to confirm against CI.

NgoQuocViet2001:
@matiasduartee I rechecked after the rebase.

Local validation:
- `npm run typecheck`
- `npm run assert-bundle`
- `npm run assert-asymmetric-drift`
- PR-specific Vitest filter: 13 passed
- Full `tests/core/server.test.ts`: 465 passed, 3 local Windows failures. I re-ran those same 3 tests on `upstream/next` and they fail there too, so I don't think they are from this PR.

One issue I found while checking the new cache/index observability: `formatReport()` appends the `## Observability` block only after the active-session dashboard path, but the normal `ctx_stats` path now passes `conversation`, and the narrative renderer returns early when `conversation.events > 0`. In that path, `cache.hit_rate` / `index.total_chunks` never render even though the handler passes `indexState`.

Could you thread the Observability block into the narrative path too, or move the append before the early returns, and add a test with `conversation.events > 0` plus `indexState`?

CI note: both Actions runs for the current head are `action_required`, with no jobs yet.

matiasduartee:
@NgoQuocViet2001 

Good catch — confirmed and fixed in 7896276 + 1d9db76.

Root cause: `formatReport()` early-returns when `conversation.events > 0`
to emit the 5-section narrative renderer, but the `## Observability` append
lived AFTER that return. So for the normal `ctx_stats` path (handler always
passes both `conversation` and `indexState`), `cache.hit_rate` and
`index.*` never rendered.

Fix: extracted the rendering into a shared `renderObservabilityBlock`
helper called from both paths — narrative path (before the early return)
and legacy path. Block stays omitted when there's no data, preserving
prior behaviour.

Test added: `tests/core/server.test.ts` now has a case with
`conversation.events > 0` + cache hits/misses + `indexState`, asserting
all four observability lines surface in the narrative output. Local run:
3696 passed, 2 failed (same 2 env-specific Windows failures as before,
unrelated).

Thanks for the careful review — this would have silently shipped broken.

mksglu:
Thanks @matiasduartee — implementation is sound, bundling of the three fixes is justified (per #697's own grouping, no need to split). Two small things before merge:

**1. Indentation glitch at `src/server.ts:~3649-3656`** — the new `indexState` block (`let indexState; try { ... } catch {}`) drops from col 10 to col 6 mid-try-block. Lints/runs fine but breaks scope alignment with the surrounding handler.

**2. Branch is conflicting on `next`** after #695, #716, and #741 landed today. Please rebase — bundle files (`cli.bundle.mjs`, `server.bundle.mjs`) will need a clean rebuild against the new bundle state on `next`.

Re-ping after rebase + indent fix.

Issue #604, opened 2026-05-18
hooks.json keeps previous-version directory path after Claude Code plugin auto-update — every hook fails MODULE_NOT_FOUND


jowch:
## Summary

After the context-mode plugin auto-updates via **Claude Code's native plugin manager** (not `/ctx-upgrade`), the new version's cached `hooks/hooks.json` keeps command paths that point at the **previous** version's directory. That directory has been cleaned up, so every hook (`SessionStart`, `PreToolUse`, `PostToolUse`, `UserPromptSubmit`, `PreCompact`) fails with `MODULE_NOT_FOUND` — while `ctx-doctor` still reports every check as PASS.

This is the same class of bug as #411 (`.mcp.json`) and #523 (`.claude-plugin/plugin.json`). #528 fixed the `/ctx-upgrade` **tmpdir** variant of `hooks.json` poisoning, but not the **previous-version-directory** variant produced by Claude Code's own auto-update pipeline.

## Environment

- OS: Linux
- Node: v24.15.0 (nvm-managed)
- Host: Claude Code
- context-mode: 1.0.135 → 1.0.136 (auto-updated by Claude Code's plugin manager, mid-use)

## Symptom

Every hook fails. `~/.claude/plugins/cache/context-mode/context-mode/1.0.136/hooks/hooks.json` contained 13 commands of the form:

```
"/home/<user>/.bun/bin/bun" "/home/<user>/.claude/plugins/cache/context-mode/context-mode/1.0.135/hooks/pretooluse.mjs"
```

Only `1.0.136/` exists on disk — `1.0.135/` was cleaned up after the update. `bun` cannot find the file, so every hook fire is `MODULE_NOT_FOUND`.

`ctx-doctor` does not catch this: it verifies that the *current* version's hook files exist (they do) and that hooks are *configured* — but not that `hooks.json`'s command paths actually point at the current version. So doctor reports a fully green run while every hook is dead.

## Root cause

`normalize-hooks.mjs` rewrites the committed `${CLAUDE_PLUGIN_ROOT}` placeholder into an absolute path on every MCP boot (Linux/Windows). This is a **one-way ratchet**: `needsHookNormalization()` only returns `true` while the `${CLAUDE_PLUGIN_ROOT}` placeholder is present. Once a command path has been made absolute, the placeholder is gone, so the normalizer can never re-point that path again.

[...]

Issue #550, opened 2026-05-13
1.0.123 npm tarball missing server.bundle.mjs and start.mjs — MCP server silently fails on CC plugin installs


rpickmans:
After upgrading to 1.0.123 via `npm i -g context-mode`, Claude Code showed `plugin:context-mode:context-mode: Failed` with nothing useful in the error. Took a while to track down: `~/.claude/plugins/cache/context-mode/context-mode/1.0.123/` is missing `server.bundle.mjs` and `start.mjs`. The `.mcp.json` calls `${CLAUDE_PLUGIN_ROOT}/start.mjs` on startup, so without that file the server dies silently.

The `better-sqlite3` binding was healthy (Node 26, probed fine). The problem was that neither file is in the `files` field in the 1.0.123 `package.json`, so they don't end up in the npm tarball.

I fixed it by symlinking the 1.0.124 global npm install into the plugin cache and updating `installed_plugins.json` to point there. Server came up immediately on the next restart.

Two things that would help: deprecating/yanking 1.0.123 on npm so new installs skip it, and gating the publish on `scripts/assert-bundle.mjs` so the next release can't ship without those files.

(Related to #245 which was the same class of bug — missing files from tarball — but for `scripts/postinstall.mjs`.)

mksglu:
Tarball verified clean — pulled `context-mode@1.0.123` from npm and listed contents:

```
$ tar -tzf context-mode-1.0.123.tgz | grep -E 'start.mjs|server.bundle.mjs'
package/server.bundle.mjs
package/start.mjs
```

Same for v1.0.124 / v1.0.125. `package.json` `files[]` has both entries continuously since v1.0.120, and `scripts/assert-bundle.mjs` is wired into `prepublishOnly` → cannot publish without the bundles. So the npm publish path was correct; whatever happened in your specific cache is local-side and I can't pinpoint the mechanism without your install logs (`~/.claude/logs/*` if Claude Code emits them).

Recommended: `npm install -g context-mode@latest` (v1.0.125) and re-fetch the marketplace plugin via `claude plugins uninstall context-mode@context-mode && claude plugins install context-mode@context-mode`. If the symptom recurs on a clean reinstall, please reopen with `ls -la` of the cache dir and any marketplace fetch logs.

mksglu:
Architectural defense shipped in v1.0.126.

[...]

Issue #181, opened 2026-03-25
Auto-update deletes plugin cache directory used by active session


liby:
## Environment

- **OS**: macOS (Apple Silicon / arm64)
- **Node.js**: v24.14.0
- **Claude Code**: 2.1.81
- **context-mode**: v1.0.51 → v1.0.53 (auto-updated)

## Description

When auto-update runs while a session is active, the old plugin cache directory is deleted but the running session's `CLAUDE_PLUGIN_ROOT` still points to it. Every subsequent hook invocation fails with "hook error" for the rest of that session.

## Timeline (from transcript and debug logs)

```
01:45:05 UTC  Session starts, CLAUDE_PLUGIN_ROOT set to .../1.0.51
01:47:36 UTC  Auto-update creates .../1.0.53 cache directory
01:48:20 UTC  sessionstart-debug.log: Cannot find module 'better-sqlite3' (from 1.0.53)
01:52:46 UTC  installed_plugins.json updated to 1.0.53
              .../1.0.51 directory no longer exists
              All hooks in the active session now fail: "Plugin directory does not exist: .../1.0.51"
```

## Observed behavior

Every tool call in the session shows:

```
PreToolUse:Bash hook error
PostToolUse:Bash hook error
UserPromptSubmit hook error
```

The PreCompact hook surfaces the actual error:

```
PreCompact [node ${CLAUDE_PLUGIN_ROOT}/hooks/precompact.mjs] failed:
Plugin directory does not exist: /Users/bryan/.claude/plugins/cache/context-mode/context-mode/1.0.51
```

New sessions started after the update work correctly.

## Notes

This may be a Claude Code platform issue (plugin manager deleting old cache dirs without checking active sessions) rather than something context-mode can fix. Posting here in case you have ideas for a workaround on the plugin side.

mksglu:
Thanks for the detailed report @liby — the timeline and debug logs made root cause analysis straightforward.

## Root Cause

This was context-mode's bug, not Claude Code's. Context-mode aggressively deleted old version cache directories in two places:

1. **`src/cli.ts`** — the upgrade function ran `rmSync` on every sibling version dir
2. **`hooks/pretooluse.mjs`** — the self-heal block nuked all dirs not in `keepDirs` set on every tool call

Since `CLAUDE_PLUGIN_ROOT` is set once at session start and is immutable mid-session, deleting the old versioned directory breaks all hooks for the rest of that session.

## Fix

[...]

PR #699, opened 2026-05-24
fix(hooks): self-heal partial plugin cache install from marketplace clone

kerneltoast:
## What / Why / How

**What**: A new `hooks/heal-partial-install.mjs` that detects a partial plugin cache install on every session start and re-copies missing files from the marketplace clone Claude Code maintains, plus a back-ported `existsSync` guard on `cli.ts`'s `/ctx-upgrade` swap loop that closes a related latent vector.

**Why** (direct observations from the reproducer machine; everything in this section is fact, not speculation):

The per-version plugin cache dir at `~/.claude/plugins/cache/context-mode/context-mode/1.0.150/` held 73 files of the marketplace's tracked tree. Present: `hooks/`, `.claude-plugin/`, `.cursor-plugin/`, `.pi/`, `web/`, plus a few docs. Absent: `cli.bundle.mjs`, `server.bundle.mjs`, `start.mjs`, `package.json`, `src/`, `bin/`, `scripts/`, `skills/`, `.codex-plugin/`, `.openclaw-plugin/`.

The cache's `.claude-plugin/plugin.json` was a verbatim carry-forward from the user's previous 1.0.146 install, with `mcpServers.command = "/usr/bin/bun"` and `mcpServers.args[0]` referencing the 1.0.146 cache dir (which had been age-gate-cleaned by `hooks/sessionstart.mjs` (#181) and no longer existed). MCP launch ENOENTs every spawn.

None of the existing defenses repairs this:

- `scripts/plugin-cache-integrity.mjs` (#550) exits 2 from `start.mjs`, but `start.mjs` is one of the missing files.
- The #604 stale-cache-version ratchet runs from `start.mjs`'s normalize-hooks call, same boot-time dependency.
- `/ctx-upgrade` needs the `cli.bundle.mjs` the partial install lost, so the user-facing escape hatch is gone too.

So the cache cannot self-recover, and the only available workaround from the user side is `rm -rf <version-dir>` plus a session restart. The user reported hitting this several times across upgrade cycles.

**What I could verify directly about the producing code path** (also fact): Claude Code runs a plugin sync regardless of the marketplace's `autoUpdate` setting. With `"autoUpdate": false` in `~/.claude/settings.json` for the `mksglu/context-mode` marketplace, a read-only `claude plugin list` invocation still created a new per-version cache dir matching the marketplace's current version. So `autoUpdate=false` does not opt out of CC's "ensure cache matches marketplace version" sync.

**What I cannot verify (speculation, flagged explicitly)**: which CC code path produced *this* particular fingerprint (partial copy plus carry-forward `plugin.json`). The reproducer's `/ctx-upgrade` transcripts show this project's own `cli.ts upgrade()` took its "Already on latest" early-return path and the rm/cp loop did not run, so that's ruled out as the producer here. A clean-marketplace `claude plugin list` repro on my machine produced a *complete* cache install, not a partial one, so the specific failure mode needs additional conditions I could not pin down. Three adjacent upstream bug reports describe overlapping failure modes in CC's plugin sync: anthropics/claude-code#48912 (Update fails to materialize newly-added top-level directories), #57645 (`installPath` in `installed_plugins.json` lags cache extraction non-atomically), and #57646 (`gitCommitSha` in the registry never refreshed). I'm **not** asserting any one of those bugs is the exact mechanism in this instance; they are cited as documented adjacent failure modes in the same subsystem.

**How**: The heal module ships in `hooks/` rather than `scripts/` because the failure mode itself can strip `scripts/` from the cache dir; `hooks/` has been intact in every observed instance. Wired from `hooks/sessionstart.mjs` (primary trigger; fires even when MCP can't boot) and from `start.mjs` (belt-and-braces, runs before the Algo-D4 integrity gate so a fixable partial install is repaired rather than just reported).

Detection is a cheap 4-existsSync probe, so the healthy case is effectively free. On a trip, the module reads `package.json files[]` (prefers `pluginRoot`'s, falls back to the marketplace's when `pluginRoot/package.json` is one of the missing files), expands to a concrete file list against the marketplace clone, and copies each entry that's missing in pluginRoot. After files land, it rewrites any carry-forward stale `mcpServers.args[0]` in `plugin.json` to the current pluginRoot, mirroring the #604 ratchet. The module is pure JS (Node built-ins only), never throws, idempotent, and path-traversal-guarded.

The secondary commit closes a separate partial-install vector on the `/ctx-upgrade` path: `cli.ts:1207-1212`'s `try { rmSync; cpSync } catch {}` would swallow `cpSync` failures (or `files[]` entries that don't exist in the cloned source) and leave files deleted without replacement. The safe pattern (`if (!existsSync(from)) continue`) is already used in `server.ts:3820`'s inline-fallback path; this back-ports it.

**Caveat (vN+1 only)**: This lands in `vN+1` and heals broken installs from that point forward. Users currently hitting the failure need to manually `rm -rf` their cache dir once to clear the old state. From then on the heal keeps the install healthy across all future CC sync cycles.

Fixes: this user's recurring partial-install issue. Related upstream (in CC's plugin sync, not this project): anthropics/claude-code#48912, anthropics/claude-code#57645, anthropics/claude-code#57646.

## Affected platforms

- [x] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

The heal probe and copy logic are CC-cache-layout-specific (`<config>/plugins/cache/<owner>/<plugin>/<version>/`). `deriveMarketplaceClonePath` returns `null` for any other layout (npm-global, dev checkout, opencode cache, ...) and the heal becomes a no-op. The secondary `cli.ts` guard is platform-agnostic since the upgrade flow is shared.

## Test plan

26 new tests added to `tests/core/cli.test.ts` (folded into existing partial-install describe block per CONTRIBUTING L282; no new test files). Coverage:

- `isPartialInstall` cheap probe across the launch-critical file matrix (healthy, missing `start.mjs`, missing `package.json`, missing both `cli.bundle.mjs` + `build/cli.js`, accepting `build/cli.js` as fallback, falsy input).
- `deriveMarketplaceClonePath` layout helper across POSIX paths, trailing slashes, Windows backslashes, npm-global, dev checkout, and falsy input.
- `healPartialInstallFromMarketplace` covering: early-return on healthy install, skip when marketplace clone is missing, path-traversal guard when pluginRoot equals marketplace clone path, full restore of missing files, fallback to marketplace's `package.json` when `pluginRoot/package.json` is missing, args[0] rewrite for carry-forward stale paths, args[0] rewrite for `${CLAUDE_PLUGIN_ROOT}` placeholder, idempotent rewrite on healthy `plugin.json`, idempotent across re-runs, skip for non-CC layouts, no-overwrite of pre-existing user files, partial coverage when the marketplace itself is incomplete, and a full reproduction of the v1.0.150 partial-install + carry-forward fingerprint with end-to-end heal verification.
- Wiring assertions on `start.mjs` and `hooks/sessionstart.mjs` that confirm the heal call shape and ordering relative to other operations.
- A guard assertion on `cli.ts` that confirms the `existsSync(from)` check precedes the `rmSync` in the swap loop.

Regression checks executed manually: each fix intentionally disabled produced informative test failures with the corresponding assertion. The wiring tests are anchored on literal import/call shapes (not substring `indexOf` matches) so a renamed file or commented-out call still fails the test cleanly.

`npm test`'s pretest builds tsc + bundles; the heal-related tests pass against the in-tree source without requiring the bundles. `npm run typecheck` is clean.

## Checklist

- [x] Tests added/updated (TDD: red → green; intentional regression breaks verified)
- [x] `npm test` passes (3708 / 3742 green; 34 pre-existing skips, none related)
- [ ] Docs updated if needed (README, platform-support.md) — N/A, no user-facing surface area changes
- [x] No Windows path regressions (`deriveMarketplaceClonePath` tested against backslash-form input; all file operations use `path.join` / `path.resolve`)
- [ ] Targets `next` branch (unless hotfix) — PR opened against this branch; happy to retarget if `next` is preferred

<details>
<summary><strong>Repro state of the original incident, for archive</strong></summary>

`~/.claude/plugins/cache/context-mode/context-mode/1.0.150/.claude-plugin/plugin.json` before heal:

```json
{
  "name": "context-mode",
  "version": "1.0.150",
  ...
  "mcpServers": {
    "context-mode": {
      "command": "/usr/bin/bun",
      "args": [
        "/home/sultan/.claude/plugins/cache/context-mode/context-mode/1.0.146/start.mjs"
      ]
    }
  }
}
```

`1.0.146/` did not exist on disk (cleaned by `sessionstart.mjs` #181 age gate).

`~/.claude/settings.json`:

```json
"marketplaces": {
  "context-mode": {
    "url": "https://github.com/mksglu/context-mode.git",
    "autoUpdate": false
  },
  ...
}
```

Sync trigger reproduction: `claude plugin list` with `autoUpdate: false` created `~/.claude/plugins/cache/context-mode/context-mode/1.0.151/` (this was a *complete* install, so the specific partial-copy failure mode was not reproduced; only the unconditional-sync behavior was).

</details>

### DIFF
--- a/hooks/heal-partial-install.mjs
+++ b/hooks/heal-partial-install.mjs
@@ -1,0 +1,712 @@
+/**
+ * heal-partial-install.mjs - self-heal a partial plugin cache install.
+ *
+ * Failure mode this addresses:
+ *
+ *   The per-version plugin cache dir at
+ *   ~/.claude/plugins/cache/<owner>/<plugin>/<version>/ holds only a
+ *   subset of the published files. start.mjs, cli.bundle.mjs,
+ *   server.bundle.mjs, package.json and several other entries from
+ *   `package.json files[]` may be absent, while hooks/ and
+ *   .claude-plugin/ tend to survive. The cache's
+ *   .claude-plugin/plugin.json can also be a verbatim carry-forward
+ *   from a prior version's install, with mcpServers.args[0] holding
+ *   an absolute path under a since-deleted cache dir; MCP launch
+ *   ENOENTs as a result.
+ *
+ *   Existing defenses don't repair this:
+ *
+ *     - scripts/plugin-cache-integrity.mjs (#550) exits 2 from
+ *       start.mjs, but start.mjs is one of the missing files.
+ *     - The #604 stale-cache-version ratchet in
+ *       hooks/normalize-hooks.mjs runs from start.mjs's
+ *       normalize-hooks call, same boot-time dependency.
+ *     - /ctx-upgrade needs the cli.bundle.mjs the partial install lost.
+ *
+ *   So the cache cannot self-recover, and the only available
+ *   workaround from the user side is `rm -rf <version-dir>` plus a
+ *   session restart.
+ *
+ * What this module does:
+ *
+ *   Detect partial install via a cheap existsSync probe of a few
+ *   launch-critical files. On a trip, re-copy the missing entries
+ *   from the marketplace clone at
+ *   ~/.claude/plugins/marketplaces/<owner>/ (Claude Code's canonical
+ *   source for the per-version cache dir) and rewrite any
+ *   carry-forward stale args[0] in plugin.json to the current
+ *   pluginRoot. Both halves are mechanism-agnostic.
+ *
+ * Placement:
+ *
+ *   Lives in hooks/ rather than scripts/ because the failure mode can
+ *   strip scripts/ from the cache dir. hooks/ has been intact in every
+ *   observed instance, so it's the most reliable available host.
+ *
+ * Scope:
+ *
+ *   CC-only. The failure mode is specific to Claude Code's
+ *   per-version cache layout at
+ *   ~/.claude/plugins/cache/<owner>/<plugin>/<version>/. Other
+ *   clients (Codex, Cursor, OpenCode, Kiro, ...) ship their own
+ *   SessionStart wrappers under hooks/<client>/, none of which call
+ *   this module. The two call sites that exist
+ *   (hooks/sessionstart.mjs and start.mjs) are themselves CC-only
+ *   entry points: sessionstart.mjs is wired only from
+ *   hooks/hooks.json (CC's hook config), and start.mjs is invoked
+ *   only by CC's .claude-plugin/plugin.json mcpServers entry. The
+ *   module also enforces this at runtime: pluginRoot must match the
+ *   CC cache layout (deriveMarketplaceClonePath returns null
+ *   otherwise) or the function short-circuits with
+ *   `skipped: "not-claude-code"`, before any further work runs.
+ *
+ * Contract:
+ *
+ *   - Pure JS, Node built-ins only.
+ *   - Never throws. Returns a structured result; callers log it.
+ *   - Idempotent: the cheap probe makes the healthy case a few
+ *     existsSync calls, not a full files[] expansion.
+ *   - Path-traversal guarded at six layers, symmetric on both ends:
+ *       1. pluginRoot must match the CC cache layout, the derived
+ *          marketplace path must exist and not equal pluginRoot.
+ *       2. files[] entries that resolve outside rootDir are dropped.
+ *       3. Directory walks use lstatSync and skip symlinks, so a
+ *          symlink-to-outside in the marketplace tree can't be
+ *          harvested as a regular file during manifest expansion.
+ *       4. After mkdirSync(dirname(to)), realpathSync(dirname(to)) is
+ *          re-checked against realpathSync(pluginRoot), catching the
+ *          case where a parent component of `to` is already a
+ *          symlink-to-outside that mkdirSync followed.
+ *       5. Just before the destination write, the source is checked
+ *          two ways: lstat(from).isSymbolicLink() drops leaf symlinks
+ *          (fast path), and realpathSync(from) is required to fall
+ *          under realpathSync(marketplaceClonePath). The realpath
+ *          check collapses ancestor symlinks (e.g. a marketplace
+ *          `scripts/` dir that's been swapped for a symlink to
+ *          outside), which the leaf-lstat would miss because the
+ *          leaf itself is a regular file at the symlink's resolved
+ *          target.
+ *       6. Just before the destination write, lstat(to) and unlink
+ *          any pre-existing symlink at `to`. The write itself uses
+ *          writeFileSync with `flag: "wx"` (O_CREAT | O_EXCL), which
+ *          per POSIX open(2) refuses to follow a symlink at the
+ *          final component. That closes the residual race window
+ *          where a same-user attacker re-plants a symlink at `to`
+ *          between our unlink and the open.
+ *   - Source-file reads that drive rewrites are symlink-checked too.
+ *     rewritePluginJsonArgs opens `.claude-plugin/plugin.json` with
+ *     O_NOFOLLOW, so the open(2) call itself fails with ELOOP when
+ *     the path is a symlink. Reading from the returned fd then binds
+ *     subsequent reads to the inode, closing the TOCTOU window that
+ *     a lstat+readFileSync(path) pair would have left open. Without
+ *     this, a same-user-planted redirect would feed attacker JSON to
+ *     the read, and the subsequent atomic rename would replace the
+ *     symlink with a regular file containing attacker-controlled
+ *     mcpServers config.
+ *   - plugin.json rewrites are atomic AND symlink-safe:
+ *     writeFileSync targets a tmp sibling with a 64-bit random
+ *     suffix (unguessable) under O_CREAT | O_EXCL (refuses to follow
+ *     a pre-planted symlink at the tmp path), then renameSync over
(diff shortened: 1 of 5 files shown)


hooks/heal-partial-install.mjs
```diff
@@ -0,0 +1,443 @@
...
+ * (relative to rootDir, OS-native separator). Entries that don't exist on
+ * disk are silently dropped, matching the semantics of
+ * scripts/plugin-cache-integrity.mjs::derivePluginManifest.
+ */
+function expandFilesArray(rootDir, files) {
+  if (!Array.isArray(files)) return [];
+  const out = new Set();
+  for (const entry of files) {
+    if (typeof entry !== "string" || !entry) continue;
+    const abs = join(rootDir, entry);
```
NgoQuocViet2001, line 238: The heal flow works in my manual partial-install repro, but I think we need one more containment guard around `files[]` entries before merge. I tested a malformed marketplace `package.json` with `files: ["../outside.txt", ...]`; `expandFilesArray()` accepts it here, and the later copy loop writes outside `pluginRoot`. Could we reject absolute / `..` entries, or resolve both `from` and `to` and ensure they stay under `marketplaceClonePath` and `pluginRoot` respectively before copying? That keeps a corrupted marketplace clone from turning the self-heal into an out-of-root write.

NgoQuocViet2001 (commented):
Manual Windows + Codex review notes:

- Targeted heal tests passed locally.
- `pnpm run typecheck` and `pnpm run build` passed locally.
- I also reproduced a partial Claude plugin cache layout manually and confirmed the heal restores `start.mjs`, `cli.bundle.mjs`, `server.bundle.mjs`, clears the partial probe, and rewrites stale `args[0]` to the current version.

One path-containment guard concern inline before merge.

hooks/heal-partial-install.mjs
kerneltoast: _Claude (Anthropic, Opus 4.7) authoring this reply on Sultan's behalf._

Thanks for the catch! `..`-escape entries are now rejected in `expandFilesArray` via lexical `resolve(abs).startsWith(rootDir + sep)`, and both `resolve(from)` and `resolve(to)` are re-checked against their roots at the copy site as defense in depth.

Follow-up adversarial review surfaced a few adjacent vectors that landed in the same PR. On the source side we use `lstat(from).isSymbolicLink()` to drop leaf symlinks and `realpathSync(from).startsWith(realpathSync(marketplaceClonePath) + sep)` to collapse ancestor symlinks (e.g. a marketplace `scripts/` dir swapped for a symlink to outside). On the destination side we `lstat(to)` and `unlinkSync` any pre-existing symlink, and the write itself uses `writeFileSync(..., {flag: "wx"})` so a re-planted symlink at `to` after the unlink can't be followed via the open. `plugin.json` reads go through `openSync(O_RDONLY | O_NOFOLLOW)` + `readFileSync(fd)` so a symlink-redirected read can't feed attacker JSON into the args rewrite, and the rewrite itself writes to a 64-bit-random tmp suffix under `flag: "wx"` plus `renameSync` for atomicity.

We also added an explicit `not-claude-code` skip for any `pluginRoot` that doesn't match the CC cache layout, so non-CC clients (Codex, Cursor, OpenCode, Kiro, ...) bail before any FS work runs. The `Contract` block at the top of `hooks/heal-partial-install.mjs` enumerates the layered guards in full.

Test count went from 191 to 198 with regressions for each closed vector.

NgoQuocViet2001 (approved):
@kerneltoast Re-tested after your containment fix. The original `files[]` traversal concern looks addressed to me.

Checked on Windows:
- Normal partial cache repro heals `start.mjs`, `cli.bundle.mjs`, `server.bundle.mjs`, and `package.json`, clears the partial probe, and rewrites stale `args[0]` to the current version.
- Malformed marketplace `package.json` with `files: ["../OUTSIDE.txt", ...]` is rejected without writing outside `pluginRoot`; legitimate entries still heal.
- Non-Claude-Code plugin roots skip with `not-claude-code`.

Ran:
- `npm run build`
- `node node_modules/vitest/vitest.mjs run tests/core/cli.test.ts`

One non-blocking Windows-local note: on my machine without symlink privilege / Developer Mode, 3 symlink-specific tests fail at `symlinkSync(..., ...)` with `EPERM`. The PR's Windows CI is green, so I don't see it as a merge blocker; just something to keep in mind for local Windows dev boxes.

mksglu:
Thanks @kerneltoast. Conflicting on `next` now after your own security PR #716 landed — both touched `src/cli.ts` swap loop.

The conflict is benign: #716 added `refuseSymlinks` filter, #699 adds `existsSync(from)` guard. They're compatible, just need textual reorder. Please rebase on `next`.

Verdict was MERGE_AS_IS — re-ping after rebase. The 712-LoC heal module + 6-layer path-traversal defense in `hooks/heal-partial-install.mjs` is ready as-is; only `src/cli.ts` needs the replay.

Issue #414, opened 2026-05-04
Windows: intermittent UserPromptSubmit/SessionStart hook errors while ctx-doctor passes


Asureu:
Hi, I am seeing an intermittent Windows runtime issue with `context-mode` where Claude Code hook execution fails in real sessions, even though `ctx-doctor` reports a healthy installation.

## Environment

- OS: Windows 10 Pro for Workstations `10.0.19045`
- Claude Code: current desktop/CLI session on Windows
- context-mode: `1.0.107`
- Node.js: `v24.15.0`
- npm: `11.12.1`

## Symptoms

In some Claude Code sessions, I still see these errors:

- `Failed with non-blocking status code: node:internal/modules/cjs/loader:1479`
- `UserPromptSubmit hook error`
- `SessionStart:startup hook error`
- `SessionStart:resume hook error`
- `SessionStart:compact hook error`
- `SessionStart:clear hook error`

A notable pattern:

- these errors appear when **user messages** are submitted
- **agent messages still work**
- the issue is **intermittent across sessions**
- it is possible for one session to behave normally while another still shows hook errors

## Important observation

At the same time, `context-mode doctor` reports a healthy installation:

- `FTS5 / SQLite: PASS`
- `Hook script: PASS`
- `Version: v1.0.107`

So this does **not** look like a simple “plugin is broken everywhere” case. It looks more like a real-session hook/runtime mismatch.

## Related background

Earlier in the same environment, I had a separate Windows install problem where `better-sqlite3` native bindings were missing. That issue was locally repaired, and after repair:

- `better_sqlite3.node` exists
- `ctx-doctor` passes
- `FTS5 / SQLite` is green

So the current report is **not mainly about the initial install failure**. It is about **hook execution still failing in some live sessions after the native module issue is already repaired**.

### 1. The failing stack points at the cache copy of the plugin

The runtime stack in the debug log points to:

- `%USERPROFILE%\\.claude\\plugins\\cache\\context-mode\\context-mode\\1.0.107\\hooks\\sessionstart.mjs`
- `%USERPROFILE%\\.claude\\plugins\\cache\\context-mode\\context-mode\\1.0.107\\hooks\\session-db.bundle.mjs`

This is useful because it suggests the live failure is happening in the cache-installed hook path.

### 2. `sessionstart-debug.log` still captured a `better-sqlite3`-related stack

The tail of the captured log included frames like:

```text
[...]

PR #713, opened 2026-05-25
fix: stop baking stale version paths during upgrade (#711)

### LINKED ISSUE #711: /ctx-upgrade bakes prior-version paths into new install's plugin.json and hooks.json
# /ctx-upgrade leaves prior-version paths baked into the new install's manifests

## Summary

After `/ctx-upgrade` from `v1.0.103` to `v1.0.151`, the new versioned directory contains a `plugin.json` and `hooks/hooks.json` that still reference the **previous** version's directory for `mcpServers.args` and every hook script path. The plugin only keeps working because the prior directory is left in place; deleting it (e.g. `npm cache prune`-style cleanup) breaks the install even though `installed_plugins.json` correctly points at the new version.

## Environment

- Windows 11
- Claude Code (latest as of 2026-05-25)
- context-mode upgraded `1.0.103` → `1.0.151` via `/ctx-upgrade`

## Reproduction

1. Start on `v1.0.103`.
2. Run `/ctx-upgrade`. Upgrade reports success; `ctx_doctor` reports `Version: v1.0.151`.
3. Inspect:

```bash
cat ~/.claude/plugins/installed_plugins.json
# → context-mode installPath: ".../1.0.151"  ✓

cat ~/.claude/plugins/cache/context-mode/context-mode/1.0.151/.claude-plugin/plugin.json
# → "args": [".../context-mode/1.0.103/start.mjs"]   ✗ wrong version

cat ~/.claude/plugins/cache/context-mode/context-mode/1.0.151/hooks/hooks.json
# → every command path references ".../1.0.103/hooks/*.mjs"   ✗ wrong version
```

4. `ctx_doctor` confirms — `Hook script: PASS — .../1.0.103/hooks/pretooluse.mjs` (and four others).

Both `1.0.103/start.mjs` and `1.0.151/start.mjs` exist with identical sizes (22537 bytes here), so the running code is in fact the new version — but the resolution path is through the old directory.

## Expected

Files placed inside the `1.0.151` versioned directory should self-reference that same directory:

- `1.0.151/.claude-plugin/plugin.json` → `args: [".../1.0.151/start.mjs"]`
- `1.0.151/hooks/hooks.json` → every command path → `".../1.0.151/hooks/*.mjs"`

## Why this matters

- Any maintenance step that cleans up old versioned directories will break the install even though the doctor and `installed_plugins.json` look healthy.
- The version mismatch is invisible to most users; only path-level inspection catches it.

## Workaround

Manually edit `1.0.151/.claude-plugin/plugin.json` and `1.0.151/hooks/hooks.json`, replacing `1.0.103` with `1.0.151` everywhere. Edits survive until the next `/ctx-upgrade`.

## Suggested fix

The installer should template the version into both files at extraction time rather than copying a snapshot of the previous version's manifests.

sebastianbreguel:
## Summary

`/ctx-upgrade` called `normalizeHooksOnStartup` during the upgrade flow, which replaced `${CLAUDE_PLUGIN_ROOT}` placeholders with absolute paths using the OLD `pluginRoot`. When Claude Code copied files to a new versioned cache directory, the baked-in paths pointed at the wrong (stale) version directory.

Fixes #711.

### Origin — `normalizeHooksOnStartup`

- **Introduced in** `622b82d` by **Mert Koseoglu** (2026-05-02)
- **PR:** `fix(windows): normalize hooks.json placeholders on startup (#378)`
- **Original intent:** Windows-specific fix. `${CLAUDE_PLUGIN_ROOT}` placeholders in `hooks.json` / `plugin.json` couldn't be resolved at runtime on Windows (MSYS path mangling, bare `node` not on PATH in Git Bash). The function replaces placeholders with absolute paths on every MCP boot. Designed to be idempotent — only rewrites when placeholders are present.

### When it broke — upgrade wiring

- **Commit** `13d1342` by **Ahmet Selcuk Ozyurt** (2026-05-11)
- **PR:** `fix(ctx-upgrade): stop baking tmpdir path into hooks.json (#528, Windows hotfix)`
- This wired `normalizeHooksOnStartup` into `src/cli.ts`'s `upgrade()` function to fix tmpdir paths leaking into hooks.json during upgrade.
- **The bug:** calling it during upgrade bakes absolute paths pointing to the **current** versioned cache directory (e.g. `.../1.0.103/`). When Claude Code later copies files to a **new** version directory (`.../1.0.151/`), those baked paths become stale and point nowhere.

## Fix approach

Two-pronged:

1. **Removed the `normalizeHooksOnStartup` call from `upgrade()`** in `src/cli.ts`. Fresh clone files ship with `${CLAUDE_PLUGIN_ROOT}` placeholders — the portable form. `start.mjs` normalizes at next MCP boot with the correct `__dirname`.

2. **Broadened the heal regex** in `scripts/heal-installed-plugins.mjs`. Both `healPluginJsonMcpServers` and `healMcpJsonArgs` previously only matched tmpdir paths (`context-mode-upgrade-<digits>`). Now they match **any absolute path ending in `start.mjs`** via `/[/\\]start\.mjs$/`, catching both tmpdir paths and stale versioned cache-dir paths. Added early-return when args already equal the placeholder.

## Files changed

| File | What changed |
|---|---|
| `src/cli.ts` | Removed `normalizeHooksOnStartup` call from `upgrade()` |
| `scripts/heal-installed-plugins.mjs` | Broadened regex from tmpdir-only to any-absolute-path; removed `TMPDIR_UPGRADE_RE`; added placeholder early-return; updated JSDoc |
| `tests/util/heal-installed-plugins.test.ts` | Added 8 asymmetric-drift assertion tests for stale versioned cache-dir paths |
| `cli.bundle.mjs` | Rebuilt bundle |

## Test plan

- [x] All existing tests pass (`npm test`)
- [x] Typecheck passes (`npm run typecheck`)
- [x] New tests cover stale versioned-dir drift shape (`heal-json-assertion`, `cli-upgrade-verification`)
- [x] All 8 asymmetric-drift assertion tests pass
- [ ] Manual: upgrade on Windows from an older version, verify `plugin.json` and `hooks.json` in the new versioned dir use `${CLAUDE_PLUGIN_ROOT}` placeholders (not absolute paths to the old version)

### DIFF
--- a/hooks/normalize-hooks.mjs
+++ b/hooks/normalize-hooks.mjs
@@ -219,23 +219,35 @@ export function normalizePluginJson(content, nodePath, pluginRoot) {
 }
 
 /**
- * Apply normalization to hooks.json and plugin.json on startup.
+ * Apply normalization to hooks/hooks.json ONLY (not plugin.json).
+ *
+ * Why a narrow variant exists (#711 + #414 / #528):
+ *   - plugin.json is read by Claude Code's plugin manager and carried forward
+ *     into NEW versioned cache dirs on auto-update. Baking absolute paths into
+ *     it during /ctx-upgrade poisons the next version (#711).
+ *   - hooks/hooks.json lives in the per-version dir and is read by the SAME
+ *     Node process that needs to spawn a child. On Windows + Git Bash, Claude
+ *     Code fires SessionStart/PreToolUse BEFORE MCP boot — the unresolved
+ *     `${CLAUDE_PLUGIN_ROOT}` placeholder yields MODULE_NOT_FOUND for the
+ *     first hook fire after /ctx-upgrade (#414).
+ *
+ * So /ctx-upgrade calls THIS narrow function (hooks.json only) to close the
+ * Windows first-hook-fire window without re-introducing #711.
  *
  * Options:
- *   - pluginRoot: absolute path to plugin install dir (e.g. __dirname of start.mjs)
+ *   - pluginRoot: absolute path to plugin install dir
  *   - nodePath:   process.execPath
  *   - platform:   process.platform ("win32" and "linux" trigger a write)
  *
  * Best-effort — never throws.
  */
-export function normalizeHooksOnStartup({ pluginRoot, nodePath, platform }) {
+export function normalizeHooksJsonOnly({ pluginRoot, nodePath, platform }) {
   // Normalize on Windows (MSYS path mangling, #369/#372/#378) and Linux
   // (bare `node` not in PATH when invoked via /bin/sh, e.g. nvm users).
   // macOS ships a system node so bare `node` resolves reliably there.
   if (platform !== "win32" && platform !== "linux") return;
   if (!pluginRoot || !nodePath) return;
 
-  // hooks/hooks.json
   try {
     const hooksPath = resolve(pluginRoot, "hooks", "hooks.json");
     if (existsSync(hooksPath)) {
@@ -250,6 +262,26 @@ export function normalizeHooksOnStartup({ pluginRoot, nodePath, platform }) {
   } catch {
     /* best effort */
   }
+}
+
+/**
+ * Apply normalization to hooks.json and plugin.json on startup.
+ *
+ * Options:
+ *   - pluginRoot: absolute path to plugin install dir (e.g. __dirname of start.mjs)
+ *   - nodePath:   process.execPath
+ *   - platform:   process.platform ("win32" and "linux" trigger a write)
+ *
+ * Best-effort — never throws.
+ */
+export function normalizeHooksOnStartup({ pluginRoot, nodePath, platform }) {
+  // Delegate the hooks.json branch to the narrow helper so /ctx-upgrade and
+  // boot share one implementation. plugin.json normalization stays here —
+  // start.mjs and postinstall still need it; /ctx-upgrade must NOT (#711).
+  normalizeHooksJsonOnly({ pluginRoot, nodePath, platform });
+
+  if (platform !== "win32" && platform !== "linux") return;
+  if (!pluginRoot || !nodePath) return;
 
   // .claude-plugin/plugin.json
   try {
--- a/scripts/heal-installed-plugins.mjs
+++ b/scripts/heal-installed-plugins.mjs
@@ -205,8 +205,6 @@ export function healSettingsEnabledPlugins({ settingsPath, pluginKey }) {
 //   - `src/cli.ts` upgrade() (post-bump)
 // ─────────────────────────────────────────────────────────────────────────
 
-/** Matches `<sep>context-mode-upgrade-<digits><sep>`. OS-agnostic. */
-const TMPDIR_UPGRADE_RE = /[/\\]context-mode-upgrade-\d+[/\\]/;
 const PLACEHOLDER_ARG = "${CLAUDE_PLUGIN_ROOT}/start.mjs";
 
 /**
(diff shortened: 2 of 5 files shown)


mksglu:
Thanks for the fix @sebastianbreguel — it correctly addresses #711's root cause.

Reviewed under our context-mode-ops mandate (Git Archaeology + 3-OS check). The change is sound for Linux/macOS; we want to flag a Windows regression risk before merging.

**Git archaeology**
- `622b82d` (#378) introduced `normalizeHooksOnStartup` to repair Windows hooks.json placeholders.
- `13d1342` (#528) wired it into `upgrade()` with an explicit two-goal comment block: (1) self-heal poisoned tmpdir paths, (2) **close the #414 gap so the first hook fires correctly between /ctx-upgrade and the next MCP boot**.

This PR removes goal (2) — `cli.ts upgrade()` no longer normalizes hooks.json. The broadened regex in `heal-installed-plugins.mjs` runs on MCP boot via `start.mjs` Layer 5, but Claude Code fires SessionStart/PreToolUse hooks **independently of MCP boot**. On Windows + Git Bash, the unresolved \`\${CLAUDE_PLUGIN_ROOT}\` placeholder will MODULE_NOT_FOUND for the first hook fire after /ctx-upgrade but before MCP server restart — exactly the gap #378/#528 closed.

CI is green but only exercises the heal script itself, not the cross-process first-hook-fire scenario.

**Path forward — pick one**
1. Keep upgrade-time normalize but narrow to hooks.json only with the broadened regex from this PR (most defensive).
2. Add a postinstall step that resolves placeholders eagerly on Windows.
3. Add a regression test driving the cross-process sequence: /ctx-upgrade → spawn a hook child → assert it can resolve start.mjs without MCP boot.

Once the Windows path is covered, this is mergeable.

Related: PR #728 is the complementary runtime/start.mjs forward-heal for the same family — both should land together.

mksglu:
## Followup commit: `5bbda87` — restore Windows first-hook-fire heal via narrow normalize

Hey @sebastianbreguel — beautiful diagnosis on #711, the asymmetric-drift insight + the broadened heal regex are exactly right. Pushed one followup to your branch (`maintainerCanModify=true`) to close one remaining gap before we merge:

### Why the followup

PR removes `normalizeHooksOnStartup` from `src/cli.ts upgrade()` entirely, which correctly fixes #711 but re-opens the Windows + Git Bash first-hook-fire window that `13d1342` (#528) explicitly closed. From the #528 commit body:

> Closes #414 for fresh installs — the first hook fire after upgrade now works without waiting for MCP boot.

Concretely: Claude Code fires `SessionStart` / `PreToolUse` hooks **before** the MCP server boots. On Windows + Git Bash, the unresolved `${CLAUDE_PLUGIN_ROOT}` placeholder in `hooks/hooks.json` yields `MODULE_NOT_FOUND` for the first hook fire after every `/ctx-upgrade`, until `start.mjs` eventually boots and rewrites it.

### What changed in the followup

The two files are not symmetric:

| File | Carried forward to new version cache dir by Claude Code? | Read by pre-MCP child process? |
|---|---|---|
| `.claude-plugin/plugin.json` | yes (auto-update) — baking absolute paths → #711 | no |
| `hooks/hooks.json` | no (per-version, regenerated each upgrade) | **yes** — first hook fire |

So plugin.json must NOT be baked during upgrade (your fix), but hooks.json **must** be normalized so the first-fire works on Windows.

Split the function in `hooks/normalize-hooks.mjs`:

- `normalizeHooksJsonOnly({pluginRoot, nodePath, platform})` — touches `hooks/hooks.json` only. Called from `src/cli.ts upgrade()`. Closes #414 without re-introducing #711.
- `normalizeHooksOnStartup` — unchanged contract for `start.mjs` and `scripts/postinstall.mjs`. Internally now delegates the hooks.json branch to the narrow helper so there's one implementation.

The #604 ratchet detection in `needsHookNormalization` still auto-heals any forward-carried `hooks.json` on the next boot, so even if Claude Code carries the per-version hooks.json forward (it doesn't, but belt-and-braces), the next MCP boot re-points it.

### Tests added

`tests/hooks/post-upgrade-first-hook-fire.test.ts` — 10 new tests including:

- `normalizeHooksJsonOnly` rewrites hooks.json on Windows / Linux, no-op on macOS
- **`normalizeHooksJsonOnly` leaves plugin.json UNTOUCHED on Windows** — the #711 invariant, asserted directly
- `normalizeHooksJsonOnly` is idempotent
- `normalizeHooksOnStartup` still rewrites BOTH files (boot-time contract regression guard)
- **Cross-process first-hook-fire**: a freshly normalized `hooks.json` yields an absolute parseable command that a child Node process can resolve without `MODULE_NOT_FOUND`, mirroring Claude Code's pre-MCP hook dispatch

### Verification

- All 60 normalize + heal tests pass (incl. 10 new + 50 existing)
- All 25 upgrade-assertion tests pass (`upgrade-mcp-json-assertion`, `upgrade-plugin-json-assertion`, `upgrade-verifies-binding`, `postinstall-heal`)
- `npx tsc --noEmit` clean
- Bundles untouched (CI rebuilds on main push per project convention)

### Archaeology trail preserved

- `622b82d` (#378) — origin of `normalizeHooksOnStartup`
- `13d1342` (#528) — wired into upgrade with the two explicit goals (a) self-heal legacy poisoned configs, (b) close #414 gap. Goal (a) is now handled by your broadened heal regex; goal (b) is now handled by the narrow `normalizeHooksJsonOnly` call.
- `4362fa5` (#531) — `healMcpJsonArgs` companion (untouched)
- `edb4b06` (this PR) — fixes #711
- `5bbda87` (followup) — restores #414 invariant without re-opening #711

### Next step

Take a look, and if it reads right, request a re-review and we're good to merge. Happy to iterate if you prefer a different split (e.g. inlining the narrow check into upgrade() vs the exported helper).

PR #716, opened 2026-05-26
security: containment, info-disclosure, and test-isolation hardening

kerneltoast:
## Summary

Deep security audit across the in-scope codebase. Nine fixes (plus one test-only slice widening) covering exploitable supply-chain / local-trust-boundary paths, an info-disclosure DoS surface, and destructive-test hygiene.

### Containment guards (the big block of work)

- `bf4c948` -- `/ctx-upgrade` swap loop in both `cli.ts` and `server.ts` (the inline-fallback variant) iterated `pkg.files[]` from the freshly cloned upstream `package.json` and handed each entry to `rmSync` + `cpSync` with no `startsWith(pluginRoot)` containment. A compromised upstream tag shipping `files: ["../../.ssh/authorized_keys"]` would rm and overwrite arbitrary files under the user's UID. Mirrors the lexical guard pattern that `hooks/heal-partial-install.mjs` already uses for its own `files[]` expansion.

- `e7d5599` -- `cli.ts` reads `~/.claude/plugins/installed_plugins.json` in two hot spots (`upgrade()` skills cpSync and `statuslineForward()` dynamic-import). The same field already had a `startsWith(cacheRoot + sep)` guard in `server.ts:790` (`healCacheMidSession`); this just propagates it to the more sensitive call sites. statusline re-fires several times per second, so a malicious `installPath` planted in the registry by a co-resident plugin or postinstall would otherwise hand the attacker durable RCE under the user account.

- `dff7bac` -- closing adversarial sweep caught that the lexical guards above (and the swap-loop guard in `bf4c948`) are bypassable via a planted symlink AT `cacheRoot/<owner>/<plugin>/<version>` because `path.resolve` is purely lexical and doesn't dereference symlinks. Adds a `realpathSync` re-check after the lexical gate for the installPath sites and an `lstatSync`-backed `cpSync({filter})` for the swap loop so neither a symlink anchor at the registry path nor a committed symlink inside the cloned source tree can sneak past.

- `b01278d` -- `ctx_insight` accepted `sessionDir` / `contentDir` MCP overrides that flowed verbatim into `INSIGHT_SESSION_DIR` / `INSIGHT_CONTENT_DIR` on the spawned dashboard subprocess. Without containment, a prompt-injected MCP caller passing `sessionDir="~/.ssh"` turns the dashboard into a confused-deputy enumerator over the user's filesystem (and, for hex-named SQLite files, can delete rows via `/api/content`). Gates explicit overrides to the adapter config root (`dirname(dirname(getSessionDir()))`).

- `760d476` -- `healClaudeJsonMcpArgs` rewrote `mcpServers.args[]` in `~/.claude.json` by slicing the arg, then `resolve(newPluginRoot, suffix)`. The slice is a literal substring; `..` characters embedded in the original arg survive into `suffix`, and `resolve` normalizes them when it builds `newArg`. A crafted `~/.claude.json` could trick the heal into rewriting MCP spawn args at attacker-chosen paths. Adds a post-resolve `startsWith` guard.

- `25606f2` -- `ctx_index`'s directory-dispatch branch used `statSync` rather than `lstatSync`, so a path like `/tmp/link` with `link -> /etc` was reported as a directory and walked into `/etc`. The deny-glob check at the head of the handler ran against `/tmp/link`, not the symlink target, so users whose deny globs included `/etc` but not `/tmp` would still see `/etc` content land in the FTS5 store. Switches to `lstatSync` + `realpathSync` + re-applied deny check.

### Info-disclosure / DoS

- `3568d9a` -- bundles two `src/server.ts` fixes. `ctx_fetch_and_index` had no size cap on either the subprocess fetch or the parent's `readFileSync(outputPath)`, so a malicious URL returning multi-GB content could OOM the long-running MCP server. Adds a 50 MB cap on both ends. Same commit sanitizes `CLAUDE_SESSION_ID` before splicing it into a `path.join`-based stats filename, so a host-controlled `CLAUDE_SESSION_ID="../../evil"` can no longer write outside `statsDir`.

### Destructive-test hygiene (matches PR #702 pattern)

- `c8a5728` -- `tests/lifecycle.test.ts` wrote `_lifecycle_test_*.ts` to `process.cwd()` (the project root, not gitignored) and `tests/adapters/cursor.test.ts` mkdir'd, wrote, and `rmSync(force:true)`'d `.cursor/mcp.json` against `process.cwd()`, silently clobbering the contributor's real `.cursor/mcp.json` on `npm test`. Both blocks now sandbox the writes inside per-test `mkdtempSync` directories (and chdir for the cursor case so the adapter still finds the test fixture).

### Test-only follow-up

- `123b21b` -- three pre-existing tests in `tests/core/cli.test.ts` sliced `cli.ts` source by fixed-width windows that the new containment comments + checks pushed assertions out of. Widened the slices to end-of-file; the textual landmarks (`sweepStaleMcpJson`, `process.exit(0)`) still anchor the regex assertions, so the wider slices don't weaken anything.

### Out-of-scope (deferred to follow-up if you want them)

- `F02` SSRF defaults: working as designed (loopback / RFC1918 reachable unless `CTX_FETCH_STRICT=1`); could be surfaced in `ctx_doctor`.
- `F08`: `...process.env` spread into the npm install subprocess on `/ctx-upgrade` still propagates `CLAUDE_*` env into transitive postinstall scripts.
- `F27`: ~20 `writeFileSync` call sites on critical config files don't use `tmp + rename`. Reliability concern, not exploitable.
- `F28`: same-uid `chmod` TOCTOU in `hooks/cache-heal-utils.mjs`.
- `CONTEXT_MODE_SECURITY_BUNDLE_PATH` env-controlled dynamic import in `hooks/core/routing.mjs:364`. Documented as a test seam; env is parent-process-controlled (same trust as the user). Fixing strictly would break legitimate test fixtures in tmpdir.

## Test plan

- [ ] `node node_modules/.bin/tsc --noEmit` is clean.
- [ ] `vitest run tests/core/cli.test.ts` -- 173 / 173 pass.
- [ ] `vitest run tests/util/heal-installed-plugins.test.ts` -- 33 / 33 pass.
- [ ] `vitest run tests/adapters/cursor.test.ts` -- 39 / 39 pass.
- [ ] `vitest run tests/core/server.test.ts` -- 463 / 463 pass when run unsandboxed (the in-sandbox EROFS failures are pre-existing in any executor / turndown / projectRoot test that needs `mkdtempSync("/tmp/.ctx-mode-")`).
- [ ] Manual /ctx-upgrade smoke on a fresh install (legitimate path) -- the cosmetic regression for users whose `~/.claude` itself is a symlink is documented in the `dff7bac` commit message and is not a security regression.

Co-authored-by: Claude Opus 4.7 (1M context) <noreply@anthropic.com>

### DIFF
--- a/scripts/heal-installed-plugins.mjs
+++ b/scripts/heal-installed-plugins.mjs
@@ -430,6 +430,18 @@ export function healClaudeJsonMcpArgs({ dotClaudeJsonPath, pluginCacheParent, ne
   }
 
   const cacheParentFwd = pluginCacheParent.replace(/\\/g, "/");
+  // Post-resolve containment on newArg. ~/.claude.json is locally user-
+  // writable (same trust boundary as installed_plugins.json), and the
+  // `suffix` slice is derived from arg strings inside the existing config.
+  // A crafted arg like
+  //   .../cache/<owner>/<plugin>/1.0.0/../../../evil/start.mjs
+  // slices to suffix="../../../evil/start.mjs", and resolve(newPluginRoot,
+  // suffix) normalizes to an attacker-chosen .mjs path outside the plugin
+  // cache. Writing that path back into ~/.claude.json mutates the mcpServers
+  // args so the next MCP boot spawns from the attacker path. Reject any
+  // suffix that escapes newPluginRoot.
+  const newPluginRootResolved = resolve(newPluginRoot);
+  const newPluginRootWithSep = newPluginRootResolved + sep;
 
   let mutated = false;
   for (const srv of Object.values(servers)) {
@@ -444,6 +456,12 @@ export function healClaudeJsonMcpArgs({ dotClaudeJsonPath, pluginCacheParent, ne
       if (slashIdx < 0) continue;
       const suffix = rel.slice(slashIdx + 1);
       const newArg = resolve(newPluginRoot, suffix);
+      if (
+        newArg !== newPluginRootResolved &&
+        !(newArg + sep).startsWith(newPluginRootWithSep)
+      ) {
+        continue;
+      }
       if (newArg !== arg) {
         srv.args[i] = newArg;
         mutated = true;
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -17,7 +17,7 @@
 import * as p from "@clack/prompts";
 import color from "picocolors";
 import { execFileSync, execSync, execFile as nodeExecFile, type ExecSyncOptions } from "node:child_process";
-import { readFileSync, writeFileSync, cpSync, accessSync, existsSync, readdirSync, rmSync, closeSync, openSync, chmodSync, mkdirSync, constants } from "node:fs";
+import { readFileSync, writeFileSync, cpSync, accessSync, existsSync, readdirSync, rmSync, closeSync, openSync, chmodSync, mkdirSync, lstatSync, realpathSync, constants } from "node:fs";
 import { request as httpsRequest } from "node:https";
 import { resolve, dirname, join, sep } from "node:path";
 import { tmpdir, devNull, homedir } from "node:os";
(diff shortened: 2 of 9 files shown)


NgoQuocViet2001:
@kerneltoast I checked this on Windows locally and found a few test-only issues, not a core security-code regression.

Findings:
- `tests/core/cli.test.ts` and `tests/core/server.test.ts` create directory symlinks with `symlinkSync(..., "dir")`. On Windows without symlink privilege / Developer Mode this fails with `EPERM`. Using `"junction"` on Windows keeps `lstatSync(...).isSymbolicLink() === true` and makes the tests pass.
- `tests/core/server.test.ts` still expects the old inline fallback snippet `cpSync(from,to,{recursive:true,force:true})`; this PR now correctly emits `filter:noSymlink`, so the assertion needs updating.
- `tests/lifecycle.test.ts` shells out to `which tsx`, which fails in my Windows local run. Resolving `node_modules/.bin/tsx.cmd` first, then falling back to `where/which`, fixes it.

Validated locally on Windows:
- `node node_modules/vitest/vitest.mjs run tests/core/cli.test.ts`
- `node node_modules/vitest/vitest.mjs run tests/core/server.test.ts -t "inline fallback copies package files to plugin root|lstatSync \\+ realpath identifies a symlinked directory"`
- `node node_modules/vitest/vitest.mjs run tests/lifecycle.test.ts`
- `node node_modules/vitest/vitest.mjs run tests/util/heal-installed-plugins.test.ts`
- `node node_modules/vitest/vitest.mjs run tests/adapters/cursor.test.ts`
- `git diff --check`

I have the tiny patch locally if useful.

kerneltoast:
> @kerneltoast I checked this on Windows locally and found a few test-only issues, not a core security-code regression.

@NgoQuocViet2001 Added it as https://github.com/mksglu/context-mode/pull/716/changes/f106f56fc355758fcdb20ed5d9a6d0acdf764d42 with you as co-author, thanks!

Issue #639, opened 2026-05-19
test(B12): add live integration test for stdin EOF propagation in Bun re-exec path


sborisov88:
## Context

Commit `4ee5168` (`fix(start): propagate parent stdin EOF to Bun re-exec'd child`) fixed a hang in `start.mjs` where the `process.stdin.on("end", ...)` handler was a literal no-op. Claude Code closing its end of the pipe caused the parent to see EOF, ignore it, and keep `child.stdin` open forever — the MCP server then blocked on `process.stdin` and the whole stack hung.

The fix was covered by a **static regression test** (`tests/util/start-mjs-stdin-eof.test.ts`) that pins three textual properties of the launcher:
- `end` handler calls `child.stdin.end()`
- `end` handler is not the literal no-op
- `end` handler honours the destroyed-stream guard

## What's missing

A **live integration test** that:
1. Spawns `start.mjs` under Linux + Bun (the path that triggers the re-exec)
2. Writes a few bytes to stdin
3. Closes stdin (EOF)
4. Asserts the MCP child exits within a reasonable timeout (e.g. 5 s)

The static guard is sufficient to prevent *regression* (nobody can remove the propagation call without breaking the test), but it cannot catch *incorrect* propagation (e.g. the call exists but targets the wrong stream, or is wrapped in a condition that never fires).

## Environment requirements

- Linux runner with `bun` installed and on `PATH`
- The test needs to be either skipped on macOS/Windows or run only in CI on a Linux matrix leg

## Labels

`test`, `good first issue`

Issue #242, opened 2026-04-08
fix(hooks): AfterTool hang on large output or stdin stall in Gemini CLI


tndhuy:
### Description
When using `context-mode` with Gemini CLI (Node v24+), the `AfterTool` hook frequently fails with a `Hook(s) failed for event AfterTool` warning. This happens because the hook tries to read from `stdin` without a timeout, and processes potentially large tool responses (multi-MB) within the strict <20ms budget of the CLI.

### Symptoms
- Constant `AfterTool failed` warnings in Gemini CLI UI.
- Hook processes hanging or being killed by the system.
- No clear error in `aftertool-debug.log` because the process is terminated mid-execution.

### Root Cause
1. **Stdin Hang**: `readStdin()` lacks a timeout guard.
2. **Performance Budget**: Extracting events from very large tool outputs (e.g., massive `run_shell_command` output) exceeds the time limit.

### Proposed Solution (Applied locally)
I have applied a patch to `hooks/gemini-cli/aftertool.mjs` with:
- A 10s `setTimeout` guard on `stdin` to prevent permanent hangs.
- A size check: if `tool_response` > 1MB, skip event extraction to preserve CLI responsiveness.

```javascript
// Stdin timeout guard
const stdinTimeout = setTimeout(() => process.exit(0), 10000);

// ... readStdin ...
clearTimeout(stdinTimeout);

// Response size guard
const responseSize = typeof input.tool_response === 'string' ? input.tool_response.length : 0;
if (responseSize > 1024 * 1024) {
  process.exit(0);
}
```

### Environment
- OS: Darwin (macOS)
- Node: v24.4.1
- Gemini CLI: Latest

murataslan1:
Validated locally on macOS + Node v24.4.1 + Gemini CLI v0.37.0.

Both guards (stdin timeout + 1MB response size check) work as expected:
- Stdin stall: process exits cleanly after 10s instead of hanging
- 2MB response: skipped in <10ms, logged as SKIP
- Normal responses: still processed correctly
- Live Gemini CLI session: no `AfterTool failed` warnings

Full test details and patch: https://gist.github.com/murataslan1/3ca3dc332a10315ddd6577d16fa2cb3e

---

[...]

PR #139, opened 2026-03-17
fix: strip UTF-8 BOM from stdin before JSON.parse in hooks

huuthanh21:
## What

Strip leading UTF-8 BOM (`U+FEFF`) from stdin in both `readStdin()` implementations before returning the string to callers.

## Why

On Windows, Cursor pipes hook stdin with a leading UTF-8 BOM. `JSON.parse()` rejects BOM at position 0, crashing every hook (`preToolUse`, `postToolUse`, `sessionStart`) with:

```
SyntaxError: Unexpected token '', "{"convers"... is not valid JSON
```

This makes context-mode completely unusable on Cursor for Windows users.

## How

Added `.replace(/^\uFEFF/, "")` to the resolve callback of both `readStdin()` functions. The `^` anchor ensures only a leading BOM is stripped — mid-string `\uFEFF` inside JSON string values is never touched. On platforms that don't send a BOM, the regex is a no-op.

Changed files:
- `hooks/core/stdin.mjs` — used by Claude Code, VS Code Copilot, Kiro, Gemini CLI pretooluse hooks
- `hooks/session-helpers.mjs` — used by Cursor, session, posttooluse, precompact hooks

## Affected platforms

- [x] Claude Code
- [x] Gemini CLI
- [x] VS Code Copilot
- [x] OpenCode
- [x] Codex CLI
- [x] All platforms / core MCP server

Both `readStdin()` implementations are patched. The fix is defensive — it's a no-op on platforms that don't send a BOM, but protects all platforms if they ever do.

### RED (failing test)

Tests written first, run without the fix applied:

```
× tests/hooks/cursor-hooks.test.ts > Cursor hooks > UTF-8 BOM handling > pretooluse.mjs parses BOM-prefixed stdin without error 83ms
✓ tests/hooks/cursor-hooks.test.ts > Cursor hooks > UTF-8 BOM handling > posttooluse.mjs parses BOM-prefixed stdin without error 77ms
✓ tests/hooks/cursor-hooks.test.ts > Cursor hooks > UTF-8 BOM handling > sessionstart.mjs parses BOM-prefixed stdin without error 90ms
× tests/hooks/integration.test.ts > UTF-8 BOM handling (core/stdin.mjs path) > pretooluse.mjs parses BOM-prefixed stdin without error 76ms
× tests/hooks/integration.test.ts > UTF-8 BOM handling (core/stdin.mjs path) > pretooluse.mjs handles BOM-prefixed Bash input correctly 77ms
```

[...]

PR #719, opened 2026-05-26
Prevent hook stdin idle hangs

looooown2006:
﻿## Summary

- add an idle timeout to hook stdin readers so hook subprocesses do not wait forever when a client keeps stdin open
- clean up stdin listeners and pause/destroy stdin once the hook input is considered complete
- add regression coverage for both shared hook stdin readers with an intentionally open stdin pipe

## Why

On Windows Codex Desktop, hook commands like `context-mode hook codex pretooluse` can remain alive when the parent process does not promptly close the hook stdin pipe. Those orphaned hook subprocesses accumulate across hook events and increase app-server/process-management pressure. Hook stdin should be a short-lived JSON payload; after the pipe is idle, the hook can safely continue with the bytes received so far.

The timeout defaults to 1500ms and can be tuned via `CONTEXT_MODE_HOOK_STDIN_IDLE_MS`.

## Verification

- `node --check hooks/session-helpers.mjs`
- `node --check hooks/core/stdin.mjs`
- `npx vitest run tests/hooks/stdin-idle-timeout.test.ts`
- `npm run typecheck`

Additional note: `npx vitest run tests/session-hooks-smoke.test.ts tests/hooks/cursor-hooks.test.ts tests/hooks/gemini-hooks.test.ts` passed the cursor/gemini suites, but `session-hooks-smoke.test.ts` hit a Windows symlink permission error (`EPERM: operation not permitted, symlink ... node_modules ...`) in the test harness.


mksglu:
Thanks for tackling this @looooown2006 — the spawn-based test is exactly the right shape for a stdin-lifecycle bug.

Reviewed under our context-mode-ops mandate. Two blockers + cleanup before merge.

**Blockers**

1. **Silent truncation on partial-buffer timeout.** `hooks/core/stdin.mjs:5-8` documents the contract as "Read all of stdin as a string". The idle timer resolves with whatever `data` happens to be buffered when the timer fires. For a hook receiving a large `tool_response` over a slow pipe (the #242 scenario), this returns a truncated payload that `JSON.parse` either rejects (loud) or — worse — parses as a valid prefix and routes on stale input (silent corruption).
   - **Fix**: when the timer fires with `data.length > 0`, **reject** with \`stdin idle for \${idleMs}ms with \${data.length} bytes buffered\`. Resolve-with-empty only when `data.length === 0` (the true zero-EOF case from #242/#639). Preserves the all-or-error contract.

2. **Docstring drift.** Both `hooks/core/stdin.mjs:1-8` and `hooks/session-helpers.mjs:213-215` still say "Read all of stdin as a string". With this PR, that's no longer true. Update both docstrings to describe the idle-timeout semantics and link the `CONTEXT_MODE_HOOK_STDIN_IDLE_MS` env var.

**Cleanup**
- Drop the `stats.json` hunk — CI bot owns that file; including it will conflict on every next push.
- Branch name targets codex, but `refs/platforms/codex/codex-rs/hooks/src/engine/command_runner.rs:36-69` shows codex DOES send EOF cleanly (`stdin` is dropped after `write_all`, closing the pipe). The real hangs are #242 (Gemini CLI AfterTool >1MB) and #639 (Bun re-exec EOF propagation). Consider re-framing the body so future archaeology lands on the right target.

Once those land, ready to merge.

mksglu:
Hi @looooown2006 — thanks for the patch. I pushed one follow-up commit to your branch (you have `maintainerCanModify=true` so it lands on your PR directly):

**Commit:** `fix(hooks): reject readStdin on partial-buffer idle to avoid silent JSON corruption` (2b9e69f)

### What changed and why

The original idle path called `resolve(data)` whenever the timer fired, regardless of how much had already buffered. That branch is fine when nothing arrived (the case this PR targets, plus #639), but it silently truncates streamed payloads — the host then sees a misleading `JSON.parse` `SyntaxError` on garbage instead of the real cause (stall). On the real-world offenders (#242 Gemini AfterTool >1MB `tool_response`, #639 Bun re-exec EOF), silent truncation is the worst-case failure mode.

Split the idle branch in both readStdin implementations (`hooks/core/stdin.mjs` + `hooks/session-helpers.mjs`):

- `data.length === 0` → `resolve("")` (preserves the zero-EOF behaviour your PR fixed)
- `data.length > 0`  → `reject(new Error("stdin idle for ${idleMs}ms with ${data.length} bytes buffered"))`

EOF (`onEnd`) still calls `resolveBuffer()` so the well-behaved fast path is unchanged, and the BOM strip from #139 (Cursor on Windows — `a6a9e7e`) still runs on every resolve.

### Branch name vs root cause

Branch is named `fix-codex-hook-stdin-idle-timeout` but codex-rs (`hooks/src/engine/command_runner.rs`) closes its end of the pipe cleanly, so codex itself does not stall. The real offenders are **#242** (Gemini AfterTool large `tool_response`) and **#639** (Bun re-exec EOF propagation). Cited in the commit body since branch names are immutable.

### Tests added (`tests/hooks/stdin-idle-timeout.test.ts`)

Two new RED tests prove the partial-buffer reject contract on both implementations:

- `session helper readStdin rejects when partial data arrives and pipe stalls` (lines 92-100)
- `core readStdin rejects when partial data arrives and pipe stalls` (lines 102-110)

Asserts: exit code 1, stderr matches `/stdin idle/` and includes the buffered byte count. The original two zero-EOF tests still pass unchanged.

### Validation

- `npx vitest run tests/hooks/` — 442 passed, 1 skipped, 0 failed
- `npx tsc --noEmit` — clean

### Cleanup notes

- The `stats.json` hunk in the PR diff is a base-divergence artifact (the bot updated it on `next` after you branched). Nothing to do on your side — the merge into `next` will resolve it automatically.
- Docstrings on both `readStdin` impls now describe the idle-timeout semantics and reference the `CONTEXT_MODE_HOOK_STDIN_IDLE_MS` env var.

### Next step

If the new commit looks good, no action needed — Mert can take it from here for review and merge. If you want to amend or override, push to the same branch and my commit is just a follow-up that can be reverted.

PR #721, opened 2026-05-26
feat(cli): add indexing and search commands

dennesneumann:
## Summary

Adds terminal and slash-style entry points for pre-warming and querying context-mode's persistent FTS5 knowledge base.

- add `context-mode index <path>` for indexing local files/directories into the per-project content DB
- add `context-mode search <query...>` for querying indexed content from the terminal
- add user-invocable `/context-mode:ctx-index` and `/context-mode:ctx-search` skills
- document the new slash, chat utility, and terminal commands
- add focused CLI surface regression tests

## Why

`ctx_index(path)` already supports indexing file and directory content through MCP, but there was no deterministic terminal entry point for pre-filling the SQLite/FTS5 store before or outside an agent session. This makes the workflow discoverable and automatable while keeping raw project bytes out of chat context.

## Validation

- `pnpm exec tsc --noEmit --pretty false`
- `pnpm exec vitest run tests/core/cli.test.ts --reporter=dot`
- smoke test: `context-mode index README.md --project . --source smoke:readme`, then `context-mode search "context savings" --project . --source smoke:readme --limit 1`
- `git diff --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,11 +89,13 @@ All checks should show `[x]`. The doctor validates runtimes, hooks, FTS5, and pl
 |---|---|
 | `/context-mode:ctx-stats` | Context savings — per-tool breakdown, tokens consumed, savings ratio. |
 | `/context-mode:ctx-doctor` | Diagnostics — runtimes, hooks, FTS5, plugin registration, versions. |
+| `/context-mode:ctx-index` | Index a local file or directory into the persistent FTS5 knowledge base. |
+| `/context-mode:ctx-search` | Search previously indexed content. |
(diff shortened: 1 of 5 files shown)


mksglu:
Thanks @dennesneumann. Conflicting on `next` after #716 modified `src/cli.ts`. Please rebase — the `index` and `search` subcommands shouldn't overlap textually with #716's swap-loop hardening, but the file header / command-registration block likely needs a clean replay.

Optional follow-up nit: pass `attribution: { tool: "cli-index" }` to `store.index()` so Insight analytics can distinguish CLI-pre-warmed chunks from agent-indexed ones. Not blocking.

Verdict was MERGE_WITH_NITS — re-ping after rebase.

PR #725, opened 2026-05-26
fix(routing): bootstrap deferred ctx_* tools in subagent prompts (#724)

### LINKED ISSUE #724: Agent PreToolUse routing stalls read-only subagents: injected block omits the ToolSearch bootstrap for deferred ctx_* tools
# PreToolUse `Agent` routing stalls read-only subagents (Explore) for many minutes

**Repo:** `mksglu/context-mode`
**Version observed:** 1.0.103 (behavior unchanged through 1.0.151)
**Harness:** Claude Code (plugin install via marketplace)

## Summary

The `Agent` PreToolUse hook rewrites every subagent spawn to append the
`<context_window_protection>` routing block, instructing the subagent that it
**MUST** use the `ctx_*` MCP tools and **must not** use Bash/Read/Grep for
analysis. For **read-only subagents (Claude Code's `Explore` type)** this is a
poor fit: the agent's entire purpose is Grep/Read, and in practice the injected
constraint makes it crawl — observed stalls of ~8 minutes with near-zero
transcript output, and one Explore agent that did not produce a result for
~63 minutes (it eventually completed). The same task done directly in the main
thread (ripgrep + Read) finished in seconds.

## Where

- `hooks/hooks.json` — PreToolUse includes `{ "matcher": "Agent" }` → `pretooluse.mjs`.
- `hooks/core/routing.mjs` (~L817–833 in 1.0.151; ~L322–337 in 1.0.103):
  ```js
  // ─── Agent: inject context-mode routing into subagent prompts ───
  if (canonical === "Agent") {
    const subagentType = toolInput.subagent_type ?? "";
    const fieldName = ["prompt","request","objective","question","query","task"].find(f => f in toolInput) ?? "prompt";
    const subagentBlock = createRoutingBlock(t, { includeCommands: false });
    const updatedInput =
      subagentType === "Bash"
        ? { ...toolInput, [fieldName]: prompt + subagentBlock, subagent_type: "general-purpose" }
        : { ...toolInput, [fieldName]: prompt + subagentBlock };
    return { action: "modify", updatedInput };
  }
  ```
- `hooks/routing-block.mjs` — the injected block: *"MUST use context-mode MCP
  tools"*, *"Primary research tool: `ctx_batch_execute`"*, *"NO Read for
  analysis"*.
- `hooks/core/tool-naming.mjs:18-19` — for `claude-code` the block names the
  fully-qualified `mcp__plugin_context-mode_context-mode__ctx_*` tools (so this
  is **not** a tool-name-resolution bug).

## Why it stalls (analysis — confirmed by probe below)

Tool-name resolution is correct, so the problem is the *reachability* of the
redirected path inside a subagent. The **confirmed dominant cause** is:

- **Deferred tools + no bootstrap instruction.** In this harness the `ctx_*`
  tools are surfaced as *deferred* — their schemas must be loaded via
  `ToolSearch` before they can be called. The injected block tells the subagent
  it **MUST** use `ctx_batch_execute` and **must not** use Bash/Read, but never
  instructs it to `ToolSearch` the `ctx_*` schemas first. A capable agent that
  reasons about deferred tools (e.g. `general-purpose`) recovers on its own; a
  narrow, fan-out search agent (`Explore`) given this contradiction tends to try
  the un-loaded tool, fail, and churn — producing near-zero output for minutes.

The two other hypotheses I initially considered were **ruled out** by the probe:
MCP cold-start (`ensure-deps`/index build) and shared-SQLite/FTS5 lock contention
were **not** observed — the first `ctx_batch_execute` call returned in ~1–2s with
no install/indexing delay.

## Empirical probe (this environment)

A `general-purpose` subagent was asked to load and call `ctx_batch_execute`.
The injected routing block was applied to it as expected. Findings:

| Question | Result |
| --- | --- |
| `ctx_*` tools callable without `ToolSearch`? | **No** — deferred; schemas not loaded |
| `ToolSearch` present in the subagent's toolset? | **Yes** |
| After `ToolSearch` loaded the schema, did `ctx_batch_execute` work? | **Yes** — returned `ctx-probe-ok`, indexed 1 section |
| Time to first `ctx_` result | **~1–2s** — no cold-start / install / index delay |
| Was plain `echo` via Bash redirected/denied? | **No** — allowed directly |

**Conclusion:** the `ctx_*` path works fine and fast from a subagent *once the
schema is loaded via `ToolSearch`*. The stall is caused by the injected block
directing subagents to "MUST use `ctx_batch_execute`" without telling them to
`ToolSearch` it first — fatal for narrow agents that won't infer that step. (The
probe also surfaced a version skew: the running MCP server reported `v1.0.122`
while the plugin hooks dir reported `1.0.103`.)

## Suggested fixes (in priority order)

1. **Make the block self-bootstrapping (primary fix — directly addresses the
   confirmed cause).** Prepend an explicit first step to the injected subagent
   block: *"The `ctx_*` tools may be deferred — call `ToolSearch` with
   `select:mcp__plugin_context-mode_context-mode__ctx_batch_execute,…` to load
   their schemas before using them."* The probe shows this is all the difference
   between a 19s success and a multi-minute stall.
2. **Skip injection for read-only/search subagent types** (`Explore`), which are
   designed around Grep/Read; forcing them onto MCP tools is a poor fit even when
   it works.
3. **Detect tool availability** before forbidding Bash/Read: only inject the
   "MUST use MCP tools / NO Bash/Read" constraint when the subagent demonstrably
   has the `ctx_*` tools callable; otherwise fall back to a softer hint.

## Impact

Anyone using context-mode + Claude Code subagents (especially the built-in
`Explore` agent) for parallel investigation hits multi-minute stalls. The
workaround is to avoid subagents and search in the main thread, which defeats the
point of parallel research agents.

beastawakens:
Fixes #724.

## Problem

The `Agent` PreToolUse hook appends the routing block to every subagent prompt,
instructing it to use the `ctx_*` MCP tools and **not** Bash/Read. But in Claude
Code those `ctx_*` tools are **deferred** — their schemas must be loaded via
`ToolSearch` before they can be called. The injected block never tells the
subagent to do that, so a narrow subagent (e.g. the built-in `Explore` type)
is told to use a tool it cannot yet invoke and stalls — observed multi-minute
hangs with near-zero output (one ran ~63 min).

A controlled probe confirmed the mechanism: a `general-purpose` subagent (which
carries `ToolSearch`) loaded the schema and ran `ctx_batch_execute` in ~1–2s with
no cold-start; the only thing separating success from a stall was the missing
`ToolSearch` bootstrap step. (Cold-start and SQLite/FTS5 lock contention were
ruled out.)

## Fix

Make the subagent routing block self-bootstrapping. `createRoutingBlock` gains an
opt-in `toolSearchBootstrap` option (default `false`, so the main-session block
and all other callers are unchanged). When enabled, it prepends a
`<deferred_tool_bootstrap>` section telling the subagent to `ToolSearch` the
`ctx_*` schemas before first use and not to fall back to Bash/Read just because a
first call failed as not-found.

The `Agent` branch in `routing.mjs` enables it **only for claude-code** (the
default when `platform` is unset); other platforms don't defer tools, so they're
untouched.

## Changes

- `hooks/routing-block.mjs` — add `toolSearchBootstrap` option + the
  `<deferred_tool_bootstrap>` section (uses the platform tool-namer for correct
  names).
- `hooks/core/routing.mjs` — `Agent` branch passes `toolSearchBootstrap` gated on
  claude-code.
- `tests/core/routing.test.ts` — update the shared `SUBAGENT_BLOCK` constant and
  add two cases (bootstrap present for claude-code, absent for `codex`).

## Testing

- `npx vitest run tests/core/routing.test.ts` → 29/29 pass.
- Default-off option; `ROUTING_BLOCK` (main session) and non-claude-code platforms
  are byte-for-byte unchanged.

## Alternative / follow-up

A complementary option (not done here, to keep the change minimal) is to skip the
injection entirely for read-only subagent types like `Explore`, since they're
designed around Grep/Read. Happy to add that if you'd prefer it.

### DIFF
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -822,7 +822,15 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
     const fieldName = ["prompt", "request", "objective", "question", "query", "task"].find(f => f in toolInput) ?? "prompt";
     const prompt = toolInput[fieldName] ?? "";
 
-    const subagentBlock = createRoutingBlock(t, { includeCommands: false });
+    // Claude Code surfaces ctx_* as DEFERRED tools (schemas loaded via ToolSearch).
+    // Without a bootstrap step the subagent is told to use ctx_* tools it cannot yet
+    // invoke and stalls (see #724). Prepend the ToolSearch bootstrap for claude-code
+    // (the default when platform is unset). Other platforms don't defer, so skip it.
+    const isClaudeCode = !platform || platform === "claude-code";
+    const subagentBlock = createRoutingBlock(t, {
+      includeCommands: false,
+      toolSearchBootstrap: isClaudeCode,
+    });
 
     const updatedInput =
       subagentType === "Bash"
--- a/hooks/routing-block.mjs
+++ b/hooks/routing-block.mjs
@@ -14,13 +14,19 @@ import { createToolNamer } from "./core/tool-naming.mjs";
 // ── Factory functions ─────────────────────────────────────
 
 export function createRoutingBlock(t, options = {}) {
-  const { includeCommands = true } = options;
+  const { includeCommands = true, toolSearchBootstrap = false } = options;
   return `
 <context_window_protection>
   <priority_instructions>
     Every byte a tool returns enters your conversation memory and costs reasoning capacity for the rest of the session. The context-mode tools let you do the work in a sandbox and surface only the derived answer — the raw bytes stay out. Think-in-Code: program the analysis, do not compute it by reading raw data into your conversation.
   </priority_instructions>
-
+${toolSearchBootstrap ? `
+  <deferred_tool_bootstrap>
+    The context-mode tools below may be DEFERRED in your harness — their schemas are not loaded yet, so calling them directly fails (e.g. "tool not found" / InputValidationError). Load them ONCE before your first ctx_* call:
+    ToolSearch(query: "select:${t("ctx_batch_execute")},${t("ctx_search")},${t("ctx_execute")},${t("ctx_execute_file")},${t("ctx_fetch_and_index")}")
+    After that they are callable. If any ctx_* call fails as not-found, ToolSearch it and retry — do NOT fall back to Bash/Read just because the schema was not loaded yet.
+  </deferred_tool_bootstrap>
+` : ''}
   <tool_selection_hierarchy>
     0. MEMORY: ${t("ctx_search")}(sort: "timeline")
        - On resume or compaction, query prior decisions, errors, plans, user prompts before asking the user — auto-captured session memory is searchable.
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "226.7k+",
+  "message": "226.6k+",
   "color": "brightgreen",
   "npm": "197k+",
-  "marketplace": "29.6k+"
+  "marketplace": "29.5k+"
 }
--- a/stats.json
+++ b/stats.json
@@ -1,8 +1,8 @@
 {
   "schemaVersion": 1,
   "label": "users",
-  "message": "226.6k+",
+  "message": "230.7k+",
   "color": "brightgreen",
-  "npm": "197k+",
+  "npm": "201.1k+",
   "marketplace": "29.5k+"
 }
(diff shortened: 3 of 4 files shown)


NgoQuocViet2001:
@beastawakens I sanity-checked this on Windows against current upstream/next by cherry-picking 2c28f30 onto a90d399.

Ran:
- `node node_modules/vitest/vitest.mjs run tests/core/routing.test.ts --reporter=dot`
- `npm run typecheck`

Both passed. The change looks scoped: claude-code Agent prompts get the ToolSearch bootstrap, codex stays unchanged.

One thing I’d still keep as final validation: a live Claude Code Explore run. If Explore does not expose ToolSearch in practice, the fallback from #724 ("skip injection for read-only/search subagents") may still be needed.

PR #728, opened 2026-05-27
fix: normalize stale version paths after CC auto-update (#727)

### LINKED ISSUE #727: Upgrade leaves stale version paths in hooks.json and plugin.json
## Bug

When upgrading context-mode (e.g. from 1.0.146 → 1.0.151), the upgrade process copies the plugin directory to the new version folder but does **not** rewrite the hardcoded absolute paths inside:

- `hooks/hooks.json` — all hook commands still reference the old version path
- `.claude-plugin/plugin.json` — the `mcpServers.context-mode.args` still reference the old `start.mjs` path

The old version directory is cleaned up, so every hook and the MCP server itself point at files that no longer exist.

## Symptoms

After upgrading and restarting:

```
error: Module not found "/home/user/.claude/plugins/cache/context-mode/context-mode/1.0.146/hooks/sessionstart.mjs"
error: Module not found "/home/user/.claude/plugins/cache/context-mode/context-mode/1.0.146/hooks/pretooluse.mjs"
error: Module not found "/home/user/.claude/plugins/cache/context-mode/context-mode/1.0.146/hooks/posttooluse.mjs"
```

`/mcp` reconnect also fails (`-32000`) because `plugin.json` points the MCP server at the old `start.mjs`.

The `ctx-upgrade` command reports "Hooks configured" as success despite not fixing the paths.

## Expected behavior

The upgrade process should find-and-replace the old version path with the new version path in at minimum:
- `hooks/hooks.json`
- `.claude-plugin/plugin.json`

## Workaround

Manually replace the old version string in both files:

```bash
PLUGIN_DIR="$HOME/.claude/plugins/cache/context-mode/context-mode"
OLD="1.0.146"  # whatever the previous version was
NEW="1.0.151"  # the current version

sed -i "s |$OLD|$NEW|g" "$PLUGIN_DIR/$NEW/hooks/hooks.json"
sed -i "s|$OLD|$NEW|g" "$PLUGIN_DIR/$NEW/.claude-plugin/plugin.json"
```

Then restart the session.

## Environment

- Claude Code CLI on Linux (WSL2)
- Bun 1.3.10
- Upgraded from 1.0.146 → 1.0.151

lbzepoqo:
## Summary

Fixes #727 — after Claude Code's native plugin manager creates a new version cache directory by carrying forward files from the old version, `hooks.json` and `.claude-plugin/plugin.json` retain absolute paths baked to the old version. The MCP server can't boot because `plugin.json`'s `mcpServers` args point at a `start.mjs` that no longer exists (chicken-and-egg with Self-heal Layer 5).

Two fixes:

- **Self-heal Layer 1 (forward heal)**: when `start.mjs` detects a newer version dir exists, call `normalizeHooksOnStartup` against the **new** dir before updating the registry. This runs from the old `start.mjs` (which CC can still launch) and fixes the new dir's files before the next session reads them.
- **Cache-heal SessionStart hook**: extend the inline hook to also run `normalizeHooksOnStartup` on existing `installPath`s (not just create symlinks for missing ones). This covers the case where the old dir is already cleaned up. The hook is now re-deployed when its content changes, not just when missing, so existing installs get the updated logic on next MCP boot.

## Root cause

1. `ctx-upgrade` copies new code into the current version dir (e.g. `1.0.146`) and `normalizeHooksOnStartup` bakes absolute paths for that dir — correct at that moment.
2. CC's plugin manager later creates a new `1.0.151` dir carrying forward the files — including `hooks.json` and `plugin.json` with `1.0.146` absolute paths.
3. `start.mjs`'s Self-heal Layer 5 would fix this on boot, but `plugin.json`'s stale `mcpServers.args` path prevents the new `start.mjs` from ever being launched.
4. Self-heal Layer 1 (running from the old dir) detects `1.0.151` and updates the registry, but doesn't normalize the new dir's files.

## Test plan

- [ ] Install context-mode, simulate CC auto-update by creating a new version dir with stale paths copied from old dir
- [ ] Verify `start.mjs` Self-heal Layer 1 normalizes the new dir's `hooks.json` and `plugin.json` on boot
- [ ] Verify cache-heal SessionStart hook normalizes existing installPaths with stale version segments
- [ ] Verify hook is re-deployed when content changes (not just when missing)
- [ ] Verify no regression on clean install (placeholder `${CLAUDE_PLUGIN_ROOT}` still works)

### DIFF
--- a/start.mjs
+++ b/start.mjs
@@ -121,11 +121,33 @@ if (cacheMatch) {
       });
       const newest = dirs[dirs.length - 1];
       if (newest && newest !== myVersion) {
+        // Issue #727: normalize hooks.json + plugin.json in the newest version
+        // dir BEFORE updating the registry. CC's auto-update carries forward
+        // files from the old cache dir, including hooks.json and plugin.json
+        // with absolute paths baked to the old version. start.mjs's Self-heal
+        // Layer 5 would catch this, but plugin.json's stale mcpServers path
+        // prevents the new start.mjs from ever booting — chicken-and-egg.
+        // Fix: normalize from HERE (the old start.mjs that CC CAN still launch)
+        // so the new dir's files are correct before the next session reads them.
+        const newestDir = resolve(cacheParent, newest);
+        try {
+          // #713: use narrow helper — wide normalizeHooksOnStartup would
+          // write plugin.json on the NEW cache dir, the exact #711 poison
+          // vector. Only hooks.json needs the placeholder→absolute rewrite
+          // pre-bump to close the first-hook-fire window.
+          const { normalizeHooksJsonOnly } = await import("./hooks/normalize-hooks.mjs");
+          normalizeHooksJsonOnly({
+            pluginRoot: newestDir,
+            nodePath: process.execPath,
+            platform: process.platform,
+          });
+        } catch { /* best effort — never block startup */ }
+
         const ip = JSON.parse(readFileSync(ipPath, "utf-8"));
         for (const [key, entries] of Object.entries(ip.plugins || {})) {
           if (key !== "context-mode@context-mode") continue;
           for (const entry of entries) {
-            entry.installPath = resolve(cacheParent, newest);
+            entry.installPath = newestDir;
             entry.version = newest;
             entry.lastUpdated = new Date().toISOString();
           }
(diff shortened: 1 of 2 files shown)


mksglu:
Thanks @lbzepoqo. #713 just landed on `next` — it introduces a narrow `normalizeHooksJsonOnly` helper specifically for the "stale version paths" symptom you're also fixing.

Right now `start.mjs:122-138` here calls the wide `normalizeHooksOnStartup`, which writes into the new cache dir's `plugin.json` — that's the exact #711 poison vector #713 documents. With #713 merged, this PR re-introduces the regression on the runtime forward-heal layer.

Please rebase on `next` and switch the Layer 1 forward-heal call to `normalizeHooksJsonOnly` so only `hooks.json` is touched. Also drop the `stats.json` hunk from the diff (CI bot owns that file).

After that the runtime forward-heal layer + #713's call-site fix stack cleanly. Re-ping when rebased.

Issue #621, opened 2026-05-18
[Bug]: ctx_batch_execute handler receives undefined commands parameter from OpenCode adapter


sacrtap:
## Debug Output

```
OS: Darwin 24.6.0 (macOS)
Host: OpenCode (CLI)
Node.js: v25.6.0
context-mode: v1.0.137
Adapter: opencode (detected via ~/.config/opencode/ directory)
```

## The Prompt

Called `ctx_batch_execute` via MCP with valid parameters:

```json
{
  "commands": [
    { "label": "check os", "command": "echo \"test1\" && uname -s" },
    { "label": "current time", "command": "date '+%Y-%m-%d %H:%M:%S'" }
  ],
  "queries": ["batch command result", "execution status"],
  "concurrency": 2
}
```

## Full Error Output

First call error:
```
Batch execution error: commands.map is not a function. (In 'commands.map((cmd) => ({ ... }))', 'commands.map' is undefined)
```

Second call error (different code path):
```
Batch execution error: commands.map is not a function. (In 'commands.map((c) => c.label)', 'commands.map' is undefined)
```

## Steps to Reproduce

1. Install context-mode v1.0.137 (`npm install -g context-mode@latest`)
2. Configure as MCP server in `~/.config/opencode/opencode.json`:
```json
{
  "mcp": {
    "context-mode": {
      "command": "context-mode",
      "transportType": "stdio"
    }
  }
}
```
3. Start opencode session
4. Call `ctx_batch_execute` with valid `commands` and `queries`
5. Error is thrown immediately

## Root Cause Analysis

The preprocessor function `cz` in `server.bundle.mjs` line ~661:

```javascript
function kh(t){
  if(typeof t=="string") try {
    let e=JSON.parse(t);
    if(Array.isArray(e)) return e;
  } catch {}
  return t;
}

function cz(t){
  let e=kh(t);
  return Array.isArray(e) ? e.map((r,n) =>
    typeof r=="string" ? {label:`cmd_${n+1}`,command:r} : r
  ) : e;
}
```

The inputSchema for `ctx_batch_execute` uses:
```javascript
commands: M.preprocess(cz, M.array(M.object({...})).min(1))
```

[...]

Issue #731, opened 2026-05-28
[Bug]: ctx_execute(language: "javascript") fails when OpenCode runs as self-contained binary (opencode.exe)


haisto:
### Platform

OpenCode

### context-mode version

v1.0.151

### Debug script output (REQUIRED)

```json
无
```

### Exact prompt that triggered the bug (REQUIRED)

```text
ctx_execute
```

### Full error output (REQUIRED)

```text
no output
```

### Steps to reproduce (REQUIRED)

无

### What have you tried to fix it?

# ctx_execute(language: "javascript") fails when OpenCode runs as self-contained binary (opencode.exe)

## Summary

`ctx_execute(language: "javascript")` returns `(no output)` when OpenCode is running as a self-contained binary (`opencode.exe`). The JavaScript sandbox executor spawns `opencode.exe` instead of `node.exe`, causing it to fail with `"Failed to change directory to ...script.js"`.

## Environment

- **context-mode version**: v1.0.151
- **Platform**: Windows 10/11 x64
- **OpenCode**: v1.15.11, installed via npm global (`opencode-ai`), running as self-contained `opencode.exe` binary
- **OpenCode process tree**:
  - PID 52228: `opencode.exe` (142MB, PE32+ self-contained binary at `<npm_prefix>\node_modules\opencode-ai\bin\opencode.exe`)
  - `OPENCODE_PROCESS_ROLE=worker`
- **Node.js**: v24.12.0, managed by nvm-windows
  - `F:\development-tools\nodejs\` is an nvm directory junction pointing to the active version
  - `NVM_SYMLINK=F:\development-tools\nodejs`
  - `NVM_HOME=E:\Program Files\nvm`
- **Git Bash**: PATH includes Git Bash (`C:\Program Files\Git\usr\bin`) — used for shell executor

## Symptoms

1. `ctx_execute(language: "javascript", code: 'console.log("test")')` returns `(no output)`
2. `ctx_doctor` shows:
   ```
   [FAIL] Server test: FAIL — exit 0 (Error: Failed to change directory to C:\...\Temp\.ctx-mode-xxx\script.js)
   ```
3. `ctx_execute(language: "shell")` works correctly
4. The child process exits with code 0 but produces no stdout; stderr contains the "Failed to change directory" error in red bold ANSI format (`\x1b[91m\x1b[1mError: \x1b[0m`)

## Root Cause

In `build/runtime.js`, `detectRuntimes()` resolves the JavaScript runtime as:

```javascript
[...]

PR #734, opened 2026-05-28
fix(opencode): convert Zod v3 schemas to v4 for in-process plugin host

maxmaxou2:
## Summary

Recent OpenCode releases (≥ ~1.14.x) bundle Zod v4 in their bun-compiled binary. The existing in-process plugin path registers tool args as raw Zod v3 schemas, which now crashes the OpenCode host on the **first message** with:

```
TypeError: undefined is not an object (evaluating 'n._zod.def')
    at process (/$bunfs/root/chunk-w7ehg0hg.js:22:198266)
    at process (/$bunfs/root/chunk-w7ehg0hg.js:22:201152)
    at Bv (/$bunfs/root/chunk-w7ehg0hg.js:24:2057)
    at <anonymous> (/$bunfs/root/chunk-8xcga49m.js:2059:1004)
    at ~effect/Effect/successCont (/$bunfs/root/chunk-812qsrfz.js:25:7738)
    at runLoop (/$bunfs/root/chunk-812qsrfz.js:25:2045)
    ...
```

The crash is the exact same pattern that #632 fixed for KiloCode in v1.0.142: a Zod v4 host accessing `_zod.def` on a schema built with Zod v3 (which only exposes `_def`).

## Root cause

The v1.0.142 release notes explicitly scoped the v3→v4 conversion layer to KiloCode only, with this note:

> Adds transparent conversion for all Zod types (objects, arrays, unions, optionals, nullables, defaults, effects) scoped to KiloCode only; **OpenCode continues using native Zod v3.**

That assumption no longer holds. OpenCode's in-process plugin host is now on Zod v4, so the same crash class now reproduces on OpenCode with no plugins-side change.

## Fix

Extend the existing KiloCode-only gate to also cover OpenCode. No converter logic changes — `zod3ShapeToV4` already handles objects, arrays, unions, optionals, nullables, defaults, effects.

```diff
- const argsForHost = platform === "kilo"
+ const argsForHost = platform === "kilo" || platform === "opencode"
    ? zod3ShapeToV4(shape as Record<string, unknown>)
    : shape as Record<string, unknown>;
```

The execute path is unchanged (still calls `inputSchema.parse(args)` on the original v3 schema), so #621 preprocessing semantics are preserved.

Updated the `zod3tov4.ts` module header comment to reflect the wider scope (no longer "KiloCode only").

## Reproduction (pre-fix)

1. `npm i -g context-mode@1.0.151`
2. `brew install opencode` (1.14.40 reproduces; older brew bottles likely affected since the OpenCode zod v4 cutover)
3. `~/.config/opencode/opencode.json`:
   ```json
   { "$schema": "https://opencode.ai/config.json", "plugin": ["context-mode"] }
   ```
4. `opencode`, send any prompt → TypeError above on every message.

Workaround (no PR): remove from `plugin` array, register context-mode as an out-of-process `mcp` server — tools work but **PreToolUse routing / SessionStart hooks are lost**, defeating the in-process plugin's whole purpose.

## Tests

Added `tests/opencode-plugin.test.ts` test "converts tool args to Zod v4 when platform is opencode (#_zod.def fix)" — mirror of the existing KiloCode test (#632 line 106), but without setting `KILO_PID`. Asserts every registered tool arg carries the `_zod` v4 marker on the default OpenCode platform.

Validation:

```
bun run typecheck                                            # clean
bun run vitest run tests/opencode-plugin.test.ts \
    tests/adapters/opencode.test.ts \
    tests/adapters/zod3tov4.test.ts
                                                             # 120/120 passed
```

## Affected platforms

- [x] OpenCode
- [ ] (no change for any other adapter)

## Checklist

- [x] Tests added (TDD-equivalent: test fails on `main` without the gate change, passes with it)
- [x] `bun run typecheck` clean
- [x] No converter logic change — reuses #632's proven conversion
- [x] PR against `next`

### DIFF
--- a/src/adapters/opencode/plugin.ts
+++ b/src/adapters/opencode/plugin.ts
@@ -407,20 +407,10 @@ async function createContextModePlugin(ctx: PluginContext) {
             ? (inputSchema._def.shape as () => unknown)()
             : {};
 
-      // Both KiloCode and recent OpenCode releases bundle Zod v4 in their
-      // host runtime. When they receive a tool definition whose `args`
-      // contain Zod v3 schemas (with `_def` but no `_zod.def`), they crash
-      // with `TypeError: undefined is not an object (evaluating 'n._zod.def')`
-      // during the first message handling. The KiloCode-only gate was
-      // introduced in v1.0.142 (#632) under the assumption that OpenCode
-      // would stay on Zod v3 — that assumption no longer holds for OpenCode
-      // ≥ ~1.14.x, which now bundles Zod v4 inside its bun-compiled binary.
-      // Converting v3 → v4 for both platforms is safe: the resulting v4
-      // schemas retain `_def` for legacy callers and add the `_zod.def`
-      // payload that v4 hosts require.
-      const argsForHost = platform === "kilo" || platform === "opencode"
-        ? zod3ShapeToV4(shape as Record<string, unknown>)
-        : shape as Record<string, unknown>;
+      // Both KiloCode and recent OpenCode bundle Zod v4 in-host; v3 schemas
+      // crash with `n._zod.def` undefined. Gate widened from kilo-only (#632)
+      // because every consumer of this file is an OpenCode-family host.
+      const argsForHost = zod3ShapeToV4(shape as Record<string, unknown>);
 
       tools[registered.name] = {
         description: String(config.description ?? ""),
--- a/src/adapters/opencode/plugin.ts
+++ b/src/adapters/opencode/plugin.ts
@@ -407,7 +407,18 @@ async function createContextModePlugin(ctx: PluginContext) {
             ? (inputSchema._def.shape as () => unknown)()
             : {};
 
-      const argsForHost = platform === "kilo"
+      // Both KiloCode and recent OpenCode releases bundle Zod v4 in their
+      // host runtime. When they receive a tool definition whose `args`
+      // contain Zod v3 schemas (with `_def` but no `_zod.def`), they crash
+      // with `TypeError: undefined is not an object (evaluating 'n._zod.def')`
+      // during the first message handling. The KiloCode-only gate was
+      // introduced in v1.0.142 (#632) under the assumption that OpenCode
+      // would stay on Zod v3 — that assumption no longer holds for OpenCode
+      // ≥ ~1.14.x, which now bundles Zod v4 inside its bun-compiled binary.
+      // Converting v3 → v4 for both platforms is safe: the resulting v4
+      // schemas retain `_def` for legacy callers and add the `_zod.def`
+      // payload that v4 hosts require.
+      const argsForHost = platform === "kilo" || platform === "opencode"
         ? zod3ShapeToV4(shape as Record<string, unknown>)
         : shape as Record<string, unknown>;
 
--- a/src/adapters/opencode/zod3tov4.ts
+++ b/src/adapters/opencode/zod3tov4.ts
@@ -1,13 +1,14 @@
 /**
- * Zod 3 → Zod 4 shape conversion (KiloCode only).
+ * Zod 3 → Zod 4 shape conversion for in-process plugin hosts that bundle Zod v4.
  *
- * KiloCode's runtime bundles Zod v4 internally. When it receives plugin tool
- * definitions whose `args` contain Zod v3 schemas (with `_def` but no `_zod`),
- * it crashes with `undefined is not an object (evaluating 'n._zod.def')`.
+ * KiloCode (since its inception) and recent OpenCode releases (≥ ~1.14.x)
+ * bundle Zod v4 internally. When they receive plugin tool definitions whose
+ * `args` contain Zod v3 schemas (with `_def` but no `_zod.def`), they crash
+ * with `TypeError: undefined is not an object (evaluating 'n._zod.def')`.
  *
- * This module converts Zod 3 schema shapes into Zod 4 equivalents so KiloCode
- * can process them natively. Only called when `platform === "kilo"`.
- * OpenCode uses Zod 3 natively and receives the original shapes unchanged.
+ * This module converts Zod 3 schema shapes into Zod 4 equivalents so both
+ * hosts can process them natively. Called when `platform === "kilo"` or
+ * `platform === "opencode"`.
  */
 import z from 'zod/v4';
 
--- a/src/adapters/opencode/zod3tov4.ts
+++ b/src/adapters/opencode/zod3tov4.ts
@@ -7,8 +7,7 @@
  * with `TypeError: undefined is not an object (evaluating 'n._zod.def')`.
  *
  * This module converts Zod 3 schema shapes into Zod 4 equivalents so both
- * hosts can process them natively. Called when `platform === "kilo"` or
- * `platform === "opencode"`.
+ * hosts can process them natively.
  */
 import z from 'zod/v4';
 
@@ -121,7 +120,7 @@ function zod3ToV4(v: unknown, depth = 0): z.ZodType {
       break;
 
     default:
-      // Never leak raw Zod 3 schemas back to KiloCode.
+      // Never leak raw Zod 3 schemas back to a v4 host.
       result = z.unknown();
       break;
   }
--- a/tests/opencode-plugin.test.ts
+++ b/tests/opencode-plugin.test.ts
@@ -103,52 +103,40 @@ describe("ContextModePlugin", () => {
       ]);
     });
 
-    it("converts tool args to Zod v4 when platform is kilo (#_zod.def fix)", async () => {
-      const prevKiloPid = process.env.KILO_PID;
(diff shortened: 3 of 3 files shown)


mikij:
First I can't reproduce this problem in OpenCode. I have tested before with 1.5.x and now with latest 1.5.12 and plugin works for me (I am not getting the message above in Opencode):
```
⚙ ctx_doctor
 
All checks passing. context-mode v1.0.151 is healthy — 8 runtimes available, Bun runtime (fast), FTS5/SQLite working, session hooks active.
```

src/adapters/opencode/plugin.ts
```diff
@@ -407,7 +407,18 @@ async function createContextModePlugin(ctx: PluginContext) {
...
+      // contain Zod v3 schemas (with `_def` but no `_zod.def`), they crash
+      // with `TypeError: undefined is not an object (evaluating 'n._zod.def')`
+      // during the first message handling. The KiloCode-only gate was
+      // introduced in v1.0.142 (#632) under the assumption that OpenCode
+      // would stay on Zod v3 — that assumption no longer holds for OpenCode
+      // ≥ ~1.14.x, which now bundles Zod v4 inside its bun-compiled binary.
+      // Converting v3 → v4 for both platforms is safe: the resulting v4
+      // schemas retain `_def` for legacy callers and add the `_zod.def`
+      // payload that v4 hosts require.
+      const argsForHost = platform === "kilo" || platform === "opencode"
```
mikij: this line should be just:
```
const argsForHost = zod3ShapeToV4(shape as Record<string, unknown>);

```

src/adapters/opencode/zod3tov4.ts
```diff
@@ -1,13 +1,14 @@
...
+ * KiloCode (since its inception) and recent OpenCode releases (≥ ~1.14.x)
+ * bundle Zod v4 internally. When they receive plugin tool definitions whose
+ * `args` contain Zod v3 schemas (with `_def` but no `_zod.def`), they crash
+ * with `TypeError: undefined is not an object (evaluating 'n._zod.def')`.
  *
- * This module converts Zod 3 schema shapes into Zod 4 equivalents so KiloCode
- * can process them natively. Only called when `platform === "kilo"`.
- * OpenCode uses Zod 3 natively and receives the original shapes unchanged.
+ * This module converts Zod 3 schema shapes into Zod 4 equivalents so both
+ * hosts can process them natively. Called when `platform === "kilo"` or
```
mikij: Remove this last centence: Called when...

tests/opencode-plugin.test.ts
```diff
@@ -126,6 +126,30 @@ describe("ContextModePlugin", () => {
       }
     });
 
+    // Mirror of the KiloCode test above, but for the default OpenCode platform.
```
mikij: now we don't need additional test but to update current one as it should work the same for both platforms

mikij (changes requested):

mikij:
> First I can't reproduce this problem in OpenCode. I have tested before with 1.5.x and now with latest 1.5.12 and plugin works for me (I am not getting the message above in Opencode):
> 
> ```
> ⚙ ctx_doctor
>  
> All checks passing. context-mode v1.0.151 is healthy — 8 runtimes available, Bun runtime (fast), FTS5/SQLite working, session hooks active.
> ```

About this I don't understand, but old version and new version seems work locally for me the same. I mean in both versions I did not see any error. This is confusing. I would like more to see the same error you get in Opencode but for some reason I don't see it, and I need to rely on finding that is not my own too. 

As I am thinking more about it, I remember original PR about using tools directly for OpenCode plugin worked in OpenCode and did not work in KiloCode. That is why I introduced this conversion from v3 to v4 for KiloCode. I just can't accept claim this happened from some 1.14.xxx version of OpenCode as I don't see any proof for it.

src/adapters/opencode/plugin.ts
mikij: not sure we need this comment

murataslan1:
I tested the OpenCode side again.

OpenCode 1.15.12 with context-mode 1.0.151 works on my side. The plugin loads, ctx tools register, ctx_doctor runs, and I could not reproduce the Zod runtime error.

mksglu:
WDYT? @maxmaxou2

mksglu:
Thanks @maxmaxou2 — your archaeology is right: upstream OpenCode dev branch IS Zod 4.1.8 catalog-wide. We re-verified against the freshly pulled `refs/platforms/opencode/` (branch=dev):

- `refs/platforms/opencode/package.json:75` — `"zod": "4.1.8"` in catalog
- `refs/platforms/opencode/packages/plugin/package.json` — plugin SDK v1.15.13 resolves zod via catalog → 4.1.8
- `refs/platforms/opencode/packages/plugin/src/tool.ts:1` — `import { z } from "zod"` (resolves to v4)
- `refs/platforms/opencode/packages/core/src/github-copilot/**/*.ts` — explicit `import { z } from "zod/v4"` calls

The directional fix is correct. @mikij's four CHANGES_REQUESTED nits (also valid) are the remaining work:

1. **Collapse the `platform === "kilo" || platform === "opencode"` gate** to an unconditional `zod3ShapeToV4(...)` — `plugin.ts` only runs under OpenCode-family hosts; the platform check has no semantic value once both branches converge.
2. **Trim the verbose inline comment block** — one line citing #632 is enough.
3. **Parametrize the existing KiloCode test** to cover both platforms instead of adding a sibling test that duplicates the structure.
4. **Drop the trailing "Called when..." sentence** from `zod3tov4.ts` docstring.

Separately, the empirical reproducer gap is worth flagging: both @mikij and @murataslan1 tested OpenCode 1.15.12 + context-mode 1.0.151 and could not reproduce the crash. That doesn't invalidate the fix — refs proves OpenCode upstream IS on Zod v4 — but a reliable reproducer (OpenCode version + bun version + exact command that triggers the crash) baked into the test suite would pin the regression class.

NB: #731 (ctx_execute fails when OpenCode runs as `opencode.exe`) is a **separate** Windows-specific bug — `process.execPath` resolving to `opencode.exe` instead of `node.exe`, unrelated to Zod versions. This PR doesn't address it, doesn't need to.

Once the four nits are in, ready to land.

mksglu:
Pushed `b66a385` to address the four review nits. Direction is correct — refs confirm OpenCode's catalog now pins `zod 4.1.8` (`refs/platforms/opencode/package.json:75`), and the plugin SDK + tool registry both consume the v4 catalog entry, so widening the v3→v4 conversion is the right call.

**Changes in this commit (review feedback only — no behavior change beyond what c2ecdb6 already shipped):**

1. **Collapsed the gate** at `src/adapters/opencode/plugin.ts:407` from `platform === "kilo" || platform === "opencode"` to an unconditional `zod3ShapeToV4(...)` call. `plugin.ts` only runs under OpenCode-family hosts and `getPlatform()` (lines 230-241) returns exclusively `"kilo"` or `"opencode"` — both now need conversion, so the disjunction is dead semantics.

2. **Trimmed the inline comment block** to a 3-line citation of #632. Long-form rationale stays in the commit body and the `zod3tov4.ts` module docstring.

3. **Parametrized the existing KiloCode test** in `tests/opencode-plugin.test.ts:106` via `it.each([{platform:"kilo",...}, {platform:"opencode",...}])` — both invariants asserted from a single test body, no duplicated structure.

4. **Dropped the trailing \"Called when...\" sentence** from `src/adapters/opencode/zod3tov4.ts` docstring. Also fixed a stale \"back to KiloCode\" comment in the default switch case → \"back to a v4 host\" (kept on-theme with the wider scope).

**Validation:**
- `npx vitest run tests/opencode-plugin.test.ts -t \"_zod.def\"` → 2/2 passed
- `npx tsc --noEmit` → clean

**Archaeology:** preserves `45100c75` (#632) — the kilo-only gate stays correct (still converts kilo), just widens to cover the new OpenCode v4 host runtime.

---

@maxmaxou2 — quick ask: could you share a concrete reproducer (exact OpenCode version, bun version, exact command/prompt that triggers the crash)? Both @mikij and @murataslan1 reported they couldn't reproduce on OpenCode 1.15.12, so we'd like to add a regression test that pins the actual crash trigger rather than just asserting the `_zod` marker invariant. A minimal failing-case repo or even just the `opencode --version && bun --version` + first-message contents would help us lock this down.

@mikij — ready for re-review when you have a minute.

maxmaxou2:
Hi, sorry for not responding I was quite busy this week.
Here was my env config :
```
opencode:      1.14.40
bun:           1.3.14
node:          v26.0.0
macOS:         15.7.7  (arm64)
context-mode:  1.0.146 at least (i'm not sure)  (npm install via opencode plugin loader)
```
Just opening a new opencode session, writing a prompt in build mode like "Say hey" and pressing enter would trigger the error.

I've tested again on opencode version 1.15.13 combined with context-mode current latest release 1.0.153  and everything works well now, thank you for the merge !

PR #747, opened 2026-06-01
Harden config writes, deny-matching, and DoS/SSRF surfaces against hostile input

kerneltoast:
This is a security hardening pass over the surfaces that take attacker-influenceable input: config paths a cloned repo controls, deny-list patterns and commands, fetched response bodies, the local insight dashboard, and session ids that land in filesystem paths. Every fix ships with a regression test that drives the real code path (not a source grep), so it fails if the guard is reverted.

## Symlink-escape on config writes

Several adapters write cwd-relative config (`.cursor/hooks.json`, `opencode.json`, `.github/hooks/context-mode.json`, `openclaw.json`). A cloned or malicious repo can plant a symlink at that path, so a plain `writeFileSync` (flag `"w"` follows symlinks) would truncate an arbitrary file such as `~/.bashrc`.

- Add `writeProjectConfigSafely`, which lstat-checks the immediate parent dir and the final file and refuses any whose real target escapes the project root. An in-root symlink (a dotfile manager's intra-repo link) is followed normally, since it can only ever land inside the project.
- Route every adapter config write through it, and guard the extra ancestor for Copilot's two-level `.github/hooks` path.
- Mirror the same guard on `backupSettings` (an escaping symlink would otherwise copy a secret into a repo-local `.bak`) and on the install/upgrade and plugin-heal paths (realpath the install child paths and the plugin root before writing or importing through them).
- A symlinked config is a warn-and-skip during setup rather than a fatal abort, so a stow/chezmoi user isn't blocked over it.

## Command and file-path deny matching

- The chained-command splitter only split on `; && || |` and treated backticks as opaque, so a denied command could hide after a newline, a lone `&`, a subshell, a brace group, or a `$()`/backtick substitution and never be checked against deny. Split aggressively on all of those (deny is evaluated per-segment while allow/ask match the whole command, so over-splitting is fail-safe), and fix a backslash-counting bug where an even run of backslashes let a closing quote read as escaped.
- The `caseInsensitive` deny default was win32-only, so on macOS `RM -rf` slipped past a `Bash(rm:*)` deny and still executed. Both command and file-path denies now track the platform default (win32 || darwin).

## DoS and resource bounds

- `ctx_fetch` buffered the whole decompressed body via `resp.text()`, so a small gzip/br response advertising a tiny Content-Length could inflate the subprocess heap toward gigabytes before any post-read check fired. Stream the body and bail the moment the running byte count crosses the cap.
- Cap the glob-to-regex compilers: a deny glob (from project-local `.claude/settings*.json`, which a hostile clone controls) or a `ctx_index` include/exclude pattern with many wildcards compiles to a chain of `.*` groups that backtracks catastrophically. An over-cap pattern compiles to a never-match regex instead of stalling the long-lived server.
- Cap `ctx_index` `maxDepth`/`maxFiles` at the schema layer, so a `path:"/"` request can't drive the walk into resource exhaustion.

## Network and override containment

- Reject non-loopback Host headers on the insight dashboard to stop DNS rebinding. The dashboard binds 127.0.0.1, but without Host validation a malicious page whose hostname rebinds to 127.0.0.1 can read every analytics endpoint (and issue the DELETE) from the user's browser.
- Contain the `ctx_insight` sessionDir/contentDir overrides: `path.resolve` doesn't dereference symlinks, so a symlink planted under the containment root but pointing outside it passed the lexical gate, then the spawned server followed it out. Realpath the dir and re-check against the canonical root.

## Injection and session-id sanitization

- Run `codesign` via argv (no shell) instead of a shell string, and escape `$`/backtick in the build-tool redirect, so a `binaryPath` or command carrying shell metacharacters can't inject.
- Sanitize `CLAUDE_SESSION_ID` and `CONTEXT_MODE_SESSION_SUFFIX` (allowlist) before they reach the SQLite DB filename, since a value with a path separator or `..` would otherwise write the file outside the sessions dir.

## Testing

The new regression tests are behavioral: they call the real write/eval/fetch handlers and assert the attacker-controlled victim is untouched (or the request is rejected), so they catch a revert to the unsafe form rather than just pinning source text. All affected suites pass.


kerneltoast:
@mksglu Could you please explain why you closed this PR?

Issue #649, opened 2026-05-20
[BUG]: Hardcoded storage path for many platforms


jetnet:
## Summary

There is no way to relocate context-mode's storage (SessionDB + ContentStore) without either changing the host platform's own config-dir env var or patching the source. For several platforms the storage location is entirely hardcoded:

| Platform | Current storage root | Configurable? |
|---|---|---|
| Claude Code | `$CLAUDE_CONFIG_DIR/context-mode/` | ✅ via `CLAUDE_CONFIG_DIR` |
| OpenCode | `$XDG_CONFIG_HOME/opencode/context-mode/` | ✅ via `XDG_CONFIG_HOME` / `APPDATA` |
| VS Code Copilot | `.github/context-mode/` or `~/.vscode/context-mode/` | ✅ via `.github/` presence |
| **Pi** | `~/.pi/context-mode/` | ❌ hardcoded in `PiAdapter` → `BaseAdapter` |
| **OMP** | `~/.omp/context-mode/` | ❌ hardcoded |
| **Gemini CLI** | `~/.gemini/context-mode/` | ❌ hardcoded |
| Cursor, Codex, … | `~/.<platform>/context-mode/` | ❌ hardcoded |

`BaseAdapter.getSessionDir()` always roots to `homedir()` with no override path:

```typescript
// src/adapters/base.ts
getSessionDir(): string {
  const dir = join(homedir(), ...this.sessionDirSegments, "context-mode", "sessions");
  mkdirSync(dir, { recursive: true });
  return dir;
}
```

## Why this matters

- **Slow or small home dir**: home dirs on NFS mounts or space-constrained system volumes are common in CI and dev-container setups. Redirecting to `/data/context-mode` or a project-local path is blocked today.
- **Multi-user machines**: each user gets their own home, but shared CI agents often share a workspace. No way to isolate or redirect storage without touching platform config.
- **Pi specifically**: Pi has no config-dir env var that context-mode honors. `~/.pi/context-mode/` is the only option regardless of disk layout.

## Proposed change

Add `CONTEXT_MODE_DATA_DIR` as a universal override checked first in `BaseAdapter.getSessionDir()` and `BaseAdapter.getConfigDir()`:

```typescript
// src/adapters/base.ts
getSessionDir(): string {
  const root = process.env.CONTEXT_MODE_DATA_DIR
    ? resolve(process.env.CONTEXT_MODE_DATA_DIR)
    : join(homedir(), ...this.sessionDirSegments);
[...]

PR #755, opened 2026-06-01
feat(opencode): point opencode/kilo plugin at the npm-global build via file://

### LINKED ISSUE #754: opencode/kilo setup forks a per-package plugin copy that "npm i -g" cannot update
## Summary

`context-mode setup` registers the OpenCode / KiloCode plugin by adding the bare npm
name `"context-mode"` to the host's `plugin` array:

```jsonc
// opencode.json
{ "plugin": ["context-mode"] }
```

On first launch the host resolves that bare name as an npm package and fetches a
**separate per-package copy** into its own cache:

```
~/.cache/opencode/packages/context-mode@latest/node_modules/context-mode
~/.cache/kilo/packages/context-mode@latest/node_modules/context-mode
```

That cached copy is completely independent of the `context-mode` **npm-global**
install. So the documented update path —

```
npm i -g context-mode@latest
```

— refreshes the global build but leaves OpenCode / Kilo running the **stale cached
plugin**. The only way to pick up the new version is to manually clear the cache
(`rm -rf ~/.cache/opencode/packages/...`) or wait for the host to bump its own cache key.

## Why this is the odd one out

Every other in-process host already loads a **single source of truth**:

- the 11 npm-global platforms run the global build directly;
- the Pi / OMP wrappers load it through a thin `index.js` that re-exports
  `file://<pluginRoot>/build/adapters/pi/extension.js`.

OpenCode and Kilo are the only hosts that fork a second, independently-versioned copy.

## Impact

- **Stale tools/hooks after update.** `ctx_*` tools and the PreToolUse / SessionStart
  hooks in OpenCode / Kilo keep running the previous version after `npm i -g
  context-mode@latest`, with no signal to the user.
- **Version skew on one machine** — the global build and the per-package cache copy
  drift apart.
- **Disk** — a full duplicate of the package per host.

## Steps to reproduce

1. `npm i -g context-mode@<old>`
2. `context-mode setup` (opencode) → `opencode.json` gets `"plugin": ["context-mode"]`
3. Launch `opencode` once → a copy is fetched into
   `~/.cache/opencode/packages/context-mode@latest/node_modules/context-mode`
4. `npm i -g context-mode@latest`
5. Launch `opencode` again → the `ctx_*` surface still reports the **old** version.

(Reproduces on KiloCode identically — `@kilocode/cli` vendors the same plugin loader.)

## Expected

`context-mode setup` should point OpenCode / Kilo at the **same npm-global build**
the other hosts use, so a single `npm i -g context-mode@latest` refreshes every host.

## Environment

- OpenCode 1.15.13, KiloCode (`@kilocode/cli`) 7.3.16
- context-mode: `next`
- OS-independent — the per-package fetch is host behaviour, not platform-specific.

## Proposed fix

OpenCode's loader already supports path specs
(`packages/opencode/src/plugin/shared.ts`):

```ts
const isPathPluginSpec = (spec: string) =>
  spec.startsWith("file://") || spec.startsWith(".") || isAbsolutePath(spec)
```

A `file://` pointer to the npm-global build's `plugin.js` is sent through
`resolvePathPluginTarget` and imported directly — no npm fetch, no second copy.
`isAbsolutePath` already accepts Windows drive letters (`/^[A-Za-z]:[\\/]/`), so the
pointer is cross-platform. PR to follow.

> PRs > Issues — fix attached.

ken-jo:
## What / Why / How

`context-mode setup` registered OpenCode / KiloCode by adding the bare npm name
`"context-mode"` to the host's `plugin` array. OpenCode — and Kilo, which vendors the
same loader — resolves a bare name as an npm package and fetches a **separate
per-package copy** into `~/.cache/<platform>/packages/context-mode@latest/node_modules/context-mode`,
independent of the `context-mode` npm-global install. So `npm i -g context-mode@latest`
— the documented update path — refreshes the global build but leaves OpenCode / Kilo
running the stale cached copy.

Fixes #754.

This PR makes setup write a `file://` pointer to the **single npm-global build**
instead of the bare name:

```diff
- "plugin": ["context-mode"]
+ "plugin": ["file:///<npm-global>/context-mode/build/adapters/opencode/plugin.js"]
```

so both hosts load the exact build every other host already uses. This mirrors the
single-source-of-truth approach the Pi / OMP wrappers take (their `index.js`
re-exports `file://<pluginRoot>/build/adapters/pi/extension.js`). One
`npm i -g context-mode@latest` now refreshes every in-process host with no
per-package re-fetch.

## Root cause

OpenCode's plugin loader (`packages/opencode/src/plugin/shared.ts`) branches on the
spec shape:

```ts
const isPathPluginSpec = (spec: string) =>
  spec.startsWith("file://") || spec.startsWith(".") || isAbsolutePath(spec)
```

- A **bare name** (`"context-mode"`) is treated as an npm package → resolved and
  fetched into the per-package cache.
- A **path spec** (`file://…`) goes through `resolvePathPluginTarget`, which imports
  the file directly — no npm, no second copy.

KiloCode (`@kilocode/cli`) vendors this loader identically. context-mode's setup only
ever emitted the bare name, so both hosts always forked a cache copy that `npm -g`
could not reach.

## Fix

1. **`src/adapters/opencode/index.ts` — `configureAllHooks()`** now writes a `file://`
   pointer to the npm-global build:

   ```ts
   const pointer = pathToFileURL(
     resolve(pluginRoot, "build", "adapters", "opencode", "plugin.js"),
   ).href;
   ```

   Any prior context-mode entry (bare name **or** a stale `file://` pointer — both
   contain `"context-mode"`) is dropped and replaced, so re-running `setup` / `upgrade`
   migrates an existing install in place. The legacy MCP-block cleanup is unchanged.

2. **`getInstalledVersion()`** now derives the version from the pointer's package root,
   keeping the legacy `~/.cache/<platform>/node_modules/context-mode/package.json` read
   as a fallback for installs that predate this change.

3. **`src/cli.ts`** — removed `cachePluginRoot()`; `getPluginRoot()` now returns
   `defaultPluginRoot()` for every platform. The per-package cache root was only needed
   by the old bare-name model; nothing else in the tree referenced `cachePluginRoot`.
   `isInProcessPluginPlatform` is still imported and used by the doctor / heal branch.

### Windows

The pointer is built with `pathToFileURL`, which emits a valid `file:///C:/…` URL, and
`isAbsolutePath` already accepts drive letters (`/^[A-Za-z]:[\\/]/`), so the path spec
loads on native Windows. No backslash paths are ever written. Verified on Windows for
both OpenCode and KiloCode.

### Caveat (documented)

The pointer is an absolute, machine-specific path — correct for a per-machine config,
but it must **not** be committed in a shared project `opencode.json`.
`docs/platform-support.md` now documents both the portable manual form
(`["context-mode"]`) and the setup-written pointer, and recommends a global
`~/.config/<platform>/` config for sharing setups across machines.

## Reproduction (pre-fix)

1. `npm i -g context-mode@<old>`
2. `context-mode setup` (opencode) → `opencode.json` gets `"plugin": ["context-mode"]`
3. Launch `opencode` once → a copy lands in
   `~/.cache/opencode/packages/context-mode@latest/node_modules/context-mode`
4. `npm i -g context-mode@latest`
5. Launch `opencode` → `ctx_*` still runs the **old** version until
   `rm -rf ~/.cache/opencode/packages/...`.

## Tests

- `tests/adapters/opencode.test.ts` — updated the 4 plugin-array assertions to the
  `file://` pointer value; added
  *"migrates a bare 'context-mode' npm name to a file:// pointer at the npm-global
  build (single source of truth)"*.
- `tests/core/cli.test.ts` — assert `cachePluginRoot` is gone, `getPluginRoot` returns
  `defaultPluginRoot()`, and the OpenCode adapter uses `pathToFileURL` with the
  `build/adapters/opencode/plugin.js` path.
- Removed `tests/core/cache-plugin-root.test.ts` (obsolete — it pinned the deleted
  `cachePluginRoot`).

## Validation

- `npx tsc --noEmit` → clean
- `npx vitest run tests/adapters/opencode.test.ts` → **47 passed**
- `npx vitest run tests/core/cli.test.ts` → **216 passed**
- Live runtime: OpenCode **1.15.13** and KiloCode **7.3.16** both load the `file://`
  pointer and register the `ctx_*` tools from the real
  `build/adapters/opencode/plugin.js`.

## Affected platforms

- [x] OpenCode
- [x] KiloCode
- [ ] Other platforms (no change)

## Checklist

- [x] Tests added (TDD red → green)
- [x] `npm test` passes (opencode + cli suites; pre-existing env-flaky cases unrelated
      to this change)
- [x] Docs updated (`docs/platform-support.md`)
- [x] No Windows path regressions — forward slashes / `file://` only
- [x] Targets `next`

> Bundles (`*.bundle.mjs`) are intentionally **not** committed — they are regenerated
> by the `bundle.yml` CI workflow.


ken-jo:
Thanks for the review — both points addressed in the latest commit.

**1. Over-broad plugin matching (`src/adapters/opencode/index.ts`).**
Replaced the `includes("context-mode")` substring test with a precise
`isContextModePluginEntry()` helper that matches **only**:
- the exact bare npm name (`"context-mode"`, optionally version-tagged `context-mode@…`), or
- a `file://` pointer whose path ends at this package's `build/adapters/opencode/plugin.js`.

So `configureAllHooks` no longer drops sibling plugins whose names merely contain
the substring (e.g. `@acme/context-mode-helper`). The same helper now backs the
`hasContextModePlugin` doctor check and `getInstalledVersion`, so detection and
removal stay consistent. Added a regression test
(*"does NOT drop sibling plugins whose names merely contain 'context-mode'"*).

**2. Hard-coded POSIX `file://` URLs in tests.**
The expected pointer is now built the same way production builds it —
`pathToFileURL(resolve(pluginRoot, "build", "adapters", "opencode", "plugin.js")).href`
— via a small `opencodePointer()` helper, so the assertions yield
`file:///C:/…` on Windows and match the adapter output instead of failing on
the windows-latest runner.

Validation: `tsc --noEmit` clean, `tests/adapters/opencode.test.ts` (48) and
`tests/core/cli.test.ts` (216) green.

ken-jo:
Follow-up: the windows-latest run surfaced a deeper portability issue in the
test expectations, now fixed.

The first attempt computed the expected pointer with `pathToFileURL(resolve(...))`.
That still failed on Windows because these tests run `configureAllHooks` in a
**spawned subprocess** (cwd under the OS temp dir, on `C:`) while the expectation
was computed in the **test process** (cwd at the repo checkout, on `D:`). A bare
`/tmp/...` plugin root is drive-relative on Windows, so `resolve` produced
`file:///C:/...` in the subprocess but `file:///D:/...` in the test:

```
- Expected: "file:///D:/tmp/plugin/build/adapters/opencode/plugin.js"
+ Received: "file:///C:/tmp/plugin/build/adapters/opencode/plugin.js"
```

The assertions now check the pointer's **shape** — a `file://` URL ending at
`.../build/adapters/opencode/plugin.js` (`expect.stringMatching`) — which is
drive- and OS-agnostic, exactly the platform-agnostic alternative suggested in
the review. (Production is unaffected: `defaultPluginRoot()` is always a real
drive-qualified absolute path.)

mikij:
I have a remark for you as a author of this PR @ken-jo and for you @mksglu. OpenCode/KiloCode plugin does not require global installation of context-mode since we introduced context mode [native tool support](https://github.com/mksglu/context-mode/commit/09efefdc4a6cf248e70a9b7f2f1ada5dfd613cce). That is clearly stated in README of context-mode. So I don't see we need this PR as only proper way to upgrade or install plugin for OpenCode/KiloCode is through plugin array in config file and do upgrades through call of `ctx upgrade` inside of tool afterwards. Having context-mode installed as global separate tool is redundant for OpenCode/KiloCode users.

ken-jo:
Closing this — @mikij is right, and thanks for the careful review.

My actual goal here was **data integration**: letting a multi-agent CLI user share a single context-mode knowledge base / sessions / memory across agents. After digging into the code, this change does not serve that goal:

- The DB / sessions / memory location is keyed by the **platform config dir** (or the `CONTEXT_MODE_DATA_DIR` override, #649) — **not** by where the plugin code lives (`pluginRoot`). So pointing OpenCode/Kilo at the npm-global build via a `file://` URL does not change which DB they use: the per-package copy and the global build resolve the *same* per-platform storage either way. And different platforms (Claude Code vs OpenCode vs Kilo) stay on separate storage roots **by design**, so a shared code copy doesn't merge them.
- As you note, OpenCode/Kilo don't need a global install at all: `plugin: ["context-mode"]` is self-contained, and `ctx upgrade` (clone + build + in-place update of the loaded copy) is the supported upgrade path. The `file://` pointer would only couple these hosts to a global install and bake a **non-portable absolute path** (breaks on nvm node-version switches, `npx`, and shared/committed configs). The per-package copy is one user-global copy per platform, self-managed and refreshed by `ctx upgrade` — not a real problem.

So the premise (`npm i -g` leaving the host stale) was the wrong upgrade path for these hosts, and the "single source of truth" framing doesn't unify the DB.

The correct mechanism for a shared cross-agent knowledge base already exists: **`CONTEXT_MODE_DATA_DIR`** — point every agent at one data dir and sessions/memory/KB converge. That's currently undocumented, so I'll pursue documenting it (and possibly a `setup`/`doctor` hint) instead.

Thanks again — closing in favor of the self-contained `["context-mode"]` install + `CONTEXT_MODE_DATA_DIR` for integration.

PR #758, opened 2026-06-01
feat(737): global / local ctx_search scope — project: current | global | <abs-path>

### LINKED ISSUE #737: [Feature]: ctx_search `project:` filter — scope FTS5 ContentStore to current project when using a global DB
## Platform

Pi / OpenClaw (Pi Agent) — but applies to all platforms using `CONTEXT_MODE_PROJECT_DIR` to opt into a global shared DB.

## Feature description

When opting into a global knowledge base via `CONTEXT_MODE_PROJECT_DIR=$HOME` (or any single shared path), `ctx_search` has no way to scope results to the current project. The only user-facing filter parameters today are `source` (label partial-match), `sort`, `contentType`, and `limit`.

The problem: auto-captured session memory — indexed under fixed source labels like `session-events`, `project/<file>`, `user/<file>`, `memory/<file>` — cannot be namespaced by project because those labels are generated automatically and contain no project identifier. Using `source: "myproject/"` only works for content explicitly indexed by the user with a matching prefix; it silently misses all auto-memory.

The internal plumbing already supports per-project filtering at every layer:
- `SessionDB.searchEvents(query, limit, projectDir, source)` — `projectDir` param exists and filters correctly
- `searchAutoMemory(queries, limit, projectDir, configDir, adapter)` — `projectDir` param exists
- `ContentStore` chunks carry `session_id` / `event_id` FK attribution (written via `index({ attribution })`) which links back to a session that has a `project_dir`

The gap is that `ctx_search` never surfaces `projectDir` as a user-facing parameter, so it is inaccessible from MCP tool calls.

## Use case

I want a single global knowledge base so I can search across all my projects in one `ctx_search` call when I need cross-project recall. But I also want to narrow results to just the current project during normal work — without having to maintain per-project DB configs or restart sessions.

Concrete workflow:

```
# default: search only current project
ctx_search(queries: ["apk install"])
# → only returns results from /home/jet/src/temp

# explicit global search
ctx_search(queries: ["apk install"], project: "global")
# → returns results from all projects
```

Today the only workaround is either:
1. Stay on per-project DBs (lose cross-project recall entirely), or
2. Use the global DB and get undifferentiated results from all projects mixed together

Related: #663 fixed cross-project auto-memory leakage in *per-project DB mode*. This feature is the complementary ask for *global DB mode* — intentional cross-project sharing with opt-in per-project scoping.

## Proposed solution

Add an optional `project` parameter to `ctx_search`:

```typescript
project?: "current" | "global" | string
// "current" (default when global DB is active) → filter to current session's projectDir
// "global"                                     → no filter, search all projects (today's behaviour)
// "/path/to/project"                           → filter to that specific project
```

When `CONTEXT_MODE_PROJECT_DIR` is not set (default per-project DB mode), the parameter is a no-op — the DB is already isolated by hash, so every search is implicitly project-scoped.

Implementation sketch:

1. **`ContentStore` chunk attribution** — the `attribution` FK (`session_id`, `event_id`) already written on indexed chunks can join to `session_events.project_dir`. A `WHERE project_dir = ?` filter on the FTS5 join would scope FTS5 results without schema changes beyond adding the join.

2. **`searchAllSources` in `src/search/unified.ts`** — already receives `projectDir` in `SearchAllSourcesOpts` and passes it to `sessionDB.searchEvents` and `searchAutoMemory`. The missing piece is forwarding it to the ContentStore search path (Source 1 in `searchAllSources`).

3. **`ctx_search` tool handler in `src/server.ts`** — add `project` to the Zod schema, resolve `"current"` to `getProjectDir()`, `"global"` to `undefined`, and thread the resolved value through to `searchAllSources`.

Backwards compat: when `project` is omitted the behaviour is identical to today.

jetnet:
## Summary

Implements the maintainer-approved design for #737 — a real `project:` scope on `ctx_search` delivering both project-local and true cross-project (global) recall, with **zero migration**.

```
ctx_search(queries: [...])                       # current project (default)
ctx_search(queries: [...], project: "global")    # fan-out across ALL projects
ctx_search(queries: [...], project: "/abs/path") # a specific project
```

### What was broken (all confirmed in the issue thread)
1. `project: "global"` reached only 1 of 3 sources (ContentStore); session events + auto-memory ignored the scope.
2. Default `sort: "relevance"` skipped session memory entirely (only timeline searched it).
3. `current`/default filtered by the **pinned** dir in shared mode, not the real cwd → silent 0 results.
4. Per-project mode physically opens only the current DB — `global` was meaningless without fan-out across DB files.

### What this PR does
- **Decouples** the row-level filter from DB-file selection; adds `getCurrentWorkingProject()` (real cwd) so `current` is correct in shared mode.
- **Relevance now searches all three sources** (content + session events + auto-memory); `sort` controls ordering only.
- **Global fan-out** (`src/search/global-fanout.ts`): unions every `sessions/*.db` + `content/*.db`, merges with **RRF** (k=60), **read-only** opens (no WAL pragmas / schema init / migration / corruption repair), bounded by `CONTEXT_MODE_GLOBAL_FANOUT_MAX` (default **1024**).
- **`project:"/abs/path"` scope**: shared mode → row filter; per-project mode → open that project's hashed DBs read-only (incl. worktree-suffixed session DBs).
- **Attribution**: each result header names its origin — `proj:` + `sess:`; orphan content in global is labelled `proj:(unattributed)`.
- **Coverage notice**: if the fan-out cap drops DBs, a visible "Global search INCOMPLETE" line reports `searched X of Y` and the exact value to raise — never silently partial.
- **Result limits**: breadth scopes return 10–12 hits/query (so specific answers aren't buried behind common-term noise); single-DB stays 1–2. Hard 40KB output cap enforced incrementally.
- **Snippet detail**: forward-biased window so explanation after a heading match is captured.

### Dedupe / fusion fix from Sweep review
The branch has been updated against `next` (base `7b2d7ca`; PR head `230c433`). The focused global-fanout suite now passes locally on the PR head.

- `src/search/global-fanout.ts:itemKey()` keys global dedupe by `title|content.length|fnv1a32(full content)`, not a title/source/prefix key.
- Byte-identical cross-project hits fuse into one result and accumulate RRF score.
- Distinct documents with the same title and shared first 80+ characters no longer false-fuse.
- Focused coverage is in `tests/core/global-fanout.test.ts`:
  - RRF merge ordering / fusion: lines 239–293.
  - Full-content dedupe regression: lines 1085–1118.

### Backwards compatibility
Zero migration. Per-project users are unaffected unless they pass `project: "global"`; shared-DB users keep working. The one intentional default-behaviour change is the relevance-mode source-set fix (#2 above).

### Tests
Validation re-run after updating the branch from `next`:

- `npx vitest run tests/core/global-fanout.test.ts --reporter=verbose` → **40 passed / 0 failed**.
- `npx vitest run tests/core/global-fanout.test.ts tests/core/ctx-search-plan.test.ts tests/core/search-project-filter.test.ts tests/core/search.test.ts tests/core/auto-memory-adapter.test.ts --reporter=verbose` → **262 passed / 0 failed**.
- `npm run build` → **passed** (`tsc`, bundle, `assert-bundle`, `assert-asymmetric-drift`). Generated bundles were restored afterward and are intentionally not included.
- Independent fresh-context reviewer gate → **PASS**, no blockers on dedupe/fusion/RRF behavior.

Validated earlier end-to-end against a real ~170-project install (read-only confirmed: 0 DB files mutated).

### Notes for reviewers
- Bundles are intentionally **not** included (gitignored; regenerated at release via `prepublishOnly`).
- New env var `CONTEXT_MODE_GLOBAL_FANOUT_MAX` is documented in README ("Search environment variables").

Closes #737


jetnet:
Sweep review follow-up:

- Updated the PR branch against `next` (base `7b2d7ca`, PR head `230c433`).
- Re-ran focused global fan-out validation on the PR head:
  - `npx vitest run tests/core/global-fanout.test.ts --reporter=verbose` → 40 passed / 0 failed.
  - Dedupe/fusion/RRF cases specifically passed: multi-DB identical fusion, RRF ordering, full-content dedupe regression.
- Re-ran targeted suites:
  - `global-fanout`, `ctx-search-plan`, `search-project-filter`, `search`, `auto-memory-adapter` → 262 passed / 0 failed.
- Re-ran `npm run build` → passed (`tsc`, bundle, `assert-bundle`, `assert-asymmetric-drift`). Generated bundles restored afterward and remain out of the PR.
- Updated the PR description with the current dedupe/fusion behavior and test evidence.
- Independent fresh-context reviewer gate: PASS, no blockers on dedupe/fusion/RRF behavior.

mksglu:
Hi @jetnet! Thanks for that. Let's revisit this later. Even though I believe this is the real solution, we need to work on it from a structural perspective.

PR #763, opened 2026-06-02
fix(security): harden shell command evaluation and fix test isolation

pg-adm1n:
## Summary

Hardens the shell security policy evaluator against command-screening bypasses in chained commands, nested command substitutions, and generated runtime bundles.

## Root cause

The previous implementation handled only top-level command separators and simple `$()` / backtick extraction. That left several paths where a denied command could be hidden inside shell syntax that Bash still executes:

- newline and single-ampersand command separators were not consistently treated as independent statements
- `$()` / backtick command substitutions were not evaluated recursively against deny rules
- operators inside `$()` could be split before the full subshell was extracted, e.g. `$(sudo rm -rf / | cat)`
- escaped-substitution handling only checked the previous character, so even backslashes before `$()` were misclassified as escaped even though Bash still executes the substitution
- the shipped `hooks/security.bundle.mjs` could remain stale while marketplace/runtime hooks load that bundle first

## Fix

- Split chained commands on `;`, pipes, `&&`, `||`, newlines, carriage returns, and single `&`, while respecting quotes, backticks, `$()` depth, nested parentheses, and backslash parity.
- Extract command substitutions from `$()` and backticks, including nested command substitutions and substitutions inside arithmetic expansion.
- Evaluate command elements recursively so `evaluateCommand()` and `evaluateCommandDenyOnly()` both inspect subshell internals segment-by-segment.
- Preserve arithmetic expansion as non-command execution unless it contains a nested command substitution.
- Default command/file matching to case-insensitive behavior on macOS as well as Windows.
- Isolate the strict project-dir integration test from the developer's real Codex home.
- Rebuild tracked runtime artifacts: `server.bundle.mjs`, `cli.bundle.mjs`, and `hooks/security.bundle.mjs`.
- Refresh session/runtime bundles after merging the current `next` branch.

## Affected platforms

- Claude Code / hook-based installs that load `hooks/security.bundle.mjs`
- MCP server deny-only evaluation paths
- macOS and Windows case-insensitive policy matching
- Linux/macOS/Windows test/build matrix

## Test plan

- `npm run build`
- `npx vitest run tests/security.test.ts`
- `npm test` — 183 files passed, 4222 passed / 28 skipped on the current PR head after merging `next`
- Manual runtime probe against `hooks/security.bundle.mjs` for:
  - `git status $(sudo rm -rf /)`
  - `git status $(sudo rm -rf / | cat)`
  - `git status $(printf '(x)' | sudo rm -rf /)`
  - `git status \\$(sudo rm -rf /)` with even backslashes before `$()`
  - `echo ok\nsudo rm -rf /`
  - `SUDO rm -rf /`
  - arithmetic expansion without command substitution remains allowed
  - command substitution inside arithmetic expansion is denied

## Checklist

- [x] Tests added/updated (TDD: red -> green)
- [x] `npm test` passes
- [x] `npm run typecheck` passes via `npm run build`
- [x] Docs updated if needed (not needed; security/runtime behavior only)
- [x] No Windows path regressions (no path separator changes)


ken-jo:
Hi @pg-adm1n, I’m a contributor helping with review/triage.

I reviewed this against the maintainer validation requirements: not only whether CI is green, but whether the PR description matches a real behavior gap, whether the history supports the root cause, and whether the patch changes the actual runtime behavior it claims to fix.

## What I checked

I reviewed the PR claim around shell command hardening, especially nested command substitutions, chained/piped commands, newline-separated commands, deny-only behavior, case-insensitive matching, and whether the hook bundle behavior matched the source change.

I also ran git archaeology on the relevant security path:

- `git log --follow --all -- src/security.ts`
- `git log --follow --all -- tests/security.test.ts`
- `git log -S 'evaluateCommand' -- src/security.ts tests/security.test.ts`
- `git log -S 'command substitution' -- src/security.ts tests/security.test.ts hooks/security.bundle.mjs`

The history supports this as a continuation of the existing shell security hardening work rather than a cosmetic or test-only change.

## Validation performed

On a fresh checkout of this PR head, I ran:

- `npm run build` passed
- `npx vitest run tests/security.test.ts tests/integration/project-dir-strict.test.ts --reporter=dot` passed: 129 tests

I also ran direct runtime smoke checks against `hooks/security.bundle.mjs`. Those checks confirmed that deny-only evaluation rejects:

- nested `$()` payloads containing denied commands
- piped command substitutions
- newline-separated denied commands
- case-insensitive deny matches

The same smoke check kept arithmetic expansion allowed, which is important because this should close the bypass without broadly blocking normal shell syntax.

## Conclusion

I think this is a meaningful security fix and has merge value.

The current blocker is merge readiness. GitHub still reports this PR as conflicted, and the local merge probe showed generated bundle conflicts:

- `cli.bundle.mjs`
- `hooks/security.bundle.mjs`
- `server.bundle.mjs`

Could you please rebase/resolve the current conflicts and align the generated artifact changes with the maintainer’s bundle-file policy, then let CI run again?

pg-adm1n:
@ken-jo  hi i’m currently on vacation. Will be handle it on 20/06

pg-adm1n:
Conflicts with `next` are resolved in head `2b2694b`. The generated CLI, server, and security bundles were rebuilt from the merged source; the stale generated-bundle drift is gone.

Local validation on macOS:
- `npm run build` — passed
- `npx vitest run tests/security.test.ts tests/integration/project-dir-strict.test.ts --reporter=dot` — 129/129 passed

The new CI and OpenClaw E2E runs were created, but GitHub marked both `action_required` because this public-fork workflow needs maintainer approval. Please use **Approve workflows to run**; I will monitor the runs and address any failures.

PR #786, opened 2026-06-05
feat: GitHub Copilot CLI integration

### LINKED ISSUE #775: [Feature]: Add official GitHub Copilot CLI support docs for Context Mode MCP and hooks
### Platform

Other (please specify)

### Summary

I would like official documentation for using `context-mode` with **GitHub Copilot CLI**, specifically covering both:

1. MCP server installation/configuration
2. Hook integration and expected behavior

The MCP configuration path seems relatively straightforward, but the hook story is less clear. Since Context Mode’s value depends not only on exposing tools, but also on routing large outputs, preserving useful context, and integrating with agent lifecycle events, it would be helpful to know whether Copilot CLI is officially supported and what the recommended hook setup should be.

I am using Copilot CLI on Windows, but this request is primarily about official Copilot CLI support in general.

### Request

Could you please add official documentation for GitHub Copilot CLI covering:

1. Whether GitHub Copilot CLI is officially supported by Context Mode.
2. The recommended MCP server configuration for Copilot CLI.
3. Whether MCP-only setup is sufficient or whether hooks are required for full Context Mode behavior.
4. Which Copilot CLI hook events Context Mode supports or expects.
5. The exact hook configuration users should add.
6. Where user-level hook files should be placed.
7. How hook commands should invoke Context Mode.
8. How to verify that hooks are firing correctly.
9. Known limitations compared with VS Code Copilot, Claude Code, Codex, Cursor, or other clients.

### Why hooks need explicit documentation

MCP-only registration appears to expose Context Mode tools to the agent, but it is not obvious whether that provides the full Context Mode experience.

For example, users need to know whether Copilot CLI hooks are required for:

* intercepting or summarizing large tool outputs;
* preserving important context before compaction;
* indexing repository state at session start;
* routing expensive or noisy commands through Context Mode;
* avoiding context-window pollution from logs, test output, build output, grep output, and other large terminal results;
* maintaining continuity across longer CLI sessions.

Without official hook guidance, users may configure the MCP server successfully but still miss the main benefits of Context Mode.

### What I tried / inferred

Based on Copilot CLI MCP configuration conventions, I expected an MCP config similar to this:

```json
{
  "mcpServers": {
    "context-mode": {
      "type": "local",
      "command": "context-mode",
      "args": [],
      "env": {},
      "tools": ["*"]
    }
  }
}
```

However, I am not sure whether this is the recommended command, whether additional arguments are needed, or whether the command should be invoked through a full path, `npx`, or a global npm install.

More importantly, I am not sure what the officially supported hook configuration should be for Copilot CLI.

For example, should users configure hooks for events such as:

```text
SessionStart
PreToolUse
PostToolUse
PreCompact
```

And should those hooks invoke Context Mode using commands such as:

```text
context-mode hook vscode-copilot sessionstart
context-mode hook vscode-copilot pretooluse
context-mode hook vscode-copilot posttooluse
context-mode hook vscode-copilot precompact
```

Or should Copilot CLI have its own Context Mode hook target, for example:

```text
context-mode hook copilot-cli ...
```

### Desired documentation

It would be helpful to have an official section like:

#### GitHub Copilot CLI

Install Context Mode:

```sh
npm install -g context-mode
context-mode doctor
```

Configure MCP:

```json
{
  "mcpServers": {
    "context-mode": {
      "type": "local",
      "command": "context-mode",
      "args": [],
      "env": {},
      "tools": ["*"]
    }
  }
}
```

Configure hooks:

```json
{
  "version": 1,
  "hooks": {
    "SessionStart": [
      {
        "type": "command",
        "command": "context-mode hook copilot-cli sessionstart"
      }
    ],
    "PreToolUse": [
      {
        "type": "command",
        "command": "context-mode hook copilot-cli pretooluse"
      }
    ],
    "PostToolUse": [
      {
        "type": "command",
        "command": "context-mode hook copilot-cli posttooluse"
      }
    ],
    "PreCompact": [
      {
        "type": "command",
        "command": "context-mode hook copilot-cli precompact"
      }
    ]
  }
}
```

Also document how to verify:

```text
/mcp show
/mcp show context-mode
ctx doctor
ctx stats
```

And how to confirm that hooks are actually running, not merely that MCP tools are available.

### Why VS Code Copilot support is not enough

VS Code Copilot support does not fully answer the Copilot CLI use case.

Copilot CLI has a different runtime model, different configuration locations, and potentially different hook behavior. A setup that works in VS Code does not necessarily prove that Context Mode is correctly integrated with Copilot CLI’s terminal-agent workflow.

Terminal-first users often rely on Copilot CLI for:

* running tests and builds;
* inspecting logs;
* searching large repositories;
* refactoring from the shell;
* using temporary or remote terminal sessions;
* working without opening VS Code;
* keeping MCP tools available across many repositories.

These workflows can produce large volumes of output, which is exactly where Context Mode’s hook-based behavior may matter most.

### Open questions

* Is Copilot CLI officially supported?
* Is MCP-only setup enough, or are hooks required for full functionality?
* Should Copilot CLI use the existing `vscode-copilot` hook target, or should there be a separate `copilot-cli` target?
* Which hook events are supported and recommended?
* What is the correct hook file format?
* What is the correct user-level hook file location?
* How can users verify that hooks are executing correctly?
* Are there known limitations in Copilot CLI around output interception, compaction, indexing, or session continuity?

Thanks for the project. Official Copilot CLI documentation, especially around hooks, would make Context Mode much easier to adopt for terminal-first Copilot users.

### Use case

See above

### Proposed solution

See above

khoffdbh:
## Summary

Adds first-class **GitHub Copilot CLI** support to context-mode, alongside the existing VS Code Copilot and JetBrains Copilot integrations — hook-based routing/session-continuity plus MCP-server operation.

Closes #775 
---
Sadly I'm unable to do a final test with Copilot because my monthly budget is reached.
Maybe I'll catch a colleagues over the weekend.
---

## What's included

- **Platform + detection** — `copilot-cli` PlatformId; env-var detection keyed on `COPILOT_CLI=1` (verified against `github/copilot-cli` `changelog.md:1124`), listed before `claude-code` so the shared `CLAUDE_PLUGIN_ROOT` can't cause misdetection; `~/.copilot` config-dir probe + `COPILOT_HOME` override.
- **Adapter** — `CopilotCliAdapter` (session storage under `~/.copilot/context-mode/…`, `COPILOT_HOME`/`CONTEXT_MODE_DATA_DIR` overrides, project-dir-relative hook config + settings path).
- **Hooks** — `pretooluse` / `posttooluse` / `precompact` / `sessionstart` wrappers, `COPILOT_CLI_OPTS`, dispatcher wiring (`context-mode hook copilot-cli <event>`).
- **Output schema** — formatter emits Copilot CLI's **flat** hook-output schema (`permissionDecision` / `permissionDecisionReason` / `modifiedArgs`, top-level `additionalContext`) — verified against the official `docs.github.com` hooks reference; **not** the VS Code `hookSpecificOutput` wrapper.
- **Config + docs** — `configs/copilot-cli/hooks.json`, `docs/copilot-cli.md`, updated `docs/platform-support.md`.
- **Tests** — adapter unit coverage (session/project-dir/validateHooks/version), hook integration suite, formatter, detection, tool-naming, and matrix coverage.

## Verification

- `npx tsc --noEmit` clean.
- Full suite: 4258 passing; only the 4 pre-existing env failures (3× Windows `symlink` EPERM in `tests/core/cli.test.ts`, 1× Rust `link.exe` in `tests/executor.test.ts`) — zero new failures.
- Session storage confirmed isolated to temp dirs (no `~/.copilot` pollution in tests).
- Reviewed via a parallel multi-agent pass (platform-claim anti-hallucination, git archaeology on the detection registry, cross-OS/Windows, adapter parity + config safety, TDD coverage) + architect sign-off.

## Known limitations (preview)

- **MCP tool-name prefix** (`context-mode_<tool>`) and **`clientInfo.name`** are **unconfirmed** against a live Copilot CLI session — not documented, but not contradicted. MCP-handshake detection is currently inert (env detection carries it). Recommend treating copilot-cli as **preview** until confirmed against a live session.
- The three Copilot-family adapters (vscode/jetbrains/copilot-cli) share `.github/hooks/context-mode.json`; re-running `upgrade` for one overwrites another's hook commands. Pre-existing limitation, extended to a third adapter — candidate for a future guard.


ken-jo:
@mksglu I reviewed this together with #787, and I recommend not landing #786 as a standalone PR.

#786 overlaps with the Copilot CLI work in #787, but #787 is the broader/current consolidation target because it covers the Copilot CLI + Antigravity CLI platform integration surface more completely.

My validation notes:

- #786 had some positive targeted test coverage, and a real Copilot CLI MCP smoke test was able to return the expected tool result.
- However, the same run showed that the plugin path was not loaded as a complete plugin surface, the context-mode hooks exited with failures, and no session events were persisted. So #786 is not a complete working integration by itself.
- #787 had stronger live validation on Linux: Copilot CLI loaded the plugin config and exposed `ctx_execute`, and Antigravity CLI also exercised MCP/hooks with persisted events.
- I also checked #787 on Windows. The build passed and targeted Copilot/Antigravity tests passed, but the run still hit an existing Windows baseline shell relative-path failure that also reproduces on `next`. So #787 still needs cleanup before merge, but it is the right PR to carry the final integration.

Recommendation: please fold the useful #786 changes into #787 and close/rebase #786 rather than merging it independently.

Before #787 is merge-ready, I would still expect:

- conflict/changes-requested state resolved
- source/history citations for the Copilot/Antigravity behavior being encoded
- repeated live Copilot/Antigravity smoke validation after the final rebase
- Windows result documented clearly, including the known baseline path issue if still present

This keeps the Copilot CLI support as one coherent feature PR instead of shipping a partial duplicate first.

Issue #774, opened 2026-06-03
[Feature]: Antigravity-cli (agy) better integration


kurokirasama:
### Platform

Antigravity

### Feature description

With the force migration from gemini-cli to agy, I'm in the process of migrating all my tools, skills, configs and mcp to agy.

With respect to context-mode, I had to manually migrate the hooks to `~/.gemini/config/hooks.json`, but agy recognize it and everything, but i noticed that the agent is not as enforced to use the server tools, as it happens with gemini-cli. In some sessions it never use them, in others it used them but not as much as i see it doing in gemini-cli.

Not sure if it's an agy issue or the mcp, but at least it would be great to have instructions on how to install it in agy in the README.

### Use case

For all people migrating from gemini-cli to agy

kurokirasama:
report:

```bash
context-mode setup antigravity-cli
┌   context-mode setup
│
●  Platform: antigravity-cli  (high confidence — explicit --platform / arg)
│
▲  Routing fidelity: best-effort (~60%) — antigravity-cli has no hook surface; routing relies on a rules file
│
●  Hooks: not applicable for this platform
│
◆  Antigravity CLI mcp_config.json mcpServers: wrote context-mode — /home/kira/.gemini/config/mcp_config.json
│
└  Setup complete. Run `context-mode doctor` to verify.
```

```bash
context-mode doctor
┌   context-mode doctor 
│
●  Platform: Claude Code (medium confidence — ~/.claude/ directory exists)
│
◇  Storage paths ────────────────────────────────────────────────────────╮
│                                                                        │
│  sessions: /home/kira/.claude/context-mode/sessions (adapter default)  │
│  content:  /home/kira/.claude/context-mode/content (adapter default)   │
│  stats:    /home/kira/.claude/context-mode/sessions (adapter default)  │
│                                                                        │
├────────────────────────────────────────────────────────────────────────╯
│
◆  Storage session: PASS — /home/kira/.claude/context-mode/sessions (adapter default)
│
◆  Storage content: PASS — /home/kira/.claude/context-mode/content (adapter default)
│
◆  Storage stats: PASS — /home/kira/.claude/context-mode/sessions (adapter default)
[?25l│
◇  Diagnostics complete
[?25h│
◇  Runtimes ────────────────────────────────────────────────────────────────╮
│                                                                           │
│    JavaScript: /home/kira/.bun/bin/bun (1.2.16) ⚡                        │
[...]

PR #765, opened 2026-06-02
fix: pin Claude ctx_execute cwd to session cwd

NgoQuocViet2001:
Fixes #756.

Reported by @chrisjenx. cc @mksglu

## What changed
- Add an optional `cwd` argument to `ctx_execute` and `ctx_batch_execute`, and pass it through to shell execution.
- Update Claude Code PreToolUse routing to resolve the active project dir from the hook payload (`cwd` / `workspace_roots`) instead of stale `CLAUDE_PROJECT_DIR`.
- Automatically pin shell `ctx_execute` / `ctx_batch_execute` calls to the active session cwd when no explicit `cwd` is provided.
- Preserve explicit `cwd` values and keep non-shell languages in their existing temp-dir sandbox.

## Validation
- `node_modules/.bin/vitest.CMD run tests/hooks/core-routing.test.ts tests/hooks/integration.test.ts tests/core/server.test.ts -t "MCP execute|MCP batch_execute|runBatchCommands|ctx_execute_file: CONTEXT_MODE_PROJECT_DIR"`
- `npm run typecheck`
- `npm run bundle`
- `npm run assert-bundle && npm run assert-asymmetric-drift`
- `git diff --check`

Issue #406, opened 2026-05-03
[Feature]: Server-enforced timeout policy for long-running build commands in execute tools


wax911:
### Platform

OpenCode

### Feature description

Long-running build commands (Gradle/Maven/SBT) routed through context-mode execute tools can still time out too early when caller-side timeout values are missing or too low.  
I’d like timeout handling to be enforced in MCP server behavior (not only via prompt guidance), with:

- A host-configurable default timeout (via env var) with sane fallback.
- A server-side timeout floor for known long-running build commands in shell mode.
- Consistent behavior across single execution and batch execution paths (serial + parallel).
- Timeout-aware user guidance in routing and skills docs.

This makes behaviours reliable even when clients/prompts forget to pass explicit timeout values.

### Use case

When running build-heavy workflows (for example Gradle test/build/verify) through routed execute calls, current generic defaults can terminate valid jobs prematurely.  
In real workflows this creates false failures, retries, and confusion because:

- Caller-side prompts are not always explicit/consistent.
- Different tool paths (`ctx_execute`, `ctx_execute_file`, `ctx_batch_execute`) may behave differently.
- Build commands often need much longer than generic command timeouts.

A server-enforced policy would make long-running tasks dependable by default while still allowing explicit overrides.

### Proposed solution

Implement timeout normalization and configurable defaults directly in the server:

1. Add a default-timeout resolver with env override:
   - Env: `CONTEXT_MODE_DEFAULT_TIMEOUT_MS`
   - Fallback: 300000 ms (5 min)

2. Add shell build timeout normalization:
   - Detect Gradle/Maven/SBT shell commands.
   - Apply a floor of 1800000 ms (30 min) for those commands.
   - Keep explicit longer timeouts unchanged.
   - Do not apply floor to background mode.

3. Apply this consistently in:
   - `ctx_execute`
   - `ctx_execute_file` default schema timeout
   - `ctx_batch_execute` default schema timeout
   - Batch serial and parallel execution internals.

4. Keep routing guidance aligned:
   - Include timeout hint for build-tool redirection.
   - Preserve opt-out env escape hatches for native build execution.

5. Add/maintain tests to lock behavior for:
   - default-timeout resolution
   - timeout normalization
   - serial batch shared-budget behavior with build-floor handling
   - parallel batch per-command timeout handling
   - routing timeout hints + opt-out env handling.

Base commit SHA used for stable references:
78b2428a3b4d4f601edac680ae97af688b4dcb6e

[...]

PR #787, opened 2026-06-06
Add GitHub Copilot CLI + Antigravity CLI (agy) support

ken-jo:
## What / Why / How

This PR adds first-class support for **two new CLI agents** plus the cross-cutting fixes they both need. Resolves [#774](https://github.com/mksglu/context-mode/issues/774) (Antigravity CLI / agy integration) and [#775](https://github.com/mksglu/context-mode/issues/775) (official GitHub Copilot CLI support).

> [!NOTE]
> 🙏 **Apology / context.** My first attempt (#779) didn't actually work end-to-end — `agy` was mis-detected as Claude Code, and the agy MCP tools never reached the model. It also pulled in `setup` machinery that wasn't wanted. This PR is narrow (only the two adapters + the cross-cutting fixes they need) and I've now validated it on both Linux and a real Windows machine (agy 1.0.5–1.0.6 + GitHub Copilot CLI). Sorry for the earlier churn — please re-test this one using the install steps below.

**What changed:**

- **GitHub Copilot CLI adapter** (`src/adapters/copilot-cli/`) — `CopilotCliAdapter` with `COPILOT_HOME`-aware config/session paths (`mcp-config.json`, `hooks/context-mode.json`, session DB root all follow `COPILOT_HOME`), a `version: 1` flat hook schema (the CLI runtime requires it or hooks never fire), top-level decision output (`permissionDecision`/`permissionDecisionReason`/`modifiedArgs`/`additionalContext` — not the VS Code `hookSpecificOutput` wrapper), and six event dispatchers (`PreToolUse` routes/vetoes via the routing decision; `PostToolUse`, `SessionStart`, `UserPromptSubmit`, `Stop`, and `PreCompact` capture/restore through the shared session DB). The TS server's session reads are kept aligned with where the hook runtime writes so relocating `COPILOT_HOME` never orphans sessions.
- **Copilot CLI plugin bundle** (`configs/copilot-cli/`) — a one-command `copilot plugin install mksglu/context-mode:configs/copilot-cli` that registers the MCP server, the routing skill, and all six capture hooks in a single shot (no follow-up `context-mode upgrade` or agent round-trip). The bundled `.mcp.json` pins `CONTEXT_MODE_PLATFORM=copilot-cli` so detection short-circuits to the explicit-override tier and self-identifies as Copilot even when Claude Code is co-installed. A single `.gitignore` negation (`!configs/copilot-cli/.mcp.json`) un-ignores exactly that one committed config, since a Copilot plugin has no other way to declare MCP.
- **Antigravity CLI (agy) adapter + plugin** (`src/adapters/antigravity-cli/`, `configs/antigravity-cli/`) — `AntigravityCliAdapter` subclasses the IDE `AntigravityAdapter` but targets agy's distinct global layout (config dir `~/.gemini/antigravity-cli`, hooks `~/.gemini/config/hooks.json`). agy's plugin system is Claude-compatible, so the bundle is `.claude-plugin/plugin.json` (identity + skill) + a root `.mcp.json` + `hooks/hooks.json`. `agy plugin install <bundle>` registers all of it in one command — verified on agy 1.0.6 it logs `mcpServers : 1 processed` and writes the server (env preserved) into its plugin profile `~/.gemini/config/plugins/context-mode/mcp_config.json`. The bundle's `.mcp.json` pins `CONTEXT_MODE_PLATFORM=antigravity-cli`, so the server self-identifies as agy — fixing the #774 mis-detection at the MCP level, not only via the dir-marker probes. The cross-platform Node installer (`npm run install:agy`, replacing the former bash-only script) wraps the install to also clear agy's stale per-server tool-schema cache (`~/.gemini/antigravity-cli/mcp/context-mode/`, which agy never refreshes) so the `ctx_*` tools reappear, and to self-verify the MCP registration. Like the Copilot bundle, this `.mcp.json` is the one committed instance, un-ignored by a single `.gitignore` negation. agy honors no PreToolUse stdout veto in auto-run mode, so the routing skill is the primary enforcement and the hook is capture-only.
- **Gemini-safe schema sanitizer** (`src/server.ts`) — `sanitizeSchemaForStrictClients()` rewrites the emitted `tools/list` schema so Gemini's function-calling API (used by **Antigravity CLI (agy) and Gemini CLI**) accepts it: every `const: X` becomes `enum: [X]` and `additionalProperties` is stripped. Both rewrites are behavior-preserving (a one-value enum equals const; `additionalProperties` is advisory because every `ctx_*` handler re-parses args with Zod). Without this, Gemini-family hosts **silently drop** every `ctx_*` tool and the agent falls back to hand-rolling MCP through Bash — so this restores `ctx_*` exposure on the existing **Gemini CLI** platform too, not just on the two new CLIs. `installStrictClientSchemaCompat()` wraps the SDK's `tools/list` handler best-effort (try/catch), so an SDK refactor degrades rather than breaks.
- **Detection + fail-open dispatch** (`src/adapters/detect.ts`, `src/adapters/client-map.ts`, `src/cli.ts`) — new config-dir marker probes run *before* the generic `~/.claude`/`~/.gemini` fallbacks (the core of #774: a gemini-cli→agy migration leaves both dirs, so `~/.claude` otherwise mis-detected agy as Claude Code), and `client-map.ts` routes the `clientInfo` name (`agy`→`antigravity-cli`, `GitHub Copilot CLI`→`copilot-cli`). The copilot-cli probe only matches a context-mode-*written* file (`mcp-config.json` or `hooks/context-mode.json`), never a bare `~/.copilot`, so a co-installed-but-unconfigured Copilot user stays on Claude Code. `hookDispatch()` now **fails open** (`exit(0)` instead of `exit(1)`) on an unknown platform/event, because a version-skewed older global binary that exits 1 with empty stdout makes Copilot CLI 1.0.59 treat it as a hook error and deny *every* tool ("Denied by preToolUse hook (hook errored)"), bricking the agent.
- **Supporting glue** — a tolerant JSONC parser (`src/util/jsonc.ts`, `stripJsonComments`/`parseJsonc`) that both new adapters import to read agy/Copilot configs (which ship comments) without false-failing `JSON.parse`; the copilot-cli top-level decision formatter and tool-prefix entries (`hooks/core/formatters.mjs`, `hooks/core/tool-naming.mjs`); per-platform `COPILOT_OPTS`/`ANTIGRAVITY_CLI_OPTS` capture wiring (`hooks/session-helpers.mjs`); and `package.json` + `scripts/version-sync.mjs` updates so the agy plugin manifest version stays in sync.

> **Source note (same as #779):** GitHub Copilot CLI is source-available (not OSS); the installed agy is a binary with no published source. So agy/copilot behavior was validated by running the real binaries and capturing MCP traffic, not by reading their source.

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [x] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms
- [x] GitHub Copilot CLI
- [x] Antigravity CLI (agy)

(Gemini CLI is checked because the schema sanitizer restores `ctx_*` tool exposure on every Gemini-family function-calling host, including the existing Gemini CLI platform — not only the two new CLIs.)

## Test plan

**Linux — unit/integration suite (`npm test`):**

- 5 new test files: `tests/adapters/copilot-cli.test.ts` (adapter unit: json-stdio paradigm, COPILOT_HOME-aware paths, flat `version:1` idempotent hook config, snake_case input parsing, top-level decision output, `checkPluginRegistration` remediation), `tests/adapters/copilot-cli-plugin.test.ts` (bundle: committed `.mcp.json` that `git check-ignore` does **not** ignore, `CONTEXT_MODE_PLATFORM=copilot-cli` env pin, six-event `hooks.json` version=1), `tests/adapters/antigravity-cli-plugin.test.ts` (agy bundle: committed env-pinned `.mcp.json` that `git check-ignore` does not ignore, no manifest `mcpServers`, capture-only PostToolUse with no PreToolUse, schema-cache clear), `tests/core/strict-client-schema.test.ts` (sanitizer: `const`→`enum`, `additionalProperties` stripped, nested recursion, input not mutated), `tests/hooks/copilot-cli-capture.test.ts` (spawns the real `userpromptsubmit`/`stop`/`precompact` dispatchers and asserts SQLite persistence — incl. the Stop `createHash(undefined)` regression guard).
- 7 extended files: `tests/adapters/antigravity.test.ts` (full `AntigravityCliAdapter` suite, idempotent capture hook preserving unrelated hooks), `tests/adapters/detect.test.ts` + `tests/adapters/detect-config-dir.test.ts` (clientInfo `agy`/`GitHub Copilot CLI` routing, #774 ordering — dedicated markers beat `~/.claude`/`~/.gemini`, bare `~/.copilot` does **not**), `tests/core/cli.test.ts` (fail-open: unmapped hook exits 0), `tests/integration/project-dir-strict.test.ts`, `tests/scripts/version-sync.test.ts`, `tests/session/multi-adapter-stats.test.ts` (15→17 adapters).
- **Build artifacts:** the committed esbuild bundles (`server.bundle.mjs`, `cli.bundle.mjs`, `hooks/session-db.bundle.mjs`, `hooks/session-extract.bundle.mjs`, `hooks/security.bundle.mjs`) were regenerated so the published runtime carries the sanitizer + both new adapters/hooks.

**Windows — live validation (real binaries):**

- **Copilot CLI plugin — real remote install:** `copilot plugin install <owner>/<repo>:configs/copilot-cli` (from a repo's default branch) extracted the bundle's `.mcp.json` + `hooks.json` + routing skill (`copilot plugin list` then shows `context-mode`), and `ctx_execute` returned **42** — the documented one-command install verified end-to-end against a real GitHub install, not only a local proxy.
- **Copilot CLI plugin — hooks fire:** loading the same bundle via `copilot --plugin-dir` (local), the `PostToolUse` hook captured a **non-MCP bash tool's** output into the session DB — direct proof that hooks fire, not just that MCP loads.
- **Env pin:** with the bundle's `CONTEXT_MODE_PLATFORM=copilot-cli`, `ctx_upgrade`/detection resolved `copilot-cli` and left the co-installed Claude Code config untouched.
- **Copilot CLI manual path** (`copilot mcp add` + `context-mode upgrade` writing the `version:1` hooks file) also verified end-to-end.
- **Antigravity CLI** via `npm run install:agy` (→ `agy plugin install configs/antigravity-cli`) on agy 1.0.6: the bundle's `.mcp.json` registered `mcpServers : 1 processed` (env `CONTEXT_MODE_PLATFORM=antigravity-cli` preserved) into agy's plugin profile, alongside the routing skill + capture hook, with the stale tool-schema cache cleared. **Runtime confirmed:** `agy -p "...ctx_execute...7 + 5..." → 12` — the plugin-scoped MCP resolves in a real agy session with **no** global-profile entry. `context-mode doctor` recognizes that plugin-scoped registration (MCP + hook), not just the global paths. The earlier (pre-#774) failure mode was agy mis-detected as Claude Code with its `ctx_*` tools dropped; both are fixed (env pin + Gemini-safe schema). Note: agy has no `owner/repo` remote install (only a local path or a full git URL, which clones the repo *root* and skips a subdirectory bundle), so the local `npm run install:agy` is the supported install — unlike Copilot's `owner/repo:path`.

### GitHub Copilot CLI
```bash
npm install -g context-mode
# Recommended — one command (MCP + 6 hooks + routing skill), no agent round-trip:
copilot plugin install mksglu/context-mode:configs/copilot-cli
# restart copilot, then verify:
copilot -p "Use the context-mode ctx_execute MCP tool to compute 6 times 7." --allow-all   # -> 42

# Alternative (no plugin) — register MCP + the version:1 hooks file manually:
copilot mcp add context-mode --env CONTEXT_MODE_PLATFORM=copilot-cli -- context-mode   # MCP -> ~/.copilot/mcp-config.json
CONTEXT_MODE_PLATFORM=copilot-cli context-mode upgrade                                  # hooks -> ~/.copilot/hooks/context-mode.json
```
- The `--env CONTEXT_MODE_PLATFORM=copilot-cli` pin (and running `context-mode upgrade` with that var set, or from inside a Copilot session) keeps detection on `copilot-cli` even when Gemini CLI / Claude Code are co-installed. Without the pin, detection falls back to the `~/.copilot/mcp-config.json` marker probe, which still resolves correctly but is more fragile when other CLI config dirs are present.
- `COPILOT_HOME` relocates the hook config, MCP config, and the context-mode session-DB root.
- `copilot doctor` equivalent: `context-mode doctor` should report platform `copilot-cli` and the 6 hooks.

### Antigravity CLI (`agy`)
```bash
npm install -g context-mode                     # the plugin's MCP server runs the global binary
git clone https://github.com/mksglu/context-mode.git
cd context-mode
npm run install:agy                             # cross-platform Node — Windows/macOS/Linux, no bash
# restart agy, then verify:
agy -p "Use the context-mode ctx_execute MCP tool to compute 7 + 5. Answer only the number." --dangerously-skip-permissions   # -> 12
```
`npm run install:agy` wraps `agy plugin install configs/antigravity-cli` — which registers the MCP server (from the bundle's `.mcp.json`), the routing skill, and the PostToolUse capture hook in one step — then clears agy's stale tool-schema cache (`~/.gemini/antigravity-cli/mcp/context-mode/`) so the model sees the current `ctx_*` tools, and self-verifies that the MCP landed in agy's plugin profile. Manual fallback if a future agy build skips it: add `{"command":"context-mode"}` under `mcpServers` in `~/.gemini/config/mcp_config.json`.

## Checklist

- [x] `npm test` passes

<details>

The agy installer was rewritten from bash to cross-platform Node specifically so it runs on Windows; the Copilot CLI adapter and both capture hooks use `path`-based joins and event-based stdin throughout, and the Windows live validation above exercised both new platforms on a real Windows machine.

</details>


ken-jo:
## Update — GitHub Copilot CLI plugin bundle (`configs/copilot-cli/`)

Per the review note, Copilot CLI now installs as a **one-command plugin** (no `context-mode upgrade` / agent call), alongside the existing manual `copilot mcp add` path:

```sh
copilot plugin install mksglu/context-mode:configs/copilot-cli
```

The bundle ships the files real Copilot marketplace plugins use:
- `.mcp.json` — the MCP server, pinned with `CONTEXT_MODE_PLATFORM=copilot-cli`
- `hooks.json` — the six capture hooks (byte-equivalent to what `context-mode upgrade` writes to `~/.copilot/hooks/`)
- `skills/context-mode/SKILL.md` — the routing skill
- `.github/plugin/plugin.json` — manifest

**Why the env pin matters:** on a machine where context-mode is also installed as a Claude Code plugin, `~/.claude/plugins/installed_plugins.json` makes platform detection resolve `claude-code` — so even `ctx_upgrade` would write Claude's config instead of Copilot's. Pinning `CONTEXT_MODE_PLATFORM=copilot-cli` in the plugin's `.mcp.json` makes the server self-identify as Copilot, so detection/`ctx_upgrade` resolve `copilot-cli`.

**Why this `.mcp.json` is committed:** a Copilot plugin discovers MCP only from a root `.mcp.json`, so `.gitignore` un-ignores exactly `configs/copilot-cli/.mcp.json`. The repo-wide `.mcp.json` ignore from #253/#531 guards the repo-root dev file, not a plugin's own config.

### Windows validation (real Copilot CLI)
Loaded the local bundle as a plugin via `copilot --plugin-dir <bundle>` (same component-discovery as `plugin install`):
- ✅ **MCP**: Copilot calls `ctx_execute` and it resolves.
- ✅ **Hooks fire**: the `PostToolUse` hook captured a **non-MCP `bash` tool's** output into the session DB — proof the hook ran, since the MCP server never sees bash tools.
- ✅ **Detection**: with the env pin, `ctx_upgrade` resolves `copilot-cli` and writes the Copilot hook, leaving a co-installed Claude Code's config untouched.
- ✅ The manual path (`copilot mcp add` + `context-mode upgrade`) also verified: MCP responds and the standalone hooks fire independently of any plugin.

New guard tests: `tests/adapters/copilot-cli-plugin.test.ts` (5 passing).

ousamabenyounes:
Thanks for the careful re-submission @ken-jo — the bundle regeneration is in sync this time (`assert-bundle` OK), the fail-open `exit(0)` reasoning is solid, and the `COPILOT_HOME` mirroring between `copilotCliHome()` and `detect.ts` is a nice touch. I ran the full suite locally on this branch: **4255 passed / 43 skipped, exit 0**, build clean. 🎉

Sharing three medium-severity items I hit during review, each with a suggested patch — none are blockers, but #1 is worth fixing before merge. 🙏

---

### 🟠 1. `hooks/copilot-cli/pretooluse.mjs` — unguarded body can brick the agent

The hook body isn't wrapped in `try/catch` (the top-level `await initSecurity` and lines 14–30). Any throw → non-zero exit + empty stdout. By your own note in `cli.ts`, **Copilot CLI 1.0.59 treats that as "hook errored" and denies *every* tool** ("Denied by preToolUse hook") — which is exactly the brick mode `hookDispatch()` already guards against. So the fail-open you added there is undone if the hook itself throws.

It matches the existing convention (codex/kimi/vscode are unguarded too), but copilot-cli is the one platform documented to brick on error, so it deserves a fail-open wrapper:

```js
const __hookDir = dirname(fileURLToPath(import.meta.url));

try {
  await initSecurity(resolve(__hookDir, "..", "..", "build"));

  const raw = await readStdin();
  const input = parseStdin(raw);
  const tool = input.tool_name ?? input.toolName ?? "";
  const toolInput = input.tool_input ?? input.toolArgs ?? {};
  const projectDir = getInputProjectDir(input, COPILOT_OPTS);

  const decision = routePreToolUse(
    tool,
    toolInput,
    projectDir,
    "copilot-cli",
    getSessionId(input, COPILOT_OPTS),
  );
  const response = formatDecision("copilot-cli", decision);
  if (response !== null) {
    process.stdout.write(JSON.stringify(response) + "\n");
  }
} catch {
  // Fail OPEN — emit nothing so Copilot CLI ALLOWS the tool. An exit-non-zero +
  // empty-stdout PreToolUse hook is treated as "hook errored" by Copilot CLI
  // 1.0.59 and DENIES every tool — the same brick mode hookDispatch() guards.
}
```

---

### 🟠 2. `src/util/jsonc.ts` — trailing-comma strip corrupts string values

`stripJsonComments` is string-aware for *comments*, but the trailing-comma removal `replace(/,(\s*[}\]])/g, "$1")` (line 33) runs over the whole output, **including string contents**. Reproduced on this branch:

```js
import { parseJsonc } from "./build/util/jsonc.js";
const jsonc = `{
  // comment forces the strip path
  "note": "array literal: [1, ]"
}`;
parseJsonc(jsonc).note;
//  actual:   "array literal: [1 ]"   ❌ comma eaten
//  expected: "array literal: [1, ]"
```

It only triggers on the strip path (a comment-bearing config — exactly the "configs we did not write" case `parseJsonc` exists for) **and** a string containing `, ]` or `, }` (command strings, regexes, prompts). Silent config corruption.

Suggested fix — make the trailing-comma removal string/comment-aware inside the scanner and drop the post-hoc regex:

```js
export function stripJsonComments(str: string): string {
  let out = "";
  let inString = false;
  let escaped = false;
  let inBlockComment = false;
  for (let i = 0; i < str.length; i++) {
    const c = str[i];
    const next = str[i + 1];
    if (inBlockComment) {
      if (c === "*" && next === "/") { inBlockComment = false; i++; }
      continue;
    }
    if (escaped) { out += c; escaped = false; continue; }
    if (c === "\\") { out += c; escaped = inString; continue; }
    if (c === '"') { inString = !inString; out += c; continue; }
    if (!inString) {
      if (c === "/" && next === "/") {
        while (i < str.length && str[i] !== "\n") i++;
        if (i < str.length) out += "\n";
        continue;
      }
      if (c === "/" && next === "*") { inBlockComment = true; i++; continue; }
      // String-aware trailing-comma removal: a comma is "trailing" only when the
      // next significant char (skipping whitespace + comments) is } or ].
      if (c === ",") {
        let j = i + 1;
        while (j < str.length) {
          const d = str[j];
          if (d === " " || d === "\t" || d === "\r" || d === "\n") { j++; continue; }
          if (d === "/" && str[j + 1] === "/") { while (j < str.length && str[j] !== "\n") j++; continue; }
          if (d === "/" && str[j + 1] === "*") { j += 2; while (j < str.length && !(str[j] === "*" && str[j + 1] === "/")) j++; j += 2; continue; }
          break;
        }
        if (str[j] === "}" || str[j] === "]") continue; // drop the trailing comma
        out += c;
        continue;
      }
    }
    out += c;
  }
  return out; // no more post-hoc regex
}
```

Worth a unit test asserting `parseJsonc` preserves `, ]` / `, }` inside a string on the JSONC path.

---

### 🟠 3. Debug instrumentation shipped to prod — `posttooluse.mjs` + `precompact.mjs`

`logDebug()` writes `~/.copilot/context-mode/*-debug.log` via `appendFileSync` on **every** invocation (`CALL:`, `OK:`, `ERR:`). That's unbounded file growth in the user's config dir on every PostToolUse. It's also inconsistent — `sessionstart`/`stop`/`userpromptsubmit` have no such logging — so it reads like leftover debug.

Simplest fix: gate it behind an env so it's opt-in (or drop it entirely):

```js
const DEBUG = process.env.CONTEXT_MODE_DEBUG === "1";
function logDebug(line) {
  if (!DEBUG) return;
  try {
    mkdirSync(dirname(DEBUG_LOG), { recursive: true });
    appendFileSync(DEBUG_LOG, line);
  } catch {
    /* silent */
  }
}
```

---

Nice work overall — the marker that gates copilot-cli detection on a context-mode-written file (so it won't hijack a co-installed Claude Code user) is exactly right. Happy to re-test once the above are in. 👍

mksglu:
@ken-jo , can we follow `CONTRIBUTING.md` and always update the existing files instead of creating new ones?Creating new files makes maintenance harder, so I’d prefer us to modify the current files whenever possible. cc: @ousamabenyounes

ken-jo:
Thanks @ousamabenyounes and @mksglu — all three review items + the CONTRIBUTING guidance are addressed. Pushed as 5 focused commits; full suite green (185 files, 4265 passed / 44 skipped, 0 failed) and the bundles are regenerated in lockstep.

**Reviewer items**
- 🟠 #1 PreToolUse brick — wrapped `hooks/copilot-cli/pretooluse.mjs` in a fail-open `try/catch`. A legitimate veto is a normal stdout write + return (never a throw), so only real errors are swallowed → empty stdout + exit 0 = host ALLOWS, no brick. Added a regression test that spawns the hook with a throwing payload and asserts exit 0 + no deny.
- 🟠 #2 jsonc trailing-comma — made the trailing-comma strip string-aware (a second pass over the comment-free output) so commas inside string values are preserved while real trailing commas — including ones separated from `}`/`]` by a comment — are still stripped. `parseJsonc` still tries strict `JSON.parse` first. Unit test added. The identical duplicates in `src/server.ts` and `src/adapters/opencode/index.ts` are deferred to a follow-up consolidation PR (they parse third-party configs — wider blast radius).
- 🟠 #3 debug logs — gated the per-invocation debug writes behind `CONTEXT_MODE_DEBUG` (posttooluse + precompact + sessionstart), matching the kimi-hook precedent; the stdout/exit contract is unchanged.

**CONTRIBUTING (@mksglu)** — consolidated to update existing files instead of adding new ones:
- Merged the 3 standalone test files into their canonical homes and deleted them: `copilot-cli-plugin` → `tests/adapters/copilot-cli.test.ts`, `antigravity-cli-plugin` → `tests/adapters/antigravity.test.ts`, `strict-client-schema` → `tests/core/server.test.ts`. Net-new test files at PR scope are now just 2 — `tests/adapters/copilot-cli.test.ts` (new adapter) and `tests/hooks/copilot-cli-hooks.test.ts` (new-platform hooks) — both required by the per-platform convention; `antigravity-cli` reuses the existing antigravity family file.
- Renamed `copilot-cli-capture.test.ts` → `copilot-cli-hooks.test.ts` to match the `<platform>-hooks.test.ts` convention (cursor/gemini/vscode/jetbrains/kiro/kimi) and added the row to the CONTRIBUTING test-file table.
- Caught + fixed a drift bug while auditing: the Copilot bundle manifest (`configs/copilot-cli/.github/plugin/plugin.json`) had a pinned `version` but wasn't in `version-sync` (the agy bundle was) — it would have frozen on the next bump. Registered it in `version-sync.mjs`, the `package.json` version git-add list, and the version-sync test.

Follow-ups (out of scope here, will send separately): collapse the 3 `stripJsonComments` copies into the shared util, and optional dispatcher-level (`src/cli.ts`) fail-open so every platform's PreToolUse hook shares the brick guard. Happy to re-test.

configs/antigravity-cli/hooks/hooks.json
```diff
@@ -0,0 +1,15 @@
+{
+  "hooks": {
+    "PostToolUse": [
```
mksglu, line 14: We need at least `preToolUse` hook in here. @ken-jo

mksglu (changes requested):

ken-jo:
@mksglu — addressed the `preToolUse` request, and it turned into a real upgrade: **agy now has bounded PreToolUse _enforcement_, not just capture.** (This corrects my earlier "capture-only" framing.)

**What I verified on real agy 1.0.6**
- agy honors its **native top-level decision contract** `{"decision":"deny"|"ask","reason"}` for PreToolUse — *not* Claude's `permissionDecision` / `hookSpecificOutput`, and it ignores PreToolUse `additionalContext`. So the routing decisions now actually block: raw `curl`/`wget`/inline-HTTP/build-tools are denied and redirected to the `context-mode/ctx_*` tools, with the per-tool guidance surfaced in the deny reason.
- Bundle migrated to the **agy-native root format**: ships `plugin.json` + `mcp_config.json` + `hooks.json` + `rules/context-mode.md` at the bundle root (agy reads the bundle-root files — confirmed via `agy plugin validate`/`install`). The old `.mcp.json` + `.claude-plugin/plugin.json` are removed.
- **Dual `hooks.json` is intentional** (agy quirk): the runtime fires hooks from the **root** `hooks.json`, but `agy plugin validate`/`install` only *reports* hooks when the **subdir** `hooks/hooks.json` also exists — so the bundle keeps both, byte-identical.
- Tool-name surface fixed to agy's `context-mode/<tool>`; routing aliases added (`run_command`→Bash, `view_file`→Read, `grep_search`→Grep, `read_url_content`→WebFetch, …) with `CommandLine`/`AbsolutePath`/`URL` arg extractors.
- `Stop` hook is registered **best-effort** — agy did not fire `Stop` in the `-p` probe (interactive unconfirmed), so it's excluded from the doctor health gate.
- **agy exec-timeout hardening:** agy does not enforce an MCP RPC timeout, so a runaway `ctx_execute` hung until interrupted. `ctx_execute`/`ctx_execute_file`/`ctx_batch_execute` now apply a default timeout (120s, tunable via `CONTEXT_MODE_AGY_EXEC_TIMEOUT_MS`) **only under agy**; every other host keeps the unbounded behavior (Issue #406).
- `version-sync` tracks the new bundle `plugin.json`; per-platform tests added (`tests/hooks/antigravity-cli-hooks.test.ts`, formatter/routing/tool-naming/version-sync coverage). Full suite green.

**Follow-up (separate PR, out of scope here):** the agy tool-name map is currently mirrored in three places (`payload.mjs` / `routing.mjs` / `extract.ts`); I left cross-reference comments and will collapse them into one shared table in a follow-up.

Happy to re-test on your side.

mksglu:
Hi @ken-jo Resolve conflicts

ken-jo:
I re-tested PR #787 on macOS after the Mac issue report.

Findings and fix:
- Reproduced the Copilot-side Mac failure with GitHub Copilot CLI 1.0.44: the plugin `.mcp.json` was rejected before MCP startup because the stdio entry had `command` but no explicit `args` array.
- Pushed `94ac918`, which adds `"args": []` to the Copilot plugin MCP config and locks it with a regression assertion.
- After the patch, the same Copilot 1.0.44 local plugin smoke reaches `MCP client for context-mode connected`. Full Copilot prompt completion is still blocked on this Mac by `Failed to list models`, and I confirmed that baseline failure happens even without the context-mode plugin, so it is separate from this PR.
- agy was also checked on macOS: `agy plugin validate` passes, `npm run install:agy` reports hooks ACTIVE, and a real `agy -p` MCP smoke returned `12`.

Local targeted tests passed, and the refreshed PR checks are green across Ubuntu/macOS/Windows.

ken-jo:
Clean-install follow-up is done.

I removed the existing global install and host registration first:
- `npm uninstall -g context-mode`
- `agy plugin uninstall context-mode`
- confirmed `context-mode` was no longer on PATH, global npm had no `context-mode`, agy had no imported plugin, and Copilot had no context-mode MCP registration.

That surfaced one more real packaging issue: the npm tarball shipped `npm run install:agy` but did not include `scripts/install-antigravity-cli-plugin.mjs`, so installed-package `npm --prefix ... run install:agy` failed with `MODULE_NOT_FOUND`. I pushed `b6ce026` to include the installer in `package.json#files` and added a regression assertion.

Clean install verification after the fix:
- `npm pack` contains both `scripts/install-antigravity-cli-plugin.mjs` and `configs/copilot-cli/.mcp.json`.
- `npm install -g context-mode-1.0.162.tgz` succeeds from the PR tarball.
- installed package contains Copilot `.mcp.json` with `args: []`.
- installed package `npm run install:agy` succeeds and reports hooks ACTIVE.
- real `agy -p` MCP smoke returned `12`.
- Copilot CLI 1.0.44 loads the installed package plugin bundle and reaches `MCP client for context-mode connected`.

Remote checks are green again on the new head: Ubuntu/macOS/Windows test + Ubuntu/macOS OpenClaw E2E.

Update after local network change: Copilot prompt completion is now verified too. Re-running the installed-package plugin smoke returned `72` for a `9 * 8` request, the Copilot log shows `MCP client for context-mode connected`, and the isolated context-mode data directory recorded one `ctx_execute` tool call. That makes the earlier `Failed to list models` result a local Copilot/model-list network issue after MCP startup, not a PR/package failure.

ken-jo:
## Maintainer-style validation update

@mksglu I validated this against the maintainer review requirements and updated the branch accordingly.

## Why this PR is needed

This PR resolves #774 and #775.

The visible product gaps were:

- Antigravity CLI / `agy` did not have a reliable first-class Context Mode integration path.
- GitHub Copilot CLI needed official MCP + hook support documentation and runtime wiring.
- The previous attempt, #779, did not work end-to-end: `agy` could be misdetected as Claude Code and the agy MCP tools did not reliably reach the model.

This PR is the narrower replacement for that earlier attempt. It focuses on the two CLI agents and the cross-cutting runtime fixes they actually need.

## Git archaeology / prior invariants checked

I checked the related history before treating this as mergeable:

- #774 / #775 define the feature need for Antigravity CLI and GitHub Copilot CLI.
- #779 is the predecessor PR and explains why the replacement needed to be narrower and validated end-to-end.
- `778c3dd` established that GitHub Copilot CLI dispatches hook config keys in camelCase only: `preToolUse`, `postToolUse`, `sessionStart`, `userPromptSubmitted`, `agentStop`, `preCompact`. PascalCase hook keys are silently ignored by the CLI.
- `e4f4379` added the Antigravity CLI timeout behavior via `resolveExecTimeout(timeout)`.
- #765 / `3cd82b0` added the upstream `cwd` passthrough invariant for `ctx_execute` and `ctx_batch_execute`.

The source conflict in `src/server.ts` had to preserve both sides: this PR's Antigravity timeout behavior and upstream #765's `cwd` passthrough behavior. The resolved code keeps both.

## Problems fixed during the dirty update

I updated the branch onto latest `next`, resolved the `src/server.ts` conflict, and rebuilt generated bundles from source.

The visible conflict risk was that resolving `src/server.ts` one-sidedly could either:

- keep Antigravity's timeout normalization but drop upstream `cwd` passthrough, or
- keep upstream `cwd` passthrough but drop Antigravity's timeout normalization.

The final resolution preserves both invariants in the same executor call.

I also fixed docs drift discovered during validation. Runtime config in `configs/copilot-cli/hooks.json` uses camelCase hook keys, but README/platform docs still described the Copilot CLI hook shape as PascalCase / VS Code-family style. That stale wording mattered because `778c3dd` showed Copilot CLI silently ignores PascalCase hook keys. The docs now match the actual runtime config.

The same fresh-checkout test-harness bug applied here too:

```sh
npx tsc --silent
```

TypeScript rejects that option, so I changed it to:

```sh
npx tsc --pretty false
```

This keeps the test's non-interactive compile purpose while making fresh worktree validation pass.

## Validation

Fresh branch head: `eb41f45`

| --- | --- | --- |
| Linux | `pnpm run build` | passed |
| Linux | bundle assertions / asymmetric drift checks | passed |
| Linux | Copilot CLI hooks/adapters, Antigravity hooks/adapters, detection/config-dir targeted tests | passed |
| Linux | targeted server coverage for `IDEA_INITIAL_DIRECTORY`, `resolveExecTimeout`, and strict-client schema sanitization | passed |
| macOS fresh worktree | hooks/adapters suite | `194 passed` |
| macOS fresh worktree | targeted server tests: `IDEA_INITIAL_DIRECTORY`, `resolveExecTimeout`, `sanitizeSchemaForStrictClients` | passed |
| Windows fresh worktree | hooks/adapters suite | `194 passed` |
| Windows fresh worktree | targeted server tests: `IDEA_INITIAL_DIRECTORY`, `resolveExecTimeout`, `sanitizeSchemaForStrictClients` | passed |
| GitHub Actions | CI `test` on ubuntu, macOS, Windows | passed |
| GitHub Actions | OpenClaw E2E on ubuntu/macOS | passed |

## Conclusion

I consider this merge-ready from the branch/validation side.

The update preserves the Antigravity timeout behavior, preserves upstream #765 `cwd` passthrough, and fixes the Copilot CLI docs/runtime mismatch so users are not directed toward hook keys that the CLI silently ignores.

ken-jo:
## Additional latest CLI retest

@mksglu I ran an additional latest-version CLI check for the two new CLI agents in this PR.

## Versions tested

- GitHub Copilot CLI: `1.0.62`
- Antigravity CLI / `agy`: `1.0.8`
- Branch under test: `eb41f45`

For the live MCP checks, I put a local `context-mode` wrapper first on `PATH` so the CLIs invoked this PR's `cli.bundle.mjs`, not any globally installed context-mode package.

## Results

| Environment | CLI | Check | Result |
| --- | --- | --- | --- |
| Linux | Copilot CLI 1.0.62 | `--plugin-dir configs/copilot-cli`, live `ctx_execute` call through this PR's MCP server | passed, returned `17` for `8 + 9` |
| Linux | agy 1.0.8 | `npm run install:agy`, then live `ctx_execute` call through this PR's MCP server | passed, returned `13` for `9 + 4` |
| macOS | Copilot CLI 1.0.62 | CLI version/update check | installed, but live prompt was not counted as validation because the macOS GitHub/Copilot auth is invalid (`gh auth status` reports the token is invalid; Copilot fails model listing) |
| macOS | agy 1.0.8 | `npm run install:agy` | plugin install succeeded: MCP server + routing rule + routing skill + hooks registered, stale schema cache cleared |
| macOS | agy 1.0.8 | live prompt / `ctx_execute` | not counted as validation because the CLI requires sign-in on that host |
| Windows | Copilot CLI 1.0.62 | `--plugin-dir configs/copilot-cli`, live `ctx_execute` call through this PR's MCP server | passed, returned `16` for `11 + 5` |
| Windows | agy 1.0.8 | `npm run install:agy` | plugin install succeeded: MCP server + routing rule + routing skill + hooks registered, stale schema cache cleared |
| Windows | agy 1.0.8 | live prompt / `ctx_execute` | not counted as validation: `agy -p` over SSH produced no stdout even for a simple `Answer only: ok` smoke prompt; direct `cmd.exe` execution hung until stopped. The log showed auth recovery and hook execution, so I am treating this as an agy 1.0.8/SSH print-mode validation blocker rather than a passing PR signal. |

## Interpretation

The latest Copilot CLI path still looks good on Linux and Windows with the local plugin bundle and this PR's MCP server.

The latest agy path is confirmed on Linux, and the installer path works on macOS and Windows. I am not claiming macOS/Windows live agy prompt success because those hosts did not produce reliable model-output evidence.

mksglu:
Hi @ken-jo Please resolve conflicts

mksglu (changes requested):
conflicts

ken-jo:
@mksglu done

ken-jo:
@mksglu I revalidated the latest #787 update against the maintainer review requirements.

Why this update is needed:

- Git history shows Copilot CLI support was introduced in this PR and then expanded from a manual `COPILOT_HOME` style setup into a real Copilot CLI plugin bundle.
- That changes the runtime invariant: when Copilot loads context-mode through the plugin bundle, MCP and hook validation must be evaluated against the plugin-scoped bundle, not only against a user-managed Copilot home config.
- The latest update makes that distinction explicit with `CONTEXT_MODE_COPILOT_PLUGIN=1`, validates the Copilot plugin bundle hooks from `configs/copilot-cli/hooks.json`, checks all six Copilot hook entries, and avoids reporting the plugin/manual setup through the misleading configured-version path.
- I also verified detection ordering: an explicit `COPILOT_HOME` with a valid Copilot config still wins over passive Antigravity markers, so this does not route Copilot sessions into the agy adapter.

Validation performed:

- `git diff --check` passed.
- `npx vitest run tests/adapters/copilot-cli.test.ts tests/adapters/detect-config-dir.test.ts --reporter=dot` passed.
- The broader related adapter/hook validation suite passed.
- `npm run typecheck` passed.
- `npm run build` passed.
- `npm pack --pack-destination /tmp` succeeded.
- Generated bundles were refreshed and included: `cli.bundle.mjs`, `server.bundle.mjs`, and `hooks/security.bundle.mjs`.

Real Agent / CLI validation:

- Linux:
  - Copilot CLI `1.0.63`: plugin runtime doctor passed, manual config doctor passed, explicit `COPILOT_HOME` detection passed, and a real Copilot agent invocation successfully called a context-mode MCP tool.
  - Antigravity CLI `agy 1.0.9`: doctor passed, and a real agy agent invocation successfully called `ctx_execute`.

- Windows:
  - Windows NT `10.0.26200`, Node `24.13.0`, npm `11.6.2`.
  - Copilot CLI `1.0.63`: plugin runtime doctor passed, manual config doctor passed, explicit `COPILOT_HOME` detection passed, and a real Copilot agent invocation successfully called a context-mode MCP tool.
  - Antigravity CLI `agy 1.0.9`: doctor passed, and the verification store recorded the real `ctx_execute` MCP tool call from agy.

- macOS:
  - Darwin `25.2.0` arm64, Node `26.0.0`, npm `11.12.1`.
  - Copilot CLI `1.0.63`: direct plugin/manual doctor gates passed. With the system HTTPS proxy exported into the CLI environment, Copilot successfully loaded the model and the debug log confirmed a real `context-mode-ctx_execute` MCP `tool_use`.
  - Antigravity CLI `agy 1.0.9`: doctor passed, and the verification store recorded the real `ctx_execute` MCP tool call from agy.

Conclusion:

This update is meaningful and merge-worthy from my validation side. It preserves the existing Copilot CLI support while fixing the plugin-runtime validation path, and the affected Copilot/agy behavior was verified through real agent invocations across Linux, Windows, and macOS.

Issue #781, opened 2026-06-05
Plain-text chunker (#chunkPlainText) has no byte cap — large command output indexed as one chunk, bloats FTS5 content DB


brandonwie:
**Version:** context-mode@1.0.162

## Bug

`#chunkPlainText` (in `build/store.js` — the path `indexPlainText` uses for command/tool output) enforces no byte cap. Its only size check is the blank-line fast-path (`< 5000` bytes per section), which applies **only** when the split yields 3–200 sections. Otherwise:

- if the text has ≤ `linesPerChunk` (default 20) lines, it returns the **entire blob as one chunk**;
- else it groups by line count with **no byte check**, so a few very long lines still produce an oversized chunk.

`#chunkMarkdown` caps at `MAX_CHUNK_BYTES`, but `#chunkPlainText` does not.

## Impact

A broad `rg` / `grep` dump, large JSON, or few-newline log auto-indexed via `ctx_batch_execute` produced a ~104 MB single chunk, bloating a content DB to ~605 MB (with a similarly large `-wal` sidecar). Subsequent `ctx_search` / `ctx_doctor` MCP calls then timed out at 120 s.

## Proposed fixes

1. Enforce `MAX_CHUNK_BYTES` inside `#chunkPlainText` — on **both** the ≤20-line fast path and the line-group path (sub-split oversized slices by bytes).
2. Add an index-time hard cap at the store choke point (`index` / `indexPlainText` / `indexJSON`): reject or truncate payloads above a configurable threshold and return a diagnostic instead of silently indexing.
3. `ctx_doctor`: warn on large DB size / WAL size / max single-chunk size so bloat is caught before it causes query timeouts.

Happy to send a PR if a maintainer confirms the approach.

PR #789, opened 2026-06-06
fix(store): enforce MAX_CHUNK_BYTES in #chunkPlainText to prevent FTS5 DB bloat

ousamabenyounes:
## Problem

`#chunkPlainText` (in `src/store.ts`) did not uphold the `MAX_CHUNK_BYTES` (4096) cap that `#chunkMarkdown` enforces. Large low-newline output (broad grep dumps, big JSON, dense logs) was indexed as oversized chunks, bloating the FTS5 DB and timing out `ctx_search` / `ctx_doctor`.

The first revision fixed the obvious cases. Review (@ken-jo, DB-level checks across Linux/macOS/Windows) surfaced **two remaining holes** where chunks could still exceed the cap:

1. **Blank-line section fast path** accepted any section under the `5000B` strategy guard, so sections in the **4097–4999B band** were persisted uncapped.
2. **Long single-line fallback** sliced by JS character count, so multibyte text overshot — 4096 CJK chars = **12288B**, and emoji surrogate pairs could be cut mid-sequence.

## Changes

- `#splitOversizedPlainChunk` helper splits lines into byte-capped sub-chunks.
- **Blank-line section path** now sub-splits any section above `maxChunkBytes` (closes hole 1).
- New `#byteCappedPrefix` walks by Unicode **code point** so long-line splitting is **byte-accurate** and never cuts a multibyte char or surrogate pair, with guaranteed forward progress (closes hole 2).
- Line-group path and `indexPlainText` / `indexJSON` propagate `maxChunkBytes` through every path.
- Touched literals (`3`, `200`, `5000`, `80`, `0.5`) extracted into named constants.
- **Regression tests** (`tests/store-bytecap.test.ts`): huge single line, oversized line-groups, blank-line sections in the 4097–4999B band, long CJK line, long emoji line.

## Testing — real RED → GREEN

`npx vitest run tests/store-bytecap.test.ts` — indexes oversized plain text, reads the `chunks` table back, asserts every persisted chunk ≤ 4096 B.

**RED** (byte-cap guards reverted in `#chunkPlainText`) — reproduces the review counterexamples:
```
FAIL  blank-line sections in the 4097–4999B band respect the cap
  AssertionError: chunk of 4500B exceeds cap 4096
FAIL  a long multibyte (CJK) line is split by bytes, not characters
  AssertionError: expected the CJK line to be split, got 1 chunk(s)
FAIL  a long emoji line never splits a surrogate pair and stays capped
  AssertionError: expected the emoji line to be split, got 1 chunk(s)
Tests  3 failed | 2 passed (5)
```

**GREEN** (fix in place):
```
Test Files  1 passed (1)
      Tests  5 passed (5)
```

**No regression:** full store suite `npx vitest run tests/store.test.ts tests/store-bytecap.test.ts` → **128 passed**.

Bundles are gitignored and regenerated by the `bundle.yml` CI job, so this PR is `src/store.ts` + tests only.

Ref #781


mksglu:
Hi @ousamabenyounes Thx! Do we need this PR?

ousamabenyounes:
Yes — #chunkPlainText still has no byte cap while #chunkMarkdown caps at MAX_CHUNK_BYTES. The regression test shows the bug concretely: without the fix, a 12 KB line is stored as a single chunk. This PR fixes that specific inconsistency.
It only addresses point 1 from #781. The index-time cap and ctx_doctor warnings are complementary and can follow as separate PRs. I update the reference from Closes to Ref #781 accordingly.

ken-jo:
@ousamabenyounes Thanks for the PR.

I validated this following the maintainer review process, including git archaeology and direct DB-level behavior checks across Linux, macOS, and Windows.

## What this PR is trying to fix

This PR claims that plain-text chunks are capped by `MAX_CHUNK_BYTES`, so large plain text cannot bloat the FTS5 DB.

The patch is meaningful. It changes real indexing behavior in `src/store.ts` and adds regression coverage in `tests/store-bytecap.test.ts`.

## What passed

Linux:

- `npx vitest run tests/store-bytecap.test.ts tests/store.test.ts --reporter=dot`: passed, 125 passed

macOS:

- Same targeted store tests passed, 125 passed

Windows:

- Same targeted store tests passed, 124 passed / 1 skipped

## Direct DB-level counterexamples

I indexed plain text through `ContentStore.indexPlainText(...)`, then read persisted `chunks.content` byte sizes from SQLite.

This reproduced on Linux, macOS, and Windows:

```json
blank4500 {"count":3,"max":4500,"over":[4500,4500,4500]}
cjk {"count":2,"max":12288,"over":[12288,12288]}
```

## Root cause

Two paths can still exceed the byte cap:

- The blank-line section fast path accepts sections below 5000 bytes, which is above `MAX_CHUNK_BYTES=4096`.
- The long single-line fallback slices by JS character count, not by byte count, so multibyte text can exceed the byte cap.

## Interpretation

The patch improves some cases, but the core invariant is still false: not every persisted plain-text chunk is capped to `MAX_CHUNK_BYTES`.

## Requested changes

1. Make the blank-line section path respect `MAX_CHUNK_BYTES`.
2. Make long-line splitting byte-accurate rather than character-count-based.
3. Add regression tests for:
   - blank-line sections between 4097 and 4999 bytes
   - multibyte/CJK or emoji long lines
4. Re-run the store byte-cap tests after those cases are added.

ousamabenyounes:
@ken-jo Thank you for the thorough cross-platform review and the direct DB-level counterexamples — both reproduced exactly as you described. Pushed 678658a addressing all three points.

## 1. Blank-line section path now respects `MAX_CHUNK_BYTES`

The fast path accepted sections under the `5000B` strategy guard, so sections in the **4097–4999B band** were stored uncapped (your `blank4500 {"max":4500,"over":[4500,4500,4500]}`). Each section is now sub-split via `#splitOversizedPlainChunk` whenever it exceeds the cap:

```ts
return sections.flatMap((section, i) => {
  const trimmed = section.trim();
  if (trimmed.length === 0) return [];
  const title = trimmed.split("\n")[0].slice(0, CHUNK_TITLE_MAX_CHARS) || `Section ${i + 1}`;
  if (Buffer.byteLength(trimmed) <= maxChunkBytes) return [{ title, content: trimmed }];
  return this.#splitOversizedPlainChunk(trimmed.split("\n"), title, maxChunkBytes);
});
```

The `5000` strategy guard is kept (it only decides *which* strategy — blank-line vs line-grouping) but emission is now byte-safe on both branches.

## 2. Long-line splitting is byte-accurate, not character-count

The fallback sliced by JS character count (`remaining.slice(0, maxChunkBytes)`), so multibyte text overshot — 4096 CJK chars = 12288B (your `cjk {"max":12288}`). New `#byteCappedPrefix` walks by Unicode **code point**, never cutting a multibyte sequence or surrogate pair, with guaranteed forward progress:

```ts
#byteCappedPrefix(str: string, maxBytes: number): string {
  if (Buffer.byteLength(str) <= maxBytes) return str;
  let prefix = "", bytes = 0;
  for (const char of str) {
    const charBytes = Buffer.byteLength(char);
    if (bytes + charBytes > maxBytes) break;
    prefix += char; bytes += charBytes;
  }
  if (prefix.length === 0) return [...str][0] ?? "";
  return prefix;
}
```

## 3. Regression tests added

- blank-line sections in the **4097–4999B band**
- a long **CJK** line (byte-split, not char-split)
- a long **emoji** line (surrogate pair never cut)

I also extracted the touched literals (`3`, `200`, `5000`, `80`, `0.5`) into named constants.

## RED → GREEN (local)

**RED** — on the unfixed code (your exact counterexamples):
```
FAIL  blank-line sections in the 4097–4999B band respect the cap
  AssertionError: chunk of 4500B exceeds cap 4096
FAIL  a long multibyte (CJK) line is split by bytes, not characters
  AssertionError: expected the CJK line to be split, got 1 chunk(s)
FAIL  a long emoji line never splits a surrogate pair and stays capped
  AssertionError: expected the emoji line to be split, got 1 chunk(s)
Tests  3 failed | 2 passed (5)
```

**GREEN** — with the fix:
```
Test Files  1 passed (1)
      Tests  5 passed (5)
```

No regression — full store suite `npx vitest run tests/store.test.ts tests/store-bytecap.test.ts` → **128 passed**.

(Bundles are gitignored and regenerated by the `bundle.yml` CI job, so this commit is src + tests only, matching the original fix commit.)

The failing `CI / test (ubuntu-latest)` on the prior run was unrelated infra — `erlef/setup-beam` failed to download Rebar from the hex.pm mirror; the same SHA `2f03c70` also has a passing ubuntu run.

PR #796, opened 2026-06-08
Fix Windows shell runtime and PowerShell UTF-8 handling

Niche-alchemy:
## Summary
  - Prefer `pwsh` over Windows PowerShell 5.1 when Git Bash/POSIX shell is unavailable.
  - Prefix PowerShell shell execution with UTF-8 input/output encoding setup.
  - Update Codex/Claude guidance to recommend `ctx_execute(language: "javascript")` for portable filtering/counting instead of bash-only shell
  examples.
  - Add regression tests for Windows shell selection and PowerShell UTF-8 script prolog.

  ## Why
  On Windows, `ctx_execute(language: "shell")` may run under PowerShell instead of POSIX shell. Bash-style examples such as `grep/head/tail` can silently fail or produce no useful output. PowerShell 5.1 also defaults to non-UTF-8 encodings, which breaks Chinese paths/output.

  ## Verification
  - `npm run typecheck`
  - `vitest run tests/runtime.test.ts -t "Windows prefers pwsh over powershell when bash unavailable"`
  - `vitest run tests/executor.test.ts -t "buildPowerShellScriptContent prefixes UTF-8 encoding setup"`
  - `npm run bundle`
  - `npm run assert-bundle`


ken-jo:
@Niche-alchemy Thanks for the PR.

I validated this following the maintainer review process, including git archaeology, targeted runtime tests, and native Windows checks.

## What this PR is trying to fix

This PR aims to improve Windows shell runtime behavior and PowerShell UTF-8 handling:

- PowerShell / pwsh shell execution gets UTF-8 input/output setup.
- Windows shell detection chooses an available shell in a defined order.
- The routing/docs guidance moves away from host-shell-specific `language: "shell"` examples where a language-neutral `ctx_execute` example is safer.

## Validation performed

Linux:

- `npx vitest run tests/runtime.test.ts tests/executor.test.ts --reporter=dot`: passed, 143 passed / 28 skipped

macOS:

- Targeted PowerShell/runtime tests passed

Windows:

- `tests/executor.test.ts -t buildPowerShellScriptContent`: passed
- `tests/runtime.test.ts -t "Windows prefers pwsh"` failed on a native Windows machine where Git Bash is installed

Failure:

```text
expected 'C:\Program Files\Git\usr\bin\bash.exe' to be 'pwsh'
```

I also ran a direct PowerShell UTF-8 e2e through `PolyglotExecutor`. The process exited with code 0, but stdout was mojibake instead of the expected non-ASCII string:

```text
powershell {"exitCode":0,"stdout":"?.... mojibake output ....\r\n","stderr":""}
```

## Interpretation

The patch direction is useful, but the current Windows behavior does not prove the core UTF-8 claim yet. The new runtime test also fails in a realistic Windows environment where Git Bash is installed.

## Requested changes

1. Clarify and fix the Windows shell-selection expectation when Git Bash is installed.
2. Add or update a native Windows e2e test proving that PowerShell/pwsh returns non-ASCII stdout correctly through `PolyglotExecutor`.
3. Re-check whether the docs/routing example in `src/server.ts` should use `execSync`, because it throws on failing commands before the filtering logic can run.

Once those are updated, this should be re-tested on Windows specifically.

Issue #344, opened 2026-04-25
Windows: spawnSync npm ENOENT/EINVAL during upgrade (Node v20+)


thej4ck:
## Bug

On Windows with Node v20+ (tested v24.15.0), `context-mode upgrade` fails:

- First with `spawnSync npm ENOENT` — Node can't find `npm` (only `npm.cmd` exists on Windows official Node distribution)
- After patching to `npm.cmd`, fails with `spawnSync npm.cmd EINVAL` — Node v20+ CVE-2024-27980 mitigation requires `shell: true` for `.cmd`/`.bat` files

## Repro

1. Windows + Node v20+ from nodejs.org official installer
2. Run `/context-mode:ctx-upgrade` (or `node cli.bundle.mjs upgrade`)
3. Observe: `GitHub pull failed — spawnSync npm ENOENT`

Doctor falls back gracefully but version stays at old release.

## Root cause

`execFileSync("npm", [...], opts)` on Windows requires:

1. Resolving `npm` → `npm.cmd` (Node won't auto-resolve via `PATHEXT` in `spawn`/`execFile`)
2. `shell: true` in options (Node v20+ security mitigation, see https://nodejs.org/en/blog/vulnerability/april-2024-security-releases-2)

## Suggested fix

```js
const NPM = process.platform === "win32" ? "npm.cmd" : "npm";
const SHELL_OPTS = process.platform === "win32" ? { shell: true } : {};
execFileSync(NPM, args, { ...opts, ...SHELL_OPTS });
```

Affected callsites in `cli.bundle.mjs` (v1.0.89 → v1.0.93): 8 occurrences across direct `execFileSync` calls AND template-string scripts (the embedded clone+install+build script).

## Workaround (local patch — confirmed working)

1. Replace all `"npm",[` → `(process.platform==="win32"?"npm.cmd":"npm"),[`
2. Wrap `execFileSync` import:
   ```js
   import { execFileSync as __orig } from "node:child_process";
   var Vn = (c, a, o) => __orig(c, a, process.platform === "win32" ? { ...o, shell: true } : o);
   ```
3. Patch template-string options to add `shell: process.platform === "win32"`

After patches: upgrade `1.0.89` → `1.0.93` succeeded cleanly. Doctor: all PASS.

## Note on `shell: true` + DEP0190

[...]

PR #802, opened 2026-06-09
fix: Windows spawnSync npm ENOENT in postinstall and start.mjs (#801)

### LINKED ISSUE #801: [Bug]: fails with `spawnSync npm ENOENT`
### Platform

Claude Code

### context-mode version

context-mode: v1.0.89 (target: v1.0.162)

```json
not found script
```

```text
/context-mode:ctx-upgrade
```

```text
not found script
```

## Reproduction
1. Install context-mode v1.0.89 via Claude Code marketplace on Windows.
2. Run `/context-mode:ctx-upgrade`.

## Root cause (likely)
On Windows, Node's `child_process.spawnSync('npm', args)` without `shell: true`
cannot resolve `npm` because the real executable is `npm.cmd` / `npm.exe`. The
bare `npm` shell script (from Git Bash's PATH) is not a valid Windows
executable for `CreateProcess`. This is the standard Windows + `spawnSync` gotcha.

## Suggested fix
In the upgrader's spawn call, either:
- pass `{ shell: true }` to `spawnSync` (simplest, cross-platform), or
- on Windows, resolve to `npm.cmd` explicitly (e.g., `process.platform === 'win32' ? 'npm.cmd' : 'npm'`).

Same fix likely applies to any other `spawnSync`/`spawn` calls targeting
`npm`, `git`, `npx`, etc. in the upgrader path.

## Workaround
None from the user side — `npm` is already on PATH; the call site needs the fix.

### Pre-submission checklist

### Operating System

Windows (PowerShell)

### JS Runtime

Node: v24.15.0

simongonzalezdc:
## Problem

On Windows, Node's `child_process.execSync/spawnSync('npm', args)` without `shell: true` cannot resolve `npm` because the real executable is `npm.cmd` / `npm.exe`. The bare `npm` shell script (from Git Bash's PATH) is not a valid Windows executable for `CreateProcess`.

This causes `spawnSync npm ENOENT` during `/context-mode:ctx-upgrade` on Windows (issue #801):
1. `ctx-upgrade` runs `npm install -g context-mode`
2. The postinstall script calls `execSync("npm config get prefix")`
3. This fails because `npm` isn't resolvable on Windows without `.cmd` extension or `shell: true`

### `scripts/postinstall.mjs`
- Use `npm.cmd` on win32 (matching existing `npmExecFile` / `npmExec` helpers in `cli.ts`)
- Pass `shell: true` on Windows to the `execSync` call

### `start.mjs` (dev fallback)
- Same fix for the dev-mode fallback block that runs `npm install` / `npx tsc` when bundles are missing
- Use `npm.cmd` / `npx.cmd` on win32 with `shell: true`

### Tests
- `tests/util/windows-npm-spawn.test.ts`: source-code assertions verifying both files contain the Windows-safe spawn pattern (matches existing static-analysis test style in the repo)

## Verification

```bash
npx vitest run tests/util/windows-npm-spawn.test.ts
# 6 passed

npx vitest run tests/util/postinstall-heal.test.ts tests/scripts/start-mjs-mcp-boot.test.ts
# 11 passed
```

Fixes #801


ken-jo:
Hi @simongonzalezdc, thanks a lot for the contribution and for digging into the Windows spawn issue! 🙏

I pulled the branch and verified it locally on **Windows 11 / Node 24** — tests pass (17/17), clean and mergeable. One thing I wanted to raise before merge though, just so we're on the same page:

#801's actual root cause looks like the bare `execFileSync("npm", [...])` calls in `cli.ts`'s upgrade path (no shell → `spawnSync npm ENOENT`, since `execFileSync` won't resolve `npm` → `npm.cmd` via PATHEXT). Those were already fixed in **#344 (`337ad7e`, shipped in v1.0.90+)** by routing through `npmExecFile()` (`npm.cmd` + `shell: true`). The reporter on the issue was on **v1.0.89**, i.e. just before that fix landed.

The two files this PR touches (`postinstall.mjs`, `start.mjs`) use `execSync`, which always goes through `cmd.exe` and resolves `npm.cmd` via PATHEXT (and ignores Git Bash's extensionless `npm` script) — so they shouldn't be vulnerable to this ENOENT in the first place. In my local run the *original* `execSync("npm config get prefix")` worked fine on Windows.

So I might well be missing something here, and I'd genuinely like to understand it better: did you manage to actually reproduce `spawnSync npm ENOENT` coming from these specific `execSync` call sites (postinstall / start.mjs) on Windows — or run a test where the original code actually failed and the patched version made it pass? If you have that repro, I'd love to see it. 🙂

simongonzalezdc:
@ken-jo You're right, thanks for the careful review. I didn't independently reproduce the ENOENT from the execSync calls in postinstall.mjs / start.mjs — I traced the issue from #801's description and
 assumed all npm spawn sites had the same vulnerability, but missed that execSync already goes through cmd.exe and handles PATHEXT resolution automatically. The actual root cause was the bare execFileSync
 calls in cli.ts, which was already fixed in #344 (337ad7e, v1.0.90+). The reporter was on v1.0.89.

 Given that:
 1. The reporter's version predates the real fix
 2. execSync already resolves npm.cmd on Windows via cmd.exe

 I think the right call is to close this PR — the issue is already fixed upstream. Thanks for catching this and explaining it clearly.

 Also — I have time and tokens to spare right now. If you have any other issues you'd like an extra pair of hands on, I'm happy to help.

mksglu:
@ken-jo You doing great job. Thank you. cc: @simongonzalezdc

PR #810, opened 2026-06-10
fix(upgrade): skip settings.json wiring only for true plugin installs

### LINKED ISSUE #198: Are PreToolUse/SessionStart hook entries in settings.json necessary?
## Observation

After running `ctx upgrade`, context-mode adds hook entries directly into `~/.claude/settings.json`:

```json
{
  "matcher": "Bash|WebFetch|Read|Grep|Agent|Task|...",
  "hooks": [
    {
      "type": "command",
      "command": "node \"/Users/.../.claude/plugins/marketplaces/context-mode/hooks/pretooluse.mjs\""
    }
  ]
}
```

I noticed that the plugin also ships with a `hooks.json` that uses `${CLAUDE_PLUGIN_ROOT}`:

```json
{
  "matcher": "Bash",
  "hooks": [
    {
      "type": "command",
      "command": "node ${CLAUDE_PLUGIN_ROOT}/hooks/pretooluse.mjs"
    }
  ]
}
```

I tried removing the `settings.json` entries and ran `ctx doctor` — all checks still passed. It seems like Claude Code's plugin system picks up hooks from `hooks.json` automatically.

Just curious — are the `settings.json` entries needed for certain scenarios, or is `hooks.json` sufficient on its own?

**Environment:**
- context-mode v1.0.54
- Claude Code CLI
- macOS

eejd:
context-mode upgrade was calling readPluginHooks() and, upon finding the bundled hooks/hooks.json in the npm install directory, deciding that settings.json registration was unnecessary ("plugin hooks.json is sufficient"). This skips hook wiring for ALL installs, including standalone MCP server installs (e.g. MacPorts) where context-mode is NOT in enabledPlugins.

The bundled hooks/hooks.json uses ${CLAUDE_PLUGIN_ROOT} in its commands, which is only set by Claude Code when context-mode is an enabledPlugin. In a standalone install that env var is never set, so those hooks would silently fail at runtime — but more critically, Claude Code never reads hooks.json from an npm package directory unless it's an enabled plugin. Only ~/.claude/settings.json is authoritative for standalone installs.

Fix: gate the plugin-hooks-sufficient path on checkPluginRegistration() returning "pass" (i.e. context-mode is actually in enabledPlugins). Standalone / MacPorts installs now always fall through to the settings.json absolute-path hook registration path.

## Problem

  `context-mode upgrade` silently skips writing hooks to `~/.claude/settings.json` for **standalone / MacPorts installs** — any
  install where context-mode is *not* in `enabledPlugins` in `settings.json`.

  ### Root cause

  `configureAllHooks()` calls `readPluginHooks()`, which finds the bundled `hooks/hooks.json` that ships inside every npm package
  directory. It interprets this as evidence that the Claude Code plugin system is handling hooks and returns early.

  The bundled `hooks/hooks.json` uses `${CLAUDE_PLUGIN_ROOT}` in its commands. That env var is only set by Claude Code when
  context-mode is an `enabledPlugin`. In standalone installs it is never set, so Claude Code never reads `hooks.json` from the npm
  package directory at all — only `~/.claude/settings.json` is authoritative. The early return leaves all hooks (PreCompact,
  SessionStart, etc.) absent from `settings.json`, silently broken after a successful-looking `upgrade`.

  ## Fix

  Gate the early-return on `checkPluginRegistration()` returning `"pass"` (context-mode actually present in `enabledPlugins`). This
  method already existed for diagnostics. Standalone installs fall through and always write absolute-path hook commands to
  `settings.json`.

  ## Testing

  - `context-mode upgrade` on a MacPorts install (no `enabledPlugins` entry) now correctly writes all required hooks to
  `settings.json`
  - `context-mode doctor` passes all hook checks after upgrade
  - Existing plugin installs unaffected — `checkPluginRegistration()` returns `"pass"` for them

  Fixes #198

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ x] All platforms

## Checklist

- [x ] `npm test` passes
- [ x] `npm run typecheck` passes

<details>

</details>


mksglu:
Hi @eejd CI has error. Did you test that on manually?

mksglu:
@eejd  Do NOT commit bundle files

ken-jo:
Hi @eejd, I’m a contributor helping with review/triage.

I reviewed this against the maintainer validation requirements: the PR description, the historical reason for the existing behavior, and the actual standalone-vs-plugin runtime behavior were checked together.

## What I checked

The PR claim is that standalone Claude Code installs were being treated like plugin installs just because packaged `hooks/hooks.json` exists. That would cause `context-mode upgrade` to skip writing hooks into `~/.claude/settings.json`, even when Context Mode is not actually enabled as a Claude Code plugin.

I ran git archaeology on the Claude Code adapter path:

- `git log --follow --all -- src/adapters/claude-code/index.ts`
- `git log -S 'configureAllHooks' -- src/adapters/claude-code/index.ts tests/adapters/claude-code.test.ts`
- `git log -S 'readPluginHooks' -- src/adapters/claude-code/index.ts tests/adapters/claude-code.test.ts`
- `git log -S 'checkPluginRegistration' -- src/adapters/claude-code/index.ts tests/adapters/claude-code.test.ts`

The important prior invariant is that real plugin installs should not duplicate hook registration into `settings.json`. This PR appears to preserve that invariant while narrowing the skip condition to true plugin registration.

## Validation performed

On a fresh checkout of this PR head, I ran:

- `npm run build` passed
- `npx vitest run tests/adapters/claude-code.test.ts --reporter=dot` passed: 66 tests

I also ran a direct runtime check with a temporary `CLAUDE_CONFIG_DIR`:

- standalone settings `{}` plus packaged `hooks/hooks.json` wrote `PreToolUse` and `SessionStart` hooks into `settings.json`
- actual plugin registration through `enabledPlugins.context-mode` still skipped `settings.json` hook wiring

That matches the intended behavior: standalone installs are repaired, while the previous duplicate-hook protection for true plugin installs is kept.

## Conclusion

I think the source/test direction is valid and has merge value.

The current blocker is merge readiness. GitHub still reports this PR as conflicted, and there are no current checks on the PR. Also, the maintainer has already requested not to commit bundle files here, so the update should keep the source/test fix while removing or resolving generated artifact changes according to that direction.

Could you please rebase/resolve the current conflicts, clean up the generated bundle/stat artifacts as requested by the maintainer, and let CI run again?

PR #812, opened 2026-06-11
fix: preserve ZodEffects preprocess and ZodType coerce in zod3tov4

LeoNardo-LB:
## What / Why / How

**What:** The `zod3ToV4` adapter had two bugs that broke the OpenCode in-process plugin validation path:

1. **All `ZodEffects` (including `z.preprocess`) were stripped.** Custom coercion transforms like `coerceJsonArray` (string → array) and `coerceBoolean` (string → boolean) were discarded. When the LLM tool-call bridge stringified array/boolean arguments, OpenCode's Zod v4 validation rejected them — `ctx_fetch_and_index`'s `requests` parameter was rejected with `Expected array, received string`.

2. **Native Zod coercion was lost.** `z.coerce.number()` (which is a `ZodNumber` with `_def.coerce=true`, NOT a `ZodEffects`) was converted to plain `z.number()`, so `concurrency: "2"` was rejected.

**Why:** `zod3tov4.ts` is used by `plugin.ts` to convert context-mode's Zod v3 schemas into Zod v4 equivalents for OpenCode/KiloCode's in-process plugin host. The conversion was too aggressive in stripping effects, and didn't account for the `coerce` flag on primitive Zod types.

**How:**
- `ZodEffects`: detect `effect.type === "preprocess"` and preserve the transform via `z.preprocess()` in v4. Refinement/transform effects still stripped (host schema only, as before).
- `ZodNumber`/`ZodString`/`ZodBoolean`: check `_def.coerce` and use native v4 `z.coerce.X()` equivalents.
- Never map `ZodEffects`→native coerce because `z.coerce.boolean()` in v4 uses `Boolean()` (which converts `"false"` → `true`, wrong behavior for custom transforms like `coerceBoolean`).

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] VS Code Copilot
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [x] OpenCode
- [x] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi / Kiro / Antigravity / Zed
- [ ] All platforms

### Unit tests (19/19 pass)
- New: `preserves ZodEffects preprocess — coerces string to array`
- New: `preserves ZodEffects preprocess — coerces string to boolean`
- New: `strips ZodEffects refinement to inner schema`
- Updated existing test to verify preprocess is preserved (not stripped)

### E2E tests — OpenCode plugin validation chain (9/9 pass)
Simulates `plugin.ts` → `zod3ShapeToV4` → `z4.object()` wrapping:
- String-ified `requests` array accepted
- Native array passes through
- `force: "true"` / `"false"` string coercion
- `concurrency: "2"` string coercion
- All fields together end-to-end

### Production tests — BUILT module (8/8 pass)
Imports the actual `server.bundle.mjs`, extracts the real `ctx_fetch_and_index` tool schema, converts it via built `zod3tov4.js`, and validates as OpenCode would:
- Actual tool schema with stringified args: ✓ PASSED
- Native array + `force:"false"`: ✓ PASSED

### Regression
- `tests/adapters/` — 40/42 files pass (6 pre-existing failures unrelated)
- `npm run typecheck` — passes
- `npm run build` — passes

## Checklist

- [x] `npm test` passes (40/42 — 6 pre-existing unrelated failures)
- [x] Docs — no docs changes needed (internal fix only)
- [x] No Windows path regressions (no path changes)
- [x] Targets `next` branch

### DIFF
--- a/src/adapters/opencode/zod3tov4.ts
+++ b/src/adapters/opencode/zod3tov4.ts
@@ -32,15 +32,15 @@ function zod3ToV4(v: unknown, depth = 0): z.ZodType {
 
   switch (def.typeName) {
     case "ZodString":
-      result = z.string();
+      result = def.coerce === true ? z.coerce.string() : z.string();
       break;
 
     case "ZodNumber":
-      result = z.number();
+      result = def.coerce === true ? z.coerce.number() : z.number();
       break;
 
     case "ZodBoolean":
-      result = z.boolean();
+      result = def.coerce === true ? z.coerce.boolean() : z.boolean();
       break;
 
     case "ZodAny":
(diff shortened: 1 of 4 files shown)


LeoNardo-LB:
暂时撤回，待进一步确认后重新提交。

PR #820, opened 2026-06-12
fix(heal): make cache-root containment guards symlink-aware (#795)

### LINKED ISSUE #795: SessionStart cache-heal hook skips itself when ~/.claude is a symlink (path.resolve vs realpath)
### Summary
The SessionStart cache-heal hook computes its cache root with `path.resolve(os.homedir(), '.claude', 'plugins', 'cache')`, which does **not** resolve symlinks. When `~/.claude` is a symlink to another volume, the plugin's registered `installPath` is the *physical* target (e.g. `/Volumes/SSD/claude-code/plugins/cache/...`), so the `installPath.startsWith(cacheRoot)` guard is `false` and the hook **skips context-mode entirely** — the very hook meant to repair a broken path after an auto-update never fires for these users.

### Repro
1. Relocate `~/.claude` to an external volume and symlink it back (common for large-SSD setups).
2. Trigger the SessionStart cache-heal path.
3. The guard fails the `startsWith` check; context-mode is skipped.

### Environment
- macOS, `~/.claude` → `/Volumes/<SSD>/.../claude-code` (symlink)

### Suggested fix
`fs.realpathSync()` (or `fs.promises.realpath`) **both** `installPath` and the computed `cacheRoot` before comparing, so symlinked layouts compare on their real paths. A test with a symlinked plugin cache would guard the regression.

nyxst4ck:
## Summary

Fixes #795 — when `~/.claude` is relocated to another volume and symlinked back, the registry stores **physical** install paths while the heal callers compute the cache root through the **symlink**. The purely lexical `resolve()+startsWith` containment guards in `scripts/heal-installed-plugins.mjs` then never match, so HEAL 3/4 and both mcpServers heals silently skip context-mode — the very heals meant to repair a broken install never fire for exactly these users.

## Fix

New `isInsideRoot(childPath, rootPath)` helper: containment holds when **either** the lexical (`resolve`) **or** the physical (`realpathSync`, with lexical fallback for non-existent paths) forms match. Used by the three guards that compare an install path against `pluginCacheRoot`:

- HEAL 3 (`healInstalledPlugins` entry loop)
- `healPluginJsonMcpServers` (#523 guard)
- `healMcpJsonArgs` (#531 guard)

A `..`-escape fails **both** forms, so the path-traversal protection keeps its teeth (covered by a dedicated regression test). This mirrors the realpath canonicalization that `cli.ts` (cacheRootCanon) and `heal-partial-install.mjs` (layered lexical+realpath guards) already do — `heal-installed-plugins.mjs` was the remaining lexical-only spot.

`sweepStaleMcpJson`'s guard is intentionally untouched: both of its sides derive from the same resolved cache root, so a symlinked config dir cannot make them diverge.

## Tests

4 new tests in `tests/util/heal-installed-plugins.test.ts` (symlinks created as junctions so they work on Windows without admin; tests bail gracefully on environments that cannot symlink):

- heals when `pluginCacheRoot` is the symlink form and the registry stores physical paths (the #795 layout)
- heals in the reverse direction (registry stores symlink paths, root is physical)
- still rejects `..` escapes when the cache root is symlinked
- `healPluginJsonMcpServers` heals through a symlinked cache root

```
bun run vitest run tests/util/heal-installed-plugins.test.ts   # 40 passed
bun run vitest run tests/util tests/scripts tests/hooks tests/cli  # 787 passed; 1 pre-existing failure (project-dir build artifact, fails on clean checkout too)
bun run typecheck                                              # clean
```

### DIFF
--- a/scripts/heal-installed-plugins.mjs
+++ b/scripts/heal-installed-plugins.mjs
@@ -17,8 +17,36 @@
  * @see https://github.com/anthropics/claude-code/issues/46915
  */
 
-import { existsSync, readFileSync, writeFileSync, readdirSync, unlinkSync, statSync } from "node:fs";
+import { existsSync, readFileSync, writeFileSync, readdirSync, unlinkSync, statSync, realpathSync } from "node:fs";
 import { resolve, sep } from "node:path";
+
+/**
+ * Best-effort realpath: dereference symlinks when the path exists, fall
+ * back to the lexical resolve when it doesn't (e.g. a stale installPath).
+ */
+function toRealPath(p) {
+  try {
+    return realpathSync(p);
+  } catch {
+    return resolve(p);
+  }
+}
+
+/**
+ * Containment check shared by the path-traversal guards below.
+ *
+ * `path.resolve` rejects `..` escapes but does not dereference symlinks.
+ * When the Claude config dir is relocated and symlinked back (#795 —
+ * `~/.claude` → another volume), the registry stores *physical* install
+ * paths while the computed cache root keeps the *symlink* path, the
+ * lexical forms never match, and every heal silently skips the plugin.
+ * Containment holds when either the lexical or the physical forms match;
+ * a `..` escape fails both, so the traversal guard keeps its teeth.
+ */
+function isInsideRoot(childPath, rootPath) {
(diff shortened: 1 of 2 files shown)


mksglu:
It's already shipped. Thanks.

PR #821, opened 2026-06-12
fix(pi): stop mutating systemPrompt, inject via context hook to preserve prefix cache

xuli500177:
## Problem

The Pi adapter uses `before_agent_start` to inject dynamic content (active_memory, resume snapshot, behavioralDirective usage stats) into the `systemPrompt`. Since `systemPrompt` sits at `messages[0]`, **any change invalidates the entire prefix prompt cache** on DeepSeek, Anthropic, and OpenAI APIs.

This causes cache hit rates to plummet from 90%+ to ~18%, as originally reported in #598 — but the fix in that issue only moved `behavioralDirective` within the systemPrompt, which does not solve the problem for API-level prefix caching (only for vLLM's auto-prefix algorithm).

## Root Cause

Unlike Claude Code (where hook output is appended to the message stream by the CC runtime), the Pi extension API's `before_agent_start` directly modifies the system prompt. Dynamic data at the prefix position = guaranteed cache miss every turn.

## Fix

Three minimal changes to `src/adapters/pi/extension.ts`:

1. **Add `_pendingContext` module variable** — buffer for context that should reach the LLM without touching systemPrompt
2. **In `before_agent_start`**: instead of `return { systemPrompt }`, store the extra context in `_pendingContext`
3. **Register a `context` hook** — same pattern used by hindsight/other extensions — appends `_pendingContext` as a user message at the **end** of the message array

This aligns the Pi adapter with how context-mode works on Claude Code and all other platforms — content injected at the end of messages, leaving the static prefix intact.

## Before / After

| | Before | After |
|---|---|---|
| Injection method | `systemPrompt` modification | `context` hook → message append |
| Position in messages | `messages[0]` (prefix) | End of `messages` (suffix) |
| Prefix cache impact | 💥 Full invalidation every turn | ✅ Fully preserved |
| Claude Code parity | ❌ Different path | ✅ Same pattern |

## Testing

- [x] Node `--check` syntax validation on both source (`.ts`) and built (`.js`) files
- [x] TypeScript typecheck on modified file: no new errors
- [x] Verified in live Pi session with DeepSeek: systemPrompt no longer changes between turns

### DIFF
--- a/src/adapters/pi/extension.ts
+++ b/src/adapters/pi/extension.ts
@@ -707,6 +707,7 @@ export default function piExtension(pi: any): void {
         _pendingContext = "";
       }
     } catch {
+      _pendingContext = ""; // Reset — ensure no stale data escapes
       // best effort — never break agent start
     }
   });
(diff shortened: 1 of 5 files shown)


mksglu:
Do NOT commit `web` directory and `stats.json` @xuli500177

Issue #807, opened 2026-06-10
Cleanup never re-creates breadcrumbs for orphaned version dirs still referenced by long-lived sessions


kouiso:
## Environment

- Claude Code 2.1.x
- context-mode 1.0.162 (auto-updated from 1.0.111)
- macOS (Apple Silicon) + WSL2 Ubuntu

## Summary

#181 added age-gated cleanup of old version dirs under `plugins/cache/context-mode/context-mode/`, and #644 fixed cleanup to *protect* fresh breadcrumb symlinks (the `lstatSync` fix in `sessionstart.mjs`). But nothing ever *creates* a breadcrumb for an orphaned version.

Sessions started before an auto-update keep the old version path baked into their in-memory hook configuration. Once cleanup removes that version dir, every hook invocation from those sessions fails with:

```
Plugin directory does not exist: ~/.claude/plugins/cache/context-mode/context-mode/1.0.111
```

The `cache-heal` hook does not cover this either: it only repairs `installPath` entries in `installed_plugins.json`, which already point at the new version after the update.

## Observed impact

After a 1.0.111 -> 1.0.162 auto-update on one machine, 9 long-lived sessions produced roughly 3,000 hook errors within 6 hours (transcript evidence available on request). The errors are non-destructive, but capture stops working for those sessions and the transcripts get flooded until every session is restarted. This re-occurs after **every** auto-update for any session that outlives it.

## Proposed fix

When cleanup removes a version dir, **leave a breadcrumb symlink to the current version in its place** (same mechanism #644 already protects):

```js
// after rmSync(oldVersionDir):
symlinkSync(currentVersion, oldVersionDir); // relative target, same parent dir
```

This needs zero extra state and structurally closes the gap: any session still referencing the old path transparently resolves to the current version, and the age-gate from #644 eventually retires the symlink once it goes stale.

## Workaround we deployed locally

A SessionStart hook that records every version ever seen in a `.version-history.json` file (one level above the version dirs, outside cleanup's scan range) and re-creates `old -> current` breadcrumb symlinks for any recorded version whose path has disappeared, pruning entries 30 days after the real dir goes away. Happy to share it if useful.

## Severity

Medium — non-destructive, but breaks capture and floods transcripts for every long-lived session after every auto-update.

Issue #644, opened 2026-05-20
sessionstart.mjs age-gated cleanup uses `statSync` (follows symlinks) — deletes fresh symlinks whose targets are stale; 4-char `lstatSync` fix


dtmsyi:
## Environment

- **OS**: macOS 15.x (Apple Silicon)
- **Node.js**: v22.x
- **Claude Code**: 2.x
- **context-mode**: v1.0.140 (installed via Claude Code plugin marketplace)

## Description

The age-gated cleanup loop added in the fix for #181 (`hooks/sessionstart.mjs` lines ~234–260) uses `statSync` to evaluate each entry in the cache parent directory. `statSync` follows symlinks, so if any entry is a symlink, the gate is evaluated against the **target's** mtime, not the symlink's own.

This causes the cleanup to incorrectly delete fresh symlinks whose targets happen to be old, even when those symlinks were just (re-)created moments ago — `force: true` in `rmSync` ignores the file type, so the symlink itself is removed.

The 4-character fix: `statSync` → `lstatSync` on line ~250.

```diff
-              const st = statSync(join(cacheParent, d));
+              const st = lstatSync(join(cacheParent, d));
               if (now - st.mtimeMs > ONE_HOUR) {
                 rmSync(join(cacheParent, d), { recursive: true, force: true });
               }
```

## Reproduction

Scenario observed in the wild after running a self-heal hook that re-creates orphaned `CLAUDE_PLUGIN_ROOT` entries from prior sessions as symlinks pointing at the current installed version:

```
~/.claude/plugins/cache/context-mode/context-mode/
├── 1.0.099  →  ./1.0.140   (symlink, mtime: just now)
├── 1.0.100  →  ./1.0.140   (symlink, mtime: just now)
├── 1.0.120  →  ./1.0.140   (symlink, mtime: just now)
├── 1.0.124  →  ./1.0.140   (symlink, mtime: just now)
└── 1.0.140/                (real dir, mtime: >1h ago)
```

When the next Claude Code session fires the SessionStart hook:

1. `readdirSync(cacheParent)` lists all 5 entries.
2. For each symlink, `statSync(...)` follows the link and returns `1.0.140`'s stat — mtime >1h ago.
3. Age gate trips. `rmSync(..., { recursive: true, force: true })` removes the symlink (not the target, because `recursive: true` doesn't follow the link to delete the target's contents — but `force: true` happily removes the link itself).
4. The other 3 symlinks pass the same test in the same iteration and are all removed.
[...]

PR #823, opened 2026-06-12
fix(cleanup): leave a breadcrumb link when sweeping old version dirs (#814, #807)

### LINKED ISSUE #814: sessionstart.mjs cleanup deletes old version dir without junction — breaks active sessions mid-session
## Bug

When the plugin auto-updates and an old session is still open, `sessionstart.mjs` deletes the old version directory after the 1-hour age gate. Any hooks still referencing the old version's absolute paths then fail with:

```
PreToolUse:Read hook error
Failed to run: Plugin directory does not exist:
D:\Users\...\plugins\cache\context-mode\context-mode\1.0.152
(context-mode@context-mode — run /plugin to reinstall)
```

## Reproduction timeline

1. Session A starts, loads `1.0.152/hooks/hooks.json` — absolute paths baked in: `.../1.0.152/hooks/pretooluse.mjs`
2. Auto-update runs: `installed_plugins.json` + `hooks.json` updated to `1.0.162`
3. ~1 hour later: `sessionstart.mjs` age-gate expires, deletes `1.0.152/`
4. Session A is still open — every subsequent hook call fails because `1.0.152/` no longer exists

## Why `context-mode-cache-heal.mjs` doesn't prevent it

The heal script reads `installed_plugins.json`, which was already updated to `1.0.162` at step 2. It finds the directory exists, calls `normalize-hooks.mjs`, and correctly skips junction creation. It has no visibility into sessions still holding stale paths from `1.0.152`.

## Root cause

`sessionstart.mjs` cleanup deletes old version dirs without leaving a forwarding junction. Any session that loaded hooks before the update is stranded.

## Proposed fix

Before deleting an old version directory in `sessionstart.mjs`, create a directory junction (Windows) / symlink (Unix) pointing `oldVersion → currentVersion`:

```js
// Before:
fs.rmSync(oldVersionDir, { recursive: true, force: true });

// After:
const junctionType = process.platform === 'win32' ? 'junction' : undefined;
try {
  fs.symlinkSync(currentVersionDir, oldVersionDir, junctionType);
} catch { /* junction already exists or no permission — skip */ }
// delay actual delete to next cleanup cycle once junction is safe to remove
```

This mirrors the existing pattern in `context-mode-cache-heal.mjs` (lines 42–46) and ensures sessions that loaded stale hook paths can still resolve their scripts without erroring.

## Environment

- Platform: Windows 11 Pro
- Plugin versions affected: `1.0.152` → `1.0.162`
- Claude Code: auto-updates channel `latest`

nyxst4ck:
## Summary

Fixes #814 and #807 — the age-gated cleanup in `sessionstart.mjs` (#181) deletes old plugin cache version dirs outright. Sessions that loaded hooks **before** an auto-update keep the old version's absolute paths baked into their hook configuration, so once the dir is gone every hook call from those sessions fails with `Plugin directory does not exist` until restart. #807 measured ~3,000 hook errors over 6 hours across 9 long-lived sessions, recurring after **every** auto-update. The cache-heal layer can't catch this: it repairs `installed_plugins.json`, which already points at the new version.

## Fix

After `rmSync`, create a symlink (junction on Windows) in the removed dir's place pointing at the live version root — the same breadcrumb pattern `healCacheMidSession` (server.ts) and `postinstall.mjs` already use, and the same mechanism #644's `lstatSync` age gate already protects:

- The fresh mtime shields the link for the next hour; each subsequent sweep refreshes it (delete + re-create), so breadcrumbs stay alive as long as cleanup keeps running.
- When the swept entry is itself a **stale breadcrumb** from an earlier update chain, it gets re-pointed at the current root instead of a since-deleted intermediate version (`1.0.050 → 1.0.150 → gone` becomes `1.0.050 → current`).
- Link creation is best-effort: on failure (no permissions), behaviour degrades to the previous plain delete.

**Trade-off to be aware of:** refreshed breadcrumbs never age out, so one zero-byte link per historical version accumulates in the cache parent. If you'd rather bound that (e.g. only breadcrumb real dirs, or cap link count), happy to adjust.

## Tests

New `tests/hooks/version-dir-breadcrumb.test.ts` runs the **real** `sessionstart.mjs` (spawnSync, fake `CLAUDE_PLUGIN_ROOT` cache layout, isolated HOME — same approach as `session-hooks-smoke.test.ts`):

1. aged-out real version dir → replaced by a working link (old `.../hooks/sessionstart.mjs` path still resolves)
2. dangling stale breadcrumb → re-pointed at the live root
3. fresh dir → untouched (#644 age gate intact)

Tests 1–2 are **red on main**, green with the fix.

```
bun run vitest run tests/hooks/version-dir-breadcrumb.test.ts   # 3 passed
bun run vitest run tests/hooks/ tests/util/                     # 733 passed; 1 pre-existing failure (project-dir build artifact, fails on clean checkout too)
bun run typecheck                                               # clean
```

Note: this is my second PR here today (#820 is the other) — both from studying the issue backlog; tell me if you'd rather I space them out.


tests/hooks/version-dir-breadcrumb.test.ts
mksglu, line 1: Do NOT create new file for that. Use existing one.

mksglu (changes requested):

nyxst4ck:
Done — moved the three tests into the existing `tests/hooks/cache-heal-self-heal.test.ts` as a new slice (alongside the other sessionstart.mjs version-dir self-heal coverage) and removed the standalone file. No test logic changed; the suite runs green (32 passed) and `typecheck` is clean. Thanks for the review!

mksglu:
Hi @nyxst4ck **CI has error.** Let’s be careful when changing this unit test and make sure we are not introducing a regression. AI can easily make the test pass, but we need to understand why it is failing and verify that the proposed approach is actually correct.

Changing the test is fine if the expected behavior has changed, but first let’s trace when and why this test case was introduced: the original PR, commit, related issue, and the behavior it was intended to protect. Once we understand that context, we can update the test with confidence.

ken-jo:
## Maintainer-style validation

@nyxst4ck I rechecked this against the maintainer review requirements. The root-cause direction looks valid, but I found a Windows blocker that needs an update before this is merge-ready.

## Why the patch is needed

The issue chain supports the fix direction:

- #814 reports that `sessionstart.mjs` cleanup deletes an old plugin cache version directory after the 1-hour age gate. Long-lived sessions that loaded hooks before an auto-update still have absolute hook paths pointing at that old version, so once the directory is gone those sessions fail with `Plugin directory does not exist`.
- #807 reports the same class of long-lived-session failure at larger scale, recurring after auto-updates.
- `013e6fa` / #181 originally moved aggressive cleanup into age-gated `sessionstart.mjs` cleanup specifically because deleting the old `CLAUDE_PLUGIN_ROOT` mid-session breaks active hooks.
- `1147411` / #644 changed the cleanup from `statSync` to `lstatSync` so fresh breadcrumb symlinks are judged by the link's own mtime, not the old target's mtime. That preserved the #181 invariant for existing breadcrumbs.

So the missing piece this PR is trying to add is real: when cleanup removes an old real version directory, it should leave a forwarding breadcrumb so old hook paths can still resolve to the live version. That is consistent with the earlier #181 and #644 invariants.

## What I validated

Branch under test: `524e415` (`fix/814-version-dir-breadcrumb`)

### Linux

- `pnpm run build`: passed
- `pnpm exec tsc --noEmit --pretty false`: passed
- `pnpm exec vitest run tests/hooks/cache-heal-self-heal.test.ts --reporter=dot`: passed, `33 passed`
- Broader `tests/hooks/ tests/util/` still has 3 `tests/util/postinstall-heal.test.ts` failures, but I reproduced the same 3 failures on clean `upstream/next`, so I do not attribute those to this PR.

### macOS

- Fresh checkout of PR #823
- `pnpm run build`: passed
- `pnpm exec vitest run tests/hooks/cache-heal-self-heal.test.ts --reporter=dot`: passed, `33 passed`

### Windows

- Fresh checkout of PR #823
- `corepack pnpm run build`: passed
- `corepack pnpm exec vitest run tests/hooks/cache-heal-self-heal.test.ts --reporter=dot`: failed

The failing Windows case is the new stale-breadcrumb case:

```text
Issues #814/#807 — cleanup leaves a breadcrumb to the live version
> a stale breadcrumb pointing at a removed intermediate version is re-pointed at the live root

Error: ENOENT: no such file or directory, stat
C:\Users\ikcha\AppData\Local\Temp\ctx-breadcrumb-...\plugins\cache\context-mode\context-mode\1.0.050

tests/hooks/cache-heal-self-heal.test.ts:785
expect(realpathSync(oldLink)).toBe(realpathSync(currentDir));
```

## Interpretation

The Linux/macOS behavior supports the patch direction, and the existing history supports the need for a breadcrumb instead of plain deletion.

However, Windows does not currently satisfy the new invariant for a dangling stale breadcrumb. The test creates a Windows junction for `1.0.050 -> 1.0.150`, removes/omits the intermediate target, and expects cleanup to re-point `1.0.050` at the current version. After running `sessionstart.mjs`, `realpathSync(oldLink)` still fails with `ENOENT`, so the breadcrumb was not successfully re-pointed on Windows.

That is important because #814 is explicitly a Windows failure mode. I would not merge this until the Windows dangling-junction case is fixed or the implementation/test is adjusted with concrete Windows evidence explaining a different intended behavior.

Could you update the PR so the stale breadcrumb/junction case passes on Windows too?

PR #813, opened 2026-06-11
fix(pi): lazy bootstrap MCP bridge for agent turns

Gandem:
## What / Why / How

Fixes https://github.com/mksglu/context-mode/issues/809 (the Pi package-command/help/version hang).

Previously, Pi extension loading could eagerly bootstrap the MCP bridge even for non-agent invocations such as package/help/version-style commands. That could spawn the long-lived MCP child in a context where no real agent session or reliable shutdown lifecycle follows, leaving bootstrap/orphan processes and causing command hangs.

This change makes bridge bootstrap lazy:
- extension registration no longer starts the MCP bridge
- `before_agent_start` starts and awaits bootstrap for real agent turns
- shutdown invalidates in-flight bootstrap generations so late handles are closed instead of orphaned
- `_mcpBridgeReady` is reset to a settled promise when no bridge should be running

I'm not sure if this qualifies for a hotfix since all `pi update`/`install` etc. commands are hanging without this change. Let me know what you think 🙏 

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] VS Code Copilot (GitHub Copilot)
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [X] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

Added/updated Pi adapter regression coverage.

- `tests/adapters/pi-help-skip-bootstrap.test.ts`
     - verifies extension discovery does not bootstrap for help/version/package-style argv
     - verifies real agent lifecycle starts bootstrap from `before_agent_start`
     - verifies bootstrap is not repeated more than once per extension registration

- `tests/adapters/pi-no-orphan-on-help.test.ts`
     - verifies non-agent argv does not construct/start an MCP child (verifies the fix from this specific issue)

- `tests/pi-extension.test.ts`
     - updates Pi MCP bridge wiring expectations for lazy bootstrap
     - verifies `before_agent_start` registers canonical `ctx_*` tools before model/tool registry snapshot
     - verifies shutdown handles in-flight lazy bootstrap without orphaning handles

## Checklist

- [X] Tests added/updated (TDD: red → green)
- [X] `npm test` passes
- [X] `npm run typecheck` passes
- [X] Docs updated if needed (README, platform-support.md)
[...]

Issue #534, opened 2026-05-12
v1.0.121: nested pi --help spawns orphaned MCP servers spinning at 100% CPU


gebeer:
# v1.0.121: nested `pi --help` spawns orphaned context-mode MCP servers that spin at 100% CPU

## Summary

On Linux/Pi with `context-mode@1.0.121`, running short-lived nested `pi` commands from inside an active Pi session can leave orphaned `context-mode/server.bundle.mjs` Bun processes spinning at 100% CPU. The processes ignored `SIGTERM` and required `SIGKILL`.

This appears related to #388/#311, but reproduced on v1.0.121 through Pi package auto-loading.

## Environment

- OS: Arch Linux
- Host: Pi Coding Agent
- context-mode: `1.0.121`
- Runtime: Bun via `/home/gbr/.bun/bin/bun`
- Pi config includes `npm:context-mode` in `~/.pi/agent/settings.json` packages.

Relevant settings:

```json
"packages": [
  "npm:pi-extmgr",
  "npm:pi-web-access",
  "npm:pi-mcp-adapter",
  {
    "source": "npm:pi-hosts",
    "extensions": ["-dist/index.js"]
  },
  "npm:context-mode"
]
```

## Trigger

From inside an active Pi session, the agent ran:

```bash
pi --help 2>&1 | head -80
pi --help 2>&1 | tail -80
```

These short-lived nested Pi invocations appear to auto-load `npm:context-mode`, start MCP servers, then exit quickly.

## Observed result

Immediately after the commands, two MCP ready files appeared:

```text
/tmp/context-mode-mcp-ready-42366
/tmp/context-mode-mcp-ready-42413
```

Corresponding processes:

```text
42366 /home/gbr/.bun/bin/bun /home/gbr/.nvm/versions/node/v24.14.1/lib/node_modules/context-mode/server.bundle.mjs
42413 /home/gbr/.bun/bin/bun /home/gbr/.nvm/versions/node/v24.14.1/lib/node_modules/context-mode/server.bundle.mjs
```

Both were in `Rl` state and pegged at ~100% CPU each:

```text
PID    PPID STAT %CPU COMMAND
42366  1391 Rl   100  bun .../context-mode/server.bundle.mjs
42413  1391 Rl   100  bun .../context-mode/server.bundle.mjs
```

`PPID=1391` was `systemd --user`, so they had become orphaned.

`kill` / `SIGTERM` did not stop them. `kill -KILL` was required.

## Expected

[...]

PR #828, opened 2026-06-14
fix(pi): backport #813 (lazy MCP bridge bootstrap) to main — fixes hang on pi list/update/install

unship:
## Problem

`pi list`, `pi update`, and `pi install` print their output and then **hang forever** when context-mode is installed (Ctrl-C required). Reported upstream as [earendil-works/pi#5687](https://github.com/earendil-works/pi/issues/5687).

## Root cause

Pi runs every extension's factory (`piExtension(pi)`) during module discovery — for *every* invocation, including the package subcommands and `--help`/`--version`, none of which start a session. The factory spawned `server.bundle.mjs` as a long-lived stdio child whose open handles keep the Node event loop alive, so a one-shot command never exits. The #534 argv denylist only covered help/version, so the package subcommands slipped through.

## Fix

This is a **clean backport of #813** ("lazy bootstrap MCP bridge for agent turns") to `main`. #813 is already on `next`, but `main` — what npm currently ships — still has the bug, so users hit the hang today.

#813 starts the bridge lazily from `before_agent_start` (the first lifecycle signal that Pi is about to dispatch a model call), removes the brittle argv denylist, and keeps the `session_shutdown` cleanup. CLI-only paths never fire `before_agent_start`, so they never spawn a child; subagents (`pi --mode json -p --no-session`) do, and the handler awaits readiness so the `ctx_*` tools are registered before Pi snapshots the tool registry.

Cherry-picked with provenance (`git cherry-pick -x`); credit to @Gandem / #813.

## Verification (local, on this branch against `main`)

- `tsc --noEmit` clean — no `next`-only dependencies.
- Full pi suite green: `pi-extension.test.ts` + `pi-help-skip-bootstrap` + `pi-no-orphan-on-help` + `pi-mcp-bridge` + `pi-extension-workspace-resolve` → **126 passed**.

## Note on base branch

Targets `main` deliberately, to fix the released line. If you'd rather let #813 reach `main` via the normal `next` → `main` merge, feel free to close — opening this so the published version is fixed sooner.

Fixes the extension side of earendil-works/pi#5687.


mksglu:
@unship We have conflicts

mksglu:
The target branch SHOULD BE `next` NOT `main`

mksglu:
Thanks for flagging the Pi install hang, @unship! 🙏 The underlying fix — lazy MCP-bridge bootstrap (#813, 5c18c673) — is **already on `next`** and ships in the next release, so this backport is now superseded and its remaining diff no longer carries the runtime change. Closing as **superseded by #813**. Really appreciate the contribution — the root cause you pointed at was exactly right.

PR #833, opened 2026-06-15
fix(server): exit on fatal uncaught exception storm

### LINKED ISSUE #830: Orphaned MCP server pegs 100% CPU for hours after client exits (uncaughtException handler logs instead of exiting)
# Orphaned MCP server pegs 100% CPU for hours after the client exits

**An uncaught-exception storm, kept alive by a `process.on("uncaughtException")` handler that logs instead of exiting.**

| | |
|---|---|
| Repo | github.com/mksglu/context-mode |
| Plugin version | 1.0.162 |
| MCP SDK | `@modelcontextprotocol/sdk` ^1.26.0 |
| OS | macOS 15 (Darwin 24.5.0), Apple Silicon |
| Node | 22.21.1, installed via **Volta** |
| Client | Claude Code (CLI) |
| Severity | High — a leaked orphan burns a full CPU core indefinitely; multiple per day |

---

## TL;DR

1. When a Claude Code session ends, its context-mode MCP server is sometimes left **orphaned**.
2. Instead of exiting, the orphan enters an **infinite uncaught-exception loop** and pins one CPU core at **100%**. One captured instance had **293 minutes (~4.9 h) of CPU time** when killed.
3. The loop never stops because the plugin installs `process.on("uncaughtException", … log …)` with **no `process.exit()`** — it swallows the fatal error and keeps running.
4. Each looped exception formats a full stack trace (custom `Error.prepareStackTrace` → per-frame source-position resolution), which is what burns the CPU.
5. The same storm runs in the event loop's *check* phase, **starving the plugin's own 30 s parent-death watchdog**, so it can't self-reap.
6. They accumulate. Several leaked orphans → the whole machine slows down.

A single fix — make the `uncaughtException` handler exit — stops both the leak and the 100% CPU.

---

## Impact / symptom (observed repeatedly)

Three separate orphans observed over ~2 days on one machine, each a `node start.mjs` pinned at ~100% CPU with CPU-time tracking wall-clock 1:1:

```
# ps aux  (TIME column = accumulated CPU time)
PID     %CPU   TIME        COMMAND
55317   100.1  1414:00.19  node …/context-mode/1.0.162/start.mjs   # ~23.5 h of CPU
25283   100.x  1722:13.44  node …/context-mode/1.0.162/start.mjs   # ~28.7 h of CPU
51696   100.8   293:48.20  node …/context-mode/1.0.162/start.mjs   # ~4.9 h, caught live
```

Contrast — other stdio MCP servers (`mcp-remote`, `context7-mcp`) orphaned on the **same machine at the same time** sit at **0.0% CPU**. Only context-mode spins. So this is plugin-specific, not generic process orphaning.

---

## Environment detail that matters: Node via a shim (Volta)

`node` resolves to a Volta shim, so every server is a **3-level tree**:

```
$ which node
/Users/me/.volta/bin/node          # this is a shim (a small Rust binary)
$ node -e 'console.log(process.execPath)'
/Users/me/.volta/tools/image/node/22.21.1/bin/node   # the real binary it forwards to
```

```
Claude Code ──▶ volta-shim (node, ~4 MB RSS) ──▶ real node (the MCP server, fat RSS)
```

After Claude Code exits uncleanly, the shim is reparented to launchd (pid 1) but **lingers**; the real server keeps running as its child:

```
PID     PPID   %CPU   COMMAND
51693   1      0.0    node …/start.mjs     # volta-shim, idle, orphaned
51696   51693  100.8  node …/start.mjs     # the server, spinning
```

The launch command is the plain single process (`.claude-plugin/plugin.json`):

```json
{ "context-mode": { "command": "node", "args": ["${CLAUDE_PLUGIN_ROOT}/start.mjs"] } }
```

The thin→fat pair is **not** a context-mode re-exec (the `start.mjs` Bun re-exec at lines 67–93 is gated `process.platform === "linux"` and is skipped on macOS). It is purely the Volta shim. This matters because the plugin's watchdog keys on `process.ppid` (see below), and under a shim that ppid is the shim, not the client.

---

## Root cause (confirmed from a live `sample`)

A 5-second `sample(1)` of a real spinner (pid 51696). Main thread, top-down — read it as a call chain:

```
node::SpinEventLoopInternal
  uv_run → uv__run_check → node::Environment::CheckImmediate          ← a setImmediate callback fires every tick
    node::InternalMakeCallback → v8::Function::Call → Execution::Call
      v8::internal::Isolate::ReportPendingMessages                     ← reporting a PENDING (uncaught) exception
        MessageHandler::ReportMessage → ReportMessageNoExceptions
          node::errors::TriggerUncaughtException                       ← UNCAUGHT EXCEPTION path, every tick
            … → v8::internal::Accessors::ErrorStackGetter              ← reads error.stack
              ErrorUtils::GetFormattedStack → ErrorUtils::FormatStackTrace
                node::PrepareStackTraceCallback → (JS) Error.prepareStackTrace
                  Builtins_ArrayPrototypeJoin
                    Builtins_CallSitePrototypeToString                 ← callSite.toString() per frame
                      SerializeJSStackFrame → AppendFileLocation
                        CallSiteInfo::GetLineNumber → ComputeSourcePosition
                          v8::internal::SourcePositionTableIterator::Advance()   ← hot leaf
```

Leaf self-time (`sample` "Sort by top of stack", excluding idle background threads which sit in `__psynch_cvwait`):

```
252  SourcePositionTableIterator::Advance
189  TranslatedState::CreateNextTranslatedValue        (deopt translation, to format frames)
180  CallSiteInfo::GetMethodName
159  DeoptTranslationIterator::NextOpcode
120  SafepointTable::GetEntry
 60  Script::GetPositionInfoInternal
 43  OptimizedJSFrame::Summarize
 41  SharedFunctionInfo::DebugName
 33  SerializeJSStackFrame
 31  Builtins_ArrayPrototypeJoinImpl
 …   (all stack-trace formatting / frame-translation machinery)
```

### The three pieces

**1. Trigger — a `setImmediate` callback throws every tick.** context-mode schedules
session-DB / index writes via `setImmediate` after activity (`server.bundle.mjs`):

```js
// ~offset 569235
…Ru(), setImmediate(() => VE(Ki(), t, r)),
(t === "ctx_execute" || t === "ctx_execute_file" || t === "ctx_batch_execute")
  && setImmediate(() => FE({ sessionDbPath: Ki(), toolName: t, bytesReturned: r })) …
// ~offset 569464
function cr(t, e = "unknown") { te.bytesIndexed += t; Ru(); t > 0 && setImmediate(() => ZE({ sessionDbPath: Ki(), source: e, bytesAvoided: t })) }
```

After the client disconnects, one of these check-phase callbacks throws (e.g. writing to a session resource that is no longer valid). The stack confirms the throw originates in a `CheckImmediate` callback.

**2. Loop-keeper — the swallowing handler.** `server.bundle.mjs` (~offset 561717):

```js
process.env.CONTEXT_MODE_EMBEDDED_PLUGIN_TOOLS !== "1" && (
  process.on("unhandledRejection", t => process.stderr.write(`[context-mode] unhandledRejection: ${t}\n`)),
  process.on("uncaughtException",  t => process.stderr.write(`[context-mode] uncaughtException: ${t?.message ?? t}\n`))
);
```

This **logs and never `process.exit()`s**. Without it, the first uncaught exception would
terminate the process (clean exit → no leak). With it, a repeating throw becomes an
**unbounded storm** the process can never escape.

**3. CPU amplifier.** Reporting each uncaught exception formats a full stack trace through a
custom `prepareStackTrace`, resolving source positions per frame
(`ComputeSourcePosition` / `SourcePositionTableIterator::Advance`) — ~1 ms+ each. Fired
continuously ⇒ 100% of one core.

### Why it never self-reaps

context-mode *does* ship a parent-death watchdog (`server.bundle.mjs`, fn `LE`, ~offset 517479):

```js
function LE(t) {
  let e = t.checkIntervalMs ?? _L(),            // _L() defaults to 30000 ms
      r = t.isParentAlive ?? yL,                // ppid-based liveness predicate
      n = false,
      o = () => { n || (n = true, t.onShutdown()) },
      s = setInterval(() => { r() || o() }, e); // poll: parent dead → shut down
  s.unref();
  let i = ["SIGTERM", "SIGINT"];
  process.platform !== "win32" && i.push("SIGHUP");
  for (let c of i) process.on(c, o);
  return process.stdin.isTTY || process.stdin.on("end", o), /* … */;
}
```

It cannot fire here for two compounding reasons:

- **Event-loop starvation.** The exception storm lives in the *check/immediate* phase, which
  never yields to the *timers* phase, so the `.unref()`'d 30 s `setInterval` never runs.
- **Shim-blind ppid check (secondary).** Even if it ran, `isParentAlive` inspects
  `process.ppid` — under Volta that is the lingering shim (still alive), so it would read
  "parent alive" and not shut down. It also only listens for stdin `"end"`, not
  `"error"`/`"close"`, so an unclean transport teardown misses that path too.

---

### A) In the wild (this is how it was confirmed)

The 100% spin requires the **real Claude Code stdio/transport teardown**. To catch it,
a `launchctl` poller (run every 60 s) snapshots any orphaned `start.mjs` that is
>50% CPU and has **no live `claude` process in its ancestry** (so live sessions are never
touched), recording the process tree, open fds, and a 5 s `sample`. It caught pid 51696
above. The catcher (also the workaround — see below):

```bash
#!/bin/zsh
# catch (and optionally kill) an orphaned context-mode server spinning at 100% CPU
THRESHOLD=50
has_claude_ancestor() {   # true if pid or any ancestor is the Claude Code binary
  local p="$1" hops=0 cmd
  while [ "${p:-0}" -gt 1 ] && [ "$hops" -lt 14 ]; do
    cmd="$(ps -o command= -p "$p" 2>/dev/null)"
    case "$cmd" in claude|claude\ *) return 0;; */share/claude/versions/*) return 0;; esac
    p="$(ps -o ppid= -p "$p" 2>/dev/null | tr -d ' ')"; [ -z "$p" ] && break; hops=$((hops+1))
  done
  return 1
}
for pid in $(pgrep -f 'context-mode.*start.mjs'); do
  cpu="$(ps -o %cpu= -p "$pid" | tr -d ' ')"; cpuint="${cpu%%.*}"
  [ "${cpuint:-0}" -lt "$THRESHOLD" ] && continue        # not spinning
  has_claude_ancestor "$pid" && continue                 # live session — skip
  {
    echo "pid=$pid cpu=$cpu"; date
    p="$pid"; while [ "${p:-0}" -gt 1 ]; do ps -o pid,ppid,%cpu,rss,etime,command -p "$p" | tail -n +2; p="$(ps -o ppid= -p "$p"|tr -d ' ')"; [ -z "$p" ] && break; done
    lsof -p "$pid" 2>/dev/null | grep -iE 'PIPE|KQUEUE|CHR|sock'
    /usr/bin/sample "$pid" 5
  } > "$HOME/ctx-capture-$(date +%s)-pid$pid.txt"
  # kill -9 "$pid"   # uncomment to auto-reclaim the core
done
```

### B) What does NOT reproduce it (itself a finding)

A synthetic harness — spawn `start.mjs` with a plain FIFO/pipe as stdin, then kill the
parent — does **not** reproduce the spin. The server exits cleanly:

- If stdin is closed cleanly, the `stdin.on("end", …)` watchdog path fires → exit within seconds.
- If no MCP `initialize` handshake is sent, the server exits on its own init-timeout (~20 s).

In both synthetic cases the server terminates normally, which is why the bug looks invisible
in a unit test. It only manifests under the real client's transport teardown, where an
in-flight/post-disconnect `setImmediate` callback throws into the swallowing handler. This
strongly implies the throw is tied to the **real session/DB state at disconnect**, not to a
clean EOF.

### Open fds at capture (no usable stdin; only pipes + kqueues remain)

```
node 51696   3u KQUEUE   4 PIPE  5 PIPE   7u KQUEUE   9u KQUEUE   11r /dev/null   12 PIPE  13 PIPE   14 NPOLICY
```

---

## Proposed fix

Fix (1) is the keystone; (2) stops the throw source; (3)–(4) are hardening.

1. **Make the fatal-error handlers terminate.** After logging, exit — especially when the
   client/transport is gone. An MCP stdio server should not log-and-continue through an
   uncaught exception:
   ```js
   process.on("uncaughtException", t => {
     process.stderr.write(`[context-mode] uncaughtException: ${t?.stack ?? t}\n`);
     process.exit(1);   // ← the missing line
   });
   ```
   (If a soft handler is desired during normal operation, gate the swallowing behind a
   "client connected" flag and always exit once the transport has closed.)

2. **Stop post-disconnect work.** On transport `close`/`error`, cancel the pending
   `setImmediate` flush/index callbacks (`Ru`/`VE`/`FE`/`ZE`) and shut down — don't keep
   scheduling session-DB writes for a dead session.

3. **Terminate on stdin `error`/`close`, not only `end`.** Unclean teardown errors the pipe
   rather than ending it; the current watchdog misses that.

4. **Don't depend on `process.ppid` for liveness.** Under a node shim (Volta/asdf/fnm/mise)
   the immediate parent is the shim, not the client, so the ppid check is wrong. And a
   `.unref()`'d `setInterval` can't fire while the event loop is saturated. Prefer a
   stdin-lifetime signal (above) as the primary exit trigger.

---

## Workaround (until fixed)

Reap any orphaned spinner (an orphaned `start.mjs` whose shim parent has ppid 1):

```bash
for s in $(pgrep -f 'context-mode.*start.mjs'); do
  [ "$(ps -o ppid= -p "$s" | tr -d ' ')" = "1" ] && { pkill -9 -P "$s"; kill -9 "$s"; }
done
```

Or run the catcher in section A on a 60 s `launchctl` interval with the `kill -9` line enabled.

ken-jo:
## What / Why / How

Fixes #830.

The reported failure is an orphaned context-mode MCP server that enters an uncaught-exception storm after the Claude Code client exits. The important part is not only orphaning: the standalone MCP process has a `process.on("uncaughtException")` handler that logs the fatal error but does not terminate, so a repeated `setImmediate` exception can keep the process alive and burn CPU indefinitely.

With this PR applied, standalone MCP `uncaughtException` handling logs the fatal error with a synchronous stderr write and then exits with status 1.

This PR:

- makes the standalone MCP server fail fast on uncaught exceptions instead of surviving the exception storm
- keeps `unhandledRejection` behavior unchanged
- preserves the `CONTEXT_MODE_EMBEDDED_PLUGIN_TOOLS !== "1"` guard so OpenCode/Kilo native plugin imports do not inherit process-wide crash handlers
- adds the regression case to the existing `tests/server-stdin-eof-exit.test.ts` server lifecycle test, using the real `server.bundle.mjs`, an injected repeated `setImmediate` exception storm, and a non-zero exit assertion
- rebuilds `server.bundle.mjs` and `cli.bundle.mjs`

## Affected platforms

Agent / CLI scope:

- [x] Claude Code
- [x] Cursor
- [x] VS Code Copilot
- [x] JetBrains Copilot
- [x] Gemini CLI
- [x] Qwen Code
- [x] Codex CLI
- [x] OpenClaw / Pi
- [x] Kiro
- [x] Antigravity
- [x] Zed
- [x] OpenCode / KiloCode legacy MCP child path, if a local MCP child is still configured
- [ ] OpenCode / KiloCode native plugin import path — intentionally not affected because `CONTEXT_MODE_EMBEDDED_PLUGIN_TOOLS=1` keeps process-wide fatal handlers out of the host process

OS scope:

- [x] Linux
- [x] macOS
- [x] Windows

This is a standalone stdio MCP server behavior change. It applies to agent/CLI integrations that launch `node start.mjs` and therefore run the bundled MCP server process. It does not change native in-process plugin imports.

## Root Cause

Git archaeology:

- `939c5ca` / #103 added the original lifecycle guard to prevent orphaned MCP servers.
- `1c66d5c` / #311 and `259077c` / #389 tightened parent/grandparent and stdin-EOF shutdown behavior for orphan cleanup.
- `09efefd` / #597 added the plugin-native OpenCode/Kilo path and the `CONTEXT_MODE_EMBEDDED_PLUGIN_TOOLS` guard; that invariant must remain because importing server code for native tools must not install host-wide crash handlers.
- Earlier server history already had the fatal-process listener pattern, but the standalone `uncaughtException` handler only wrote to stderr and returned.

The #830 failure happens because once a standalone MCP process is in an uncaught-exception loop, returning from the `uncaughtException` listener keeps Node alive. That can starve the lifecycle watchdog the older orphan fixes rely on. The fix is therefore to terminate only the standalone MCP process on uncaught exceptions, while preserving the embedded-plugin guard added for OpenCode/Kilo.

- lifecycle orphan cleanup remains intact
- stdin EOF behavior is not changed
- plugin-native OpenCode/Kilo imports still skip the process-wide fatal handlers

## Validation

| --- | --- | --- |
| Linux local, commit `1395a1c`, Node v24.14.0, Claude Code 2.1.177 | `npm run build` | passed |
| Linux local, commit `1395a1c`, Node v24.14.0 | `npm run typecheck` | passed |
| Linux local, commit `1395a1c`, Node v24.14.0 | `npx vitest run tests/server-stdin-eof-exit.test.ts --reporter=dot` | 5 passed |
| macOS via `ssh my-mac`, commit `1395a1c`, Node v26.0.0, npm 11.12.1 | `npm install`, `npm run build`, `npm run typecheck`, `npx vitest run tests/server-stdin-eof-exit.test.ts --reporter=dot` | passed; 5 tests passed |
| Windows via `ssh my-window`, commit `1395a1c`, Node v24.13.0, npm 11.6.2 | `npm install`, `npm run build`, `npm run typecheck`, `npx vitest run tests/server-stdin-eof-exit.test.ts --reporter=dot` | passed; 5 tests passed |
| Linux local, Claude Code 2.1.177 / Haiku | strict `--mcp-config` to this branch's `node start.mjs`; actual `ctx_execute` call ran `console.log('cm830-agent-smoke')` | `mcp_connected=1`, `ctx_execute_called=1`, `ctx_execute_completed=1` |
| macOS via `ssh my-mac`, Claude Code 2.1.177 / Haiku | strict `--mcp-config` to this branch's `node start.mjs`; actual `ctx_execute` call ran `console.log('cm830-agent-smoke')` | `mcp_connected=1`, `ctx_execute_called=1`, `ctx_execute_completed=1` |
| Windows via `ssh my-window`, Claude Code 2.1.177 / Haiku | strict file-based `--mcp-config` to this branch's `node start.mjs`; actual `ctx_execute` call ran `console.log('cm830-agent-smoke')` | `mcp_connected=1`, `ctx_execute_called=1`, `ctx_execute_completed=1` |
| Linux local | `npx vitest run tests/opencode-plugin.test.ts --reporter=dot` | 57 passed |

Behavioral evidence:

- BEFORE: the same injected `setImmediate` uncaught-exception storm left the server process alive until the harness killed it.
- AFTER: the server exits non-zero under the same storm, and the regression test passes.

Agent-working smoke:

- Claude Code 2.1.177 was run with Haiku and strict MCP config pointing directly at this branch's `node start.mjs`.
- The model performed a real context-mode `ctx_execute` MCP call on Linux, macOS, and Windows.
- Each OS confirmed MCP connection, tool dispatch, and successful tool completion in Claude debug output.

- the new regression case is integrated into the existing server lifecycle test file instead of adding a separate test file
- the regression uses the real generated `server.bundle.mjs`, not a mocked handler
- OpenCode/Kilo embedded import behavior remains protected by the existing `CONTEXT_MODE_EMBEDDED_PLUGIN_TOOLS` guard
- stdin EOF lifecycle behavior remains covered in the same lifecycle test file

The changed handler is installed only for standalone MCP server processes. That path is used by Claude Code and other stdio MCP clients that launch `start.mjs`. Embedded native plugin imports are intentionally excluded by the existing guard, so OpenCode/Kilo host-process behavior is not changed.

The fix is OS-neutral process behavior and was verified on Linux, macOS, and Windows, including real Claude Code MCP tool calls on all three.

## Checklist

- [x] Regression coverage integrated into an existing relevant test file, or new test file justified
- [x] `npm run build` passes
- [x] Targeted tests pass
- [x] Generated bundles updated when source changes require it
- [x] Docs not needed: fatal standalone MCP crash behavior only
- [x] 3OS validation completed
- [x] Agent-working smoke completed for agent-facing changes
- [x] Targets `next` branch


mksglu:
Don't use new test file @ken-jo

ken-jo:
Sorry, I will fix quickly.

ken-jo:
Manual validation completed on Windows.

Environment:
- Windows local
- Node v24.13.0
- pnpm 10.23.0
- Claude Code 2.1.177
- PR worktree commit: `1395a1c`
- Pre-patch negative-control commit: `79e7bb5`

Checks passed on the PR worktree:
- `pnpm run build` passed
- `pnpm run typecheck` passed
- `pnpm exec vitest run tests/server-stdin-eof-exit.test.ts --reporter=dot` passed: 1 file, 5 tests

Real agent smoke:
- Ran Claude Code with `--strict-mcp-config` pointing directly at this PR worktree's `start.mjs`.
- Claude successfully connected to the `context-mode` stdio MCP server.
- `mcp__context-mode__ctx_execute` was allowed and called successfully.
- The MCP tool returned the expected output: `cm833-agent-smoke-debug`.
- Debug log showed `ctx_execute` completed successfully and the MCP connection closed cleanly.
- Checked for leftover `context-mode-pr-833 ... start.mjs` processes after the run: none found.

Before/after storm harness comparison:

Pre-patch parent commit `79e7bb5`:
```json
{
  "closedWithin2s": false,
  "exitCode": null,
  "killedByHarness": true,
  "stderrTail": "[issue-830] starting storm with uncaughtException listeners=1\n",
  "verdict": "bug-survived-storm"
}
```

Patched commit `1395a1c` with the same harness:
```json
{
  "closedWithin2s": true,
  "exitCode": 1,
  "killedByHarness": false,
  "stderrTail": "[issue-830] starting storm with uncaughtException listeners=1\n[context-mode] uncaughtException: issue-830-storm\n",
  "verdict": "fixed-exits-nonzero"
}
```

Note: I initially hit a Windows-only false positive in my external harness because `node --import C:\...` is not a valid ESM import URL on Windows. I corrected the harness to pass a `file://` URL before recording the final before/after results above.

Conclusion: verified. The pre-patch server survives the injected uncaught-exception storm and has to be killed by the harness; this PR exits non-zero under the same storm. The real Claude Code MCP smoke also passed.

PR #680, opened 2026-05-23
feat(routing): skip routing-block injection for MCP-less subagents (#641)

sebastianbreguel:
## Summary

Implements **Option B** from issue #641: filter routing-block injection per subagent based on `subagent_type`, skipping the ~580-token XML payload for agents known to lack `ctx_*` MCP tools.

The block is currently injected unconditionally into every `Agent`/`Task` spawn at [`hooks/core/routing.mjs:819-833`](https://github.com/mksglu/context-mode/blob/next/hooks/core/routing.mjs#L819-L833). Issue #641 surveyed common harnesses and found ~70% of spawned subagents have no `ctx_*` surface — for them the block is 100% dead text, paid for in tokens but never actionable.

## Approach

**Default-on, no opt-in flag.** Per the issue's data (`~70% dead text`), skip is the majority case; an opt-in flag would put the burden on the user to discover and enable the right behavior.

**Name-based skip-list, not tool-detection.** Claude Code's `PreToolUse` payload for a Task call exposes `{prompt, subagent_type}` only — there is no `tools_available` field at hook time. The skip-list-by-name is the only signal available today. (Forward-compatible: if a tool list ever lands in the payload, a tools-present short-circuit can sit ahead of the name check.)

**Override via `CTX_ROUTING_SKIP_AGENTS` env var.** Comma-separated tokens; bare or `+`-prefixed names ADD to the skip-list, `-`-prefixed names REMOVE. Wildcard patterns (`plugin-ns:*`) supported and removable.

**Bash → general-purpose conversion preserved.** Orthogonal to skip — fires whether or not the block is injected.

## Baked skip-list

- `Explore`, `Plan` — Claude Code built-ins (read-only/planning, no MCP)
- `caveman:*`, `pr-review-toolkit:*`, `feature-dev:*` — plugin-namespaced agents called out as MCP-less in #641

Users with custom MCP-less agents add via env var without forking.

## Return-shape contract

| Case | Returned decision |
|---|---|
| skip = false (current behavior) | `{action: "modify", updatedInput: {...prompt + block...}}` |
| skip = true + `subagent_type === "Bash"` | `{action: "modify", ...}` with Bash → general-purpose rewrite but no block |
| skip = true + `subagent_type !== "Bash"` | `null` (passthrough, no transform) |

## Test plan

- [x] All existing `tests/core/routing.test.ts` Agent scenarios still pass (general-purpose injection, Bash conversion, fallback field discovery)
[...]

Issue #641, opened 2026-05-19
proposal: MINIMAL_ROUTING env flag + per-agent tools_available filter to reduce SessionStart routing-block context cost


alex-hairexplained:
## Summary

The current routing block emitted by `hooks/routing-block.mjs` (lines 18–79) injects ~700 tokens of XML as SessionStart `additionalContext` on every Claude Code session — including every background sub-agent spawn. For harnesses that orchestrate many agents per day, this becomes a meaningful tax on the context window, and a large fraction of that tax is paid by agents that have no `ctx_*` MCP tools available and therefore cannot act on the instructions.

This issue proposes two complementary mitigations that preserve current behaviour by default.

## Quantified cost

Measured against a real multi-agent harness (one parent thread + 6–10 background spawns per session, several sessions per day):

- Block size: ~700 tokens (75 lines of XML after frontmatter strip).
- Per session arc (parent + ~6 spawns): ~7 × 700 = ~4.9K tokens.
- Daily rate at typical harness throughput (10–15 spawns/day): ~7–10K tokens/day.
- Cumulative across a 30-day sprint: ~210–300K tokens — roughly one full session's worth of context burned on identical boilerplate.

## ~70% of agent types receive the block as dead text

A survey of the agent types this harness routinely spawns shows the majority are loaded with `Read` / `Edit` / `Write` / `Grep` / `Glob` / `Bash` only — no MCP tools, and therefore no `ctx_*` surface to route to:

| Agent type | Has `ctx_*` MCP | Block relevance |
|---|---|---|
| general-purpose, parent Claude Code thread, researcher | Yes | Relevant |
| caveman:cavecrew-builder / investigator / reviewer | No | Dead text |
| pr-review-toolkit:code-reviewer, silent-failure-hunter | No | Dead text |
| feature-dev:code-architect / explorer / reviewer | No | Dead text |
| Explore, Plan | No | Dead text |

For these agents the block is 100% inert — they cannot call the tools it tells them to prefer, and the instructions are paid for in tokens but never consulted.

## Proposed options

Either option is backward-compatible. Both gated behind an opt-in flag would also be acceptable.

### Option A — `MINIMAL_ROUTING=1` env flag returning a 12-line caveman version

[...]

Issue #233, opened 2026-04-06
Skip ctx_commands injection in subagent sessions


movax01h:
## Problem

The `SessionStart` hook injects the full `<context_window_protection>` block into **every** Claude Code session, including subagents spawned via the `Agent` tool.

The `<ctx_commands>` section (~400 tokens) contains instructions for user-facing commands (`ctx doctor`, `ctx purge`, `ctx upgrade`, `ctx stats`) that are **useless in subagent contexts** — users cannot type these commands in a subagent session.

This wastes ~400 tokens per subagent spawn. In workflows that use multiple agents (e.g., plan-mode with Explore + Plan agents), this adds up quickly.

## Where

- `hooks/routing-block.mjs`, lines 64-78 — the `<ctx_commands>` section
- `hooks/sessionstart.mjs`, line 35 — `additionalContext = ROUTING_BLOCK` (always the full block, no conditional logic)

## Suggested fix

1. Detect subagent context — either via an environment variable (e.g., `CLAUDE_PARENT_SESSION_ID`) or a field in the stdin JSON that Claude Code passes to hooks
2. When in a subagent session, omit the `<ctx_commands>` section from the routing block (or omit the entire routing block if MCP tools aren't available to subagents anyway)

A simple approach would be to split `createRoutingBlock()` into core guidance + user commands, and only include the user commands section for root sessions.

## Environment

- context-mode version: 1.0.75
- Claude Code CLI

murataslan1:
Validated the bug — `ctx_commands` (~407 tokens) is injected even when `CLAUDE_PARENT_SESSION_ID` is set:

```
CLAUDE_PARENT_SESSION_ID=parent-abc echo '...' | node hooks/sessionstart.mjs
→ ctx_commands present: True, token count approx: 407
```

The fix in `routing-block.mjs` is straightforward (add `{ includeCommands }` option to `createRoutingBlock()`), but I want to verify the detection method first:

**Question for @movax01h:** Does Claude Code actually set `CLAUDE_PARENT_SESSION_ID` for subagent sessions? Or is there a different env var / stdin field we should check? I don't want to ship a detection that's a no-op.

If you can share the env vars available in a subagent SessionStart hook (e.g. `env | grep CLAUDE` from inside a subagent), that would nail the detection.

The `routing-block.mjs` refactor to support `includeCommands` is ready to go once we confirm the detection signal.

murataslan1:
Confirmed in the current code.

[...]

PR #834, opened 2026-06-15
fix(hooks): respect subagent ctx tool availability

### LINKED ISSUE #794: PreToolUse WebFetch deny hard-blocks subagents that cannot access ctx_* MCP tools (isMCPReady is machine-global, tool availability is per-agent-context)
## Summary

The PreToolUse `WebFetch` redirect hard-blocks **subagents** whose tool context cannot invoke `ctx_*` MCP tools. `isMCPReady()` is a **machine-global** check (PID sentinel scan), but tool availability is **per-agent-context** — so the deny fires inside subagent contexts where the suggested alternatives don't exist, cutting off their only fetch path.

## Environment

- context-mode **1.0.162** (plugin via marketplace, autoUpdate on; `hooks/core/routing.mjs` identical to repo HEAD at report time)
- Claude Code CLI on **Windows 11**, model Opus 4.8
- context-mode MCP server running normally in the main session

## Symptom

Spawn a subagent via the `Agent` tool using a **fixed-tool agent type** — e.g. the built-in `claude-code-guide` (tools: Glob, Grep, Read, WebFetch, WebSearch — **no MCP, no ToolSearch**) — and have it call `WebFetch`. It receives:

```
context-mode: WebFetch redirected. Call mcp__plugin_context-mode_context-mode__ctx_fetch_and_index(url: "https://example.com", source: "...") to fetch + index the page, then mcp__plugin_context-mode_context-mode__ctx_search(queries: [...]) ...
```

The subagent cannot invoke any `mcp__plugin_context-mode_*` tool (not in its tool list, and it has no ToolSearch to load deferred schemas), so the advice is unusable and web access is fully blocked. In our runs, `general-purpose` subagents also reported the `ctx_*` tools as unregistered in their context.

The existing #724 mitigation (Agent-prompt ToolSearch bootstrap injection, `routing.mjs` ~817-841) does not help these agent types — they have no ToolSearch at all.

## Root cause

`routing.mjs` WebFetch branch (~line 800) wraps the deny in `mcpRedirect()` → `isMCPReady()` (`core/mcp-ready.mjs`), which scans `tmpdir()` for live `context-mode-mcp-ready-<PID>` sentinels. That check answers *"is a context-mode MCP server process alive on this machine?"* — not *"can the calling context invoke ctx_* tools?"*. While the main session's server is alive, the deny fires for **every** hook invocation on the machine, including subagent-originated tool calls.

## Discriminator (verified)

Claude Code's PreToolUse hook stdin carries **`agent_id` and `agent_type` fields for subagent-originated tool calls, and omits them for main-loop calls**. Verified by dumping hook stdin for both contexts (2026-06-08): the subagent payload's `agent_id` matched the ID returned by the `Agent` tool exactly; the main-loop payload had neither field. `session_id`, `transcript_path`, and `cwd` are identical in both — they cannot discriminate.

## Suggested fix (tested locally)

In `hooks/pretooluse.mjs` (the Claude Code adapter — the right layer, since `agent_id` is a Claude-Code-specific stdin field), pass `WebFetch` through in subagent contexts:

```js
const isSubagentContext = input.agent_id != null || input.agent_type != null;

const decision = (tool === "WebFetch" && isSubagentContext)
  ? null
  : routePreToolUse(tool, toolInput, process.env.CLAUDE_PROJECT_DIR, "claude-code", getSessionId(input));
```

Rationale: a subagent is itself a context-isolation boundary — the fetched page bytes never enter the main conversation; only the subagent's summary returns. Allowing WebFetch there is consistent with context-mode's goal of keeping raw bytes out of the primary context window.

Verified after patching:
- `claude-code-guide` subagent `WebFetch https://example.com` → succeeds ("Example Domain")
- main-session `WebFetch` → still redirected to `ctx_fetch_and_index` (behavior preserved)

## Possibly related (not reproduced)

The Bash `curl`/`wget` → `ctx_execute` redirect likely shares the same class of issue in subagent contexts (deny with advice the context can't follow). Left untouched here; reporting only the reproduced WebFetch case.

## Repro steps

1. Claude Code session with context-mode enabled (MCP server running).
2. Spawn: `Agent(subagent_type: "claude-code-guide", prompt: "Call WebFetch on https://example.com and report the verbatim response. No other tools, no retries.")`
3. Observe the verbatim `context-mode: WebFetch redirected...` error inside a context with no `ctx_*` tools.

### LINKED ISSUE #832: [Feature]: Env-var opt-out for subagent Agent-injection - foreground agents can hang on forced ctx_* calls
## Summary

The `Agent` PreToolUse hook appends the full `<context_window_protection>` routing block to **every** subagent prompt, instructing it that it **MUST** use `ctx_*` tools for analysis/fetch. When a subagent runs in the **foreground** (the default for Claude Code's `Agent` tool) and one of those forced `ctx_*` calls blocks (sandbox runtime startup, or a network call in `ctx_fetch_and_index` with no surfaced timeout), the **entire parent session freezes** until the user hard-interrupts, which can lose the subagent's work.

There is currently **no opt-out**. I'd like to request a first-class one (env var and/or per-subagent-type skip).

## Environment

- context-mode 1.0.162 (plugin via marketplace, autoUpdate on)
- Claude Code CLI, Windows 11, model Opus 4.8
- MCP server running normally in the main session (`ctx_doctor` all green)

## Why the existing fixes don't fully cover this

- **#724** (closed via PR #725) added the ToolSearch bootstrap so deferred `ctx_*` resolve. That fixed *one* stall cause (the tool-not-found loop). It did **not** remove the broader failure mode: a subagent that successfully *starts* a `ctx_*` call which then blocks. The follow-up mentioned there, "skip injection entirely for Explore-type read-only subagents", does not appear to have shipped: `hooks/core/routing.mjs` (1.0.162, around L817-838) still injects for all `Agent` spawns regardless of `subagent_type` (only `Bash` is special-cased, and it's rewritten to `general-purpose`).
- **#155** (closed): you noted the real cost is "latency inside the subagent", same root cause. The reporter asked for a whitelist/blacklist config; I don't see one in 1.0.162.
- **#794** (open): related per-agent-context `ctx_*` unavailability for fixed-tool subagents.

## Proposed (any one resolves it)

1. **Env-var opt-out**, e.g. `CONTEXT_MODE_DISABLE_AGENT_INJECTION=1`, checked in the `canonical === "Agent"` branch of `routing.mjs`, returning pass-through instead of the modified prompt.
2. **Per-subagent-type skip** (delivers the #724 follow-up): skip injection for read-only types (`Explore`, etc.), optionally a configurable allow/deny list as suggested in #155.
3. At minimum, **document** that the Agent injection exists and how to disable it.

Rationale for opt-out over removal: the injection is genuinely valuable in the **main** session (it kept ~28 MB out of context in one of my sessions). But a subagent has its own isolated, short-lived context, and the parent only ever sees the subagent's summary, so the in-subagent `ctx_*` benefit is small while the hang risk is real. An opt-out keeps the default behavior for those who want it.

## Current workaround

A local self-healing SessionStart hook that strips the `Agent` matcher from the plugin's `hooks.json` after each update (idempotent, pure Node stdlib, never imports from the plugin dir). It works, but it edits plugin-cache files and must re-apply on every version bump, so a supported flag would be much cleaner.

Happy to send a PR for option 1 if you're open to it.

ken-jo:
## What / Why / How

Fixes #794.
Fixes #832.

Claude Code subagent-originated `WebFetch` calls can be blocked by the main-session MCP readiness check even when that subagent cannot invoke any `ctx_*` MCP tools. In that state the hook denies `WebFetch` and tells the subagent to call `ctx_fetch_and_index`, but the suggested tool is not available in the subagent's tool surface.

Claude Code `Agent` prompt injection also had no supported opt-out. The default injection is useful for subagents that can use context-mode tools, but foreground subagents can hang or become unusable when forced into `ctx_*` calls that are unavailable or blocking.

With this PR applied, Claude Code PreToolUse distinguishes machine-global MCP readiness from the current caller's `ctx_*` availability:

- main-session `WebFetch` still redirects to `ctx_fetch_and_index`
- subagent-originated `WebFetch` passes through when the hook payload has `agent_id` or `agent_type`
- default `Agent` prompt injection remains unchanged
- `CONTEXT_MODE_DISABLE_AGENT_INJECTION=1` disables only `Agent` prompt injection
- existing subagent invariants are preserved: ToolSearch bootstrap remains present, and `<ctx_commands>` remains omitted
- regression coverage is added to the existing `tests/hooks/core-routing.test.ts`
- the new env var is documented in `README.md`

## Affected platforms

Agent / CLI scope:

- [x] Claude Code
- [ ] Cursor
- [ ] VS Code Copilot
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw / Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All standalone MCP server launchers
- [ ] All platforms

OS scope:

- [x] Linux
- [x] macOS
- [x] Windows
- [x] All operating systems

Scope notes:

- This is a Claude Code PreToolUse hook-routing change. The `agent_id` / `agent_type` discriminator is Claude Code-specific, so other adapters are not changed to infer subagent tool availability.
- `routePreToolUse()` keeps `mcpToolsAvailable` defaulting to `true`, so existing callers preserve the current redirect behavior unless an adapter explicitly marks the caller as unable to invoke `ctx_*`.
- The `CONTEXT_MODE_DISABLE_AGENT_INJECTION` opt-out affects `Agent` prompt injection only. It does not disable SessionStart routing guidance, MCP redirects in the main session, or the context-mode MCP server.

## Root Cause

Git archaeology:

- `9e29a94` / PR #683 introduced the WebFetch redirect/refusal path and the current "redirect, not restriction" wording.
- `b1ba5cc` / PR #725 added the Claude Code ToolSearch bootstrap to subagent prompts so deferred `ctx_*` schemas can be loaded before use.
- `245cd8a` / PR #233 made subagent routing blocks omit `<ctx_commands>`, because subagents cannot use meta commands such as stats/doctor/upgrade/purge.

The failure happens because `isMCPReady()` is machine-global. It answers whether a context-mode MCP server process is alive on the machine, not whether the current Claude Code caller context can invoke `ctx_*` tools. Claude Code PreToolUse payloads include `agent_id` / `agent_type` for subagent-originated calls, so the adapter can pass that caller context into shared routing.

- main-session WebFetch continues to redirect to `ctx_fetch_and_index`
- default Claude Code `Agent` prompt injection remains enabled
- the #725 ToolSearch bootstrap remains in default Claude Code subagent prompts
- the #233 `<ctx_commands>` omission remains in subagent routing blocks

Related prior work checked:

- #680 was a closed, unmerged #641 proposal using a subagent-name skip-list for routing-block token cost. It is not a duplicate of this PR because it does not address #794 WebFetch denial and is not present on `next`.
- #207 was a closed OpenCode AGENTS.md global-injection-location toggle. It does not affect Claude Code `Agent` prompt injection.

This PR is intentionally scoped to Claude Code's hook adapter because `agent_id` and `agent_type` are Claude Code PreToolUse payload fields. The shared routing API keeps the previous default for all other adapters, so Cursor, VS Code Copilot, JetBrains Copilot, Gemini CLI, Qwen Code, OpenCode/Kilo, Codex CLI, OpenClaw/Pi, Kiro, Antigravity, Zed, and standalone MCP launchers do not get new subagent tool-availability behavior unless their adapter explicitly opts in later.

The runtime code is plain Node.js hook code and was validated on Linux, macOS, and Windows. No platform-specific path handling, shell parsing, or generated server bundle path is changed.

## Validation

Detailed maintainer-style validation results are reported in a PR comment rather than kept in the body.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1482,7 +1482,6 @@ That blocks loopback + RFC1918 + ULA in addition to the always-blocked ranges. U
 
 | Variable | Default | Purpose |
 |---|---|---|
-| `CONTEXT_MODE_DISABLE_AGENT_INJECTION` | unset | Set to `1`, `true`, `yes`, or `on` to disable Claude Code `Agent` prompt injection. Main-session routing guidance and MCP redirects stay active; this only skips the extra routing block appended to spawned `Agent` prompts. Useful when foreground subagents should not be forced into `ctx_*` calls. |
 | `CONTEXT_MODE_EXTERNAL_MCP_NUDGE_EVERY` | `10` | Cadence (in tool calls) at which the PreToolUse hook re-injects the "wrap large external-MCP payloads in `ctx_execute`" guidance. The original implementation ([#529](https://github.com/mksglu/context-mode/pull/529)) fired only once per session, which got lost after context compaction in MCP-heavy sessions (e.g. 50+ Jira/Slack/Notion calls — see [#567](https://github.com/mksglu/context-mode/issues/567) follow-up). The default re-fires every 10th matching call, keeping the guidance in the model's recent window. Range `[1, 100]`; invalid values fall back to `10`. Set to `1` for "every call" (most aggressive — adds ~250 tokens/call) or to a larger value for less frequent reminders. |
 
 ## Contributing
--- a/README.md
+++ b/README.md
@@ -1482,6 +1482,7 @@ That blocks loopback + RFC1918 + ULA in addition to the always-blocked ranges. U
 
 | Variable | Default | Purpose |
 |---|---|---|
+| `CONTEXT_MODE_DISABLE_AGENT_INJECTION` | unset | Set to `1`, `true`, `yes`, or `on` to disable Claude Code `Agent` prompt injection. Main-session routing guidance and MCP redirects stay active; this only skips the extra routing block appended to spawned `Agent` prompts. Useful when foreground subagents should not be forced into `ctx_*` calls. |
 | `CONTEXT_MODE_EXTERNAL_MCP_NUDGE_EVERY` | `10` | Cadence (in tool calls) at which the PreToolUse hook re-injects the "wrap large external-MCP payloads in `ctx_execute`" guidance. The original implementation ([#529](https://github.com/mksglu/context-mode/pull/529)) fired only once per session, which got lost after context compaction in MCP-heavy sessions (e.g. 50+ Jira/Slack/Notion calls — see [#567](https://github.com/mksglu/context-mode/issues/567) follow-up). The default re-fires every 10th matching call, keeping the guidance in the model's recent window. Range `[1, 100]`; invalid values fall back to `10`. Set to `1` for "every call" (most aggressive — adds ~250 tokens/call) or to a larger value for less frequent reminders. |
 
 ## Contributing
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -76,11 +76,6 @@ function getExternalMcpNudgeEvery() {
   return parsed;
 }
 
-function isTruthyEnv(name) {
-  const raw = process.env[name];
-  return raw === "1" || /^(true|yes|on)$/i.test(raw ?? "");
-}
-
 function defaultGuidanceId() {
   return process.env.VITEST_WORKER_ID
     ? `${process.ppid}-w${process.env.VITEST_WORKER_ID}`
@@ -829,8 +824,6 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
   // ─── Agent: inject context-mode routing into subagent prompts ───
   // Subagents cannot use ctx commands (stats/doctor/upgrade/purge) — omit that section (#233)
   if (canonical === "Agent") {
-    if (isTruthyEnv("CONTEXT_MODE_DISABLE_AGENT_INJECTION")) return null;
-
     const subagentType = toolInput.subagent_type ?? "";
     // Detect the correct field name for the prompt/request/objective/question/query
     const fieldName = ["prompt", "request", "objective", "question", "query", "task"].find(f => f in toolInput) ?? "prompt";
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -25,7 +25,8 @@ import { existsSync, mkdirSync, rmSync, rmdirSync, readdirSync, unlinkSync, open
  * redirect action — prevents agent from getting stuck when MCP tools
  * are unavailable. Applies to deny and modify actions that mention MCP alternatives.
  */
-function mcpRedirect(result) {
+function mcpRedirect(result, mcpToolsAvailable = true) {
+  if (!mcpToolsAvailable) return null;
   if (!isMCPReady()) return null;
   return result;
 }
@@ -73,6 +74,11 @@ function getExternalMcpNudgeEvery() {
     return EXTERNAL_MCP_NUDGE_DEFAULT;
   }
   return parsed;
+}
+
+function isTruthyEnv(name) {
+  const raw = process.env[name];
+  return raw === "1" || /^(true|yes|on)$/i.test(raw ?? "");
 }
 
 function defaultGuidanceId() {
@@ -604,8 +610,14 @@ function getPlatformSettingsPath(platform) {
  * @param {string} [sessionId] - Stable session identifier from hook payload. When
  *   provided, the guidance throttle uses it to scope marker files across hook
  *   invocations even when process.ppid shifts (Windows/Git Bash — see #298).
+ * @param {object} [options] - Runtime routing context from the adapter.
+ * @param {boolean} [options.mcpToolsAvailable=true] - False when the current
+ *   caller context cannot invoke ctx_* MCP tools even though an MCP server is
+ *   live on the machine (Claude Code fixed-tool subagents — #794).
  */
-export function routePreToolUse(toolName, toolInput, projectDir, platform, sessionId) {
+export function routePreToolUse(toolName, toolInput, projectDir, platform, sessionId, options = {}) {
+  const mcpToolsAvailable = options.mcpToolsAvailable !== false;
+
   // ─── Opt-in fail-closed gate (#468 follow-up) ───
   // Default behavior on security-module load failure is fail-OPEN (a stderr
   // warning is emitted but routing continues). Security-conscious users can
@@ -715,7 +727,7 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
             bytesAvoided: 8192,
             commandSummary: command.slice(0, 200),
           },
-        });
+        }, mcpToolsAvailable);
       }
       // All segments safe → allow through
       return null;
@@ -737,7 +749,7 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
         updatedInput: {
           command: `echo "context-mode: Inline HTTP redirected. Call ${t("ctx_execute")}(language, code) to fetch, derive your answer in code, and console.log() only the result — the raw response body stays in the sandbox instead of entering your conversation. Full network access. Retry the same call on a transient DNS error (EAI_AGAIN, ETIMEDOUT, ENETUNREACH)."`,
         },
-      });
+      }, mcpToolsAvailable);
     }
 
     // Build tools (gradle, maven, sbt) → redirect to execute sandbox (Issue #38, #406).
@@ -750,7 +762,7 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
         updatedInput: {
           command: `echo "context-mode: Build tool redirected. Call ${t("ctx_execute")}(language: \\"shell\\", code: \\"${safeCmd} 2>&1 | tail -30\\") to run the build and print only the tail — the verbose build log stays in the sandbox instead of entering your conversation. For more targeted output, replace \\"tail -30\\" with \\"grep -E '(error|warning|FAIL|✗|×)'\\" or similar, so only the lines that matter come back."`,
         },
-      });
+      }, mcpToolsAvailable);
     }
 
     // Skip the routing nudge for commands whose output is structurally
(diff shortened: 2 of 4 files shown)


ken-jo:
Maintainer-style validation report for the final PR head `e3ac14d`.

I validated this as a core hook-routing change, not only as a build/typecheck check.

### Claim / behavior checked

This PR is intended to fix two Claude Code hook-routing behaviors:

- #794: subagent-originated `WebFetch` should not be denied with an instruction to call `ctx_fetch_and_index` when that subagent cannot invoke `ctx_*` tools.
- #832: `Agent` prompt injection should have an explicit opt-out, without changing the default injection behavior.

The important invariant is that this must not weaken the normal main-session protection:

- main-session `WebFetch` must still redirect to `ctx_fetch_and_index`
- curl/wget, inline HTTP, and build-tool MCP-backed redirects must keep their existing behavior when MCP tools are available
- default Claude Code `Agent` injection must still include the ToolSearch bootstrap added for #725
- subagent routing must still omit `<ctx_commands>` from #233
- the new opt-out must only affect `Agent` prompt injection

### Git archaeology / root cause

I checked the relevant history before accepting the fix shape:

- `9e29a94` / PR #683 introduced the WebFetch redirect/refusal path and the current "redirect, not restriction" behavior.
- `b1ba5cc` / PR #725 added the Claude Code ToolSearch bootstrap so subagents can load deferred `ctx_*` schemas before trying to use them.
- `245cd8a` / PR #233 made subagent routing blocks omit `<ctx_commands>`, because subagents cannot use meta commands such as stats/doctor/upgrade/purge.
- #680 was checked as a possible duplicate, but it was a closed, unmerged #641 proposal around subagent-name skip-list/token cost. It does not fix #794 WebFetch denial and is not present on `next`.
- #207 was also checked, but it was an OpenCode AGENTS.md global-injection-location issue and does not cover Claude Code `Agent` prompt injection.

The root cause is that `isMCPReady()` is machine-global. It proves that a context-mode MCP server exists on the machine, but it does not prove that the current Claude Code caller context can invoke `ctx_*` tools. Claude Code PreToolUse payloads expose subagent-originated calls through `agent_id` / `agent_type`, so this PR passes caller-context availability into shared routing without changing the default for other adapters.

### Before / after behavior

Using an upstream/next comparison harness:

Before this PR:

- main `WebFetch`: denied and redirected to `ctx_fetch_and_index`
- subagent `WebFetch`: also denied and redirected to `ctx_fetch_and_index`, even though the subagent tool surface cannot call that tool
- default `Agent`: modified with the context-mode routing block, ToolSearch bootstrap present, `<ctx_commands>` omitted
- `CONTEXT_MODE_DISABLE_AGENT_INJECTION=1`: no effect

After this PR:

- main `WebFetch`: still denied and redirected to `ctx_fetch_and_index`
- subagent `WebFetch`: passes through when `agent_id` / `agent_type` is present
- default `Agent`: still modified with the context-mode routing block
- ToolSearch bootstrap remains present for Claude Code
- `<ctx_commands>` remains omitted from subagent routing
- `CONTEXT_MODE_DISABLE_AGENT_INJECTION=1`: `Agent` passes through

### Regression coverage

Coverage was integrated into the existing owner file `tests/hooks/core-routing.test.ts`; no new test file was added.

The final test coverage asserts:

- subagent `WebFetch` passes through when `mcpToolsAvailable: false`
- omitted options preserve the existing default redirect behavior
- a real `hooks/pretooluse.mjs` subprocess treats Claude Code subagent hook payloads as `ctx_*` unavailable
- MCP-backed redirects for curl, inline HTTP, build tools, and WebFetch pass through only when the caller context cannot invoke `ctx_*`
- default Claude Code `Agent` injection still adds `<context_window_protection>` and ToolSearch bootstrap
- default Claude Code `Agent` injection still omits `<ctx_commands>`
- `CONTEXT_MODE_DISABLE_AGENT_INJECTION=1` only disables `Agent` prompt injection

I also ran an independent Claude Haiku adversarial review against the final diff. It found no blockers. The suggested extra tests were the default Agent injection check, WebFetch in the MCP-backed redirect batch, and omitted-options default behavior; those were added in `e3ac14d`.

### 3OS validation

Final commit under test: `e3ac14d`.

Linux local:

- Node `v24.14.0`
- Claude Code `2.1.177`
- `npm run typecheck`: passed
- `npm run build`: passed, including `assert-bundle` and `assert-asymmetric-drift`
- `npx vitest run tests/hooks/core-routing.test.ts tests/core/routing.test.ts --reporter=dot`: passed, 120 tests
- direct `hooks/pretooluse.mjs` stdin/stdout smoke: passed
- Claude Code Haiku MCP smoke: `ctx_execute` returned `cm794832-agent-smoke-linux-final`

macOS via `ssh my-mac`:

- Node `v26.0.0`
- Claude Code `2.1.177`
- `npm run typecheck`: passed
- `npm run build`: passed, including `assert-bundle` and `assert-asymmetric-drift`
- `npx vitest run tests/hooks/core-routing.test.ts tests/core/routing.test.ts --reporter=dot`: passed, 120 tests
- direct `hooks/pretooluse.mjs` stdin/stdout smoke: passed
- Claude Code Haiku MCP smoke: `ctx_execute` returned `cm794832-agent-smoke-mac-final`

Windows via `ssh my-window`:

- Node `v24.13.0`
- Claude Code `2.1.177`
- `npm run typecheck`: passed
- `npm run build`: passed, including `assert-bundle` and `assert-asymmetric-drift`
- `npx vitest run tests/hooks/core-routing.test.ts tests/core/routing.test.ts --reporter=dot`: passed, 120 tests
- direct `hooks/pretooluse.mjs` stdin/stdout smoke: passed
- Claude Code Haiku MCP smoke: `ctx_execute` returned `cm794832-agent-smoke-windows-final`

The direct hook smoke on all three OSes verified the actual hook process contract, not only `routePreToolUse()` unit calls:

- main `WebFetch` denied and mentioned `ctx_fetch_and_index`
- subagent `WebFetch` passed through with no hook output
- default `Agent` prompt was modified
- default `Agent` prompt included `<context_window_protection>`
- default `Agent` prompt included the ToolSearch bootstrap
- default `Agent` prompt did not include `<ctx_commands>`
- `CONTEXT_MODE_DISABLE_AGENT_INJECTION=1` made `Agent` pass through

### Broad-suite note

I also ran the broader local Vitest suite. It did not pass cleanly, but the failures were separated from this PR:

- local broad run: 7 failed / 4271 passed / 38 skipped
- the failures were in `tests/executor.test.ts`, `tests/lifecycle.test.ts`, `tests/adapters/detect-claude-code-in-vscode.test.ts`, and `tests/util/postinstall-heal.test.ts`
- the same adapter/postinstall failures reproduce on clean `upstream/next`
- the lifecycle SIGTERM failure passed when isolated on both this PR and `upstream/next`, so it appears order/environment-sensitive rather than caused by this routing change
- none of the broad-suite failures touch the files changed by this PR: `README.md`, `hooks/core/routing.mjs`, `hooks/pretooluse.mjs`, `tests/hooks/core-routing.test.ts`

I did not treat the broad-suite failure as a pass, but I also do not see evidence that those failures are introduced by this PR.

### Conclusion

The #794 / #832 behavior is confirmed, the root cause is history-backed, and the final patch preserves the older #683, #725, and #233 invariants. I consider this merge-worthy after the final `e3ac14d` validation above.

PR #839, opened 2026-06-17
Fix PowerShell Unicode path handling

pradigmaz:
## Summary
- write Windows PowerShell temp scripts with a UTF-8 BOM
- set PowerShell console/output encoding to UTF-8 before user code runs
- add Windows coverage for non-ASCII stdout and cwd handling

## Verification
- npx vitest run tests/core/executor.test.ts
- npm run typecheck
- npm run build
- direct build/executor.js matrix: PowerShell 5 + pwsh7, ASCII + Japanese + Korean + Arabic + Hebrew + Greek + Hindi + Thai + accent + emoji paths


pradigmaz:
Context against #796: this PR is intentionally narrower.

It only fixes PowerShell script/output encoding for non-ASCII paths/stdout:
- UTF-8 BOM for `.ps1` temp scripts so Windows PowerShell reads them correctly
- UTF-8 console/output prelude for PowerShell/pwsh
- regression coverage for non-ASCII stdout + cwd

It does not change shell runtime selection, routing docs, or guidance text. Local Windows verification covered PowerShell 5 + pwsh7 across ASCII, Japanese, Korean, Arabic, Hebrew, Greek, Hindi, Thai, accented, and emoji path prefixes.

Issue #225, opened 2026-04-03
Codex PostToolUse hook fails: `context-mode hook codex posttooluse` is not a valid CLI subcommand


benzntech:
## Problem

The `~/.codex/hooks.json` is configured with the following commands:

```json
{
  "hooks": {
    "PreToolUse": [{ "hooks": [{ "type": "command", "command": "context-mode hook codex pretooluse" }] }],
    "PostToolUse": [{ "hooks": [{ "type": "command", "command": "context-mode hook codex posttooluse" }] }],
    "SessionStart": [{ "hooks": [{ "type": "command", "command": "context-mode hook codex sessionstart" }] }]
  }
}
```

However, `context-mode hook codex posttooluse` (and the other variants) are **not valid CLI subcommands**. Running `context-mode hook codex posttooluse` exits with a non-zero code, causing every Codex PostToolUse hook to fail.

## Root Cause

The `context-mode` binary does not expose a `hook codex <type>` dispatcher. When given `hook` as an argument, it starts the MCP server on stdio instead of dispatching to the hook script.

By contrast, `CodexAdapter.generateHookConfig` in the CLI bundle correctly generates **direct `node` invocations**:

```js
generateHookConfig(pluginRoot) {
  return {
    PostToolUse: [{ matcher: "", hooks: [{ type: "command", command: `node ${pluginRoot}/hooks/posttooluse.mjs` }] }],
    // ...
  }
}
```

This is inconsistent with the `hooks.json` that ends up in `~/.codex/`.

## Expected Behaviour

Either:
1. The CLI should expose a `context-mode hook codex <type>` dispatcher (consistent with the `cursor`/`kiro` adapters), **or**
2. The `~/.codex/hooks.json` setup should use direct `node` invocations as generated by `CodexAdapter.generateHookConfig`.

## Workaround

Replace `~/.codex/hooks.json` with direct node invocations:

```json
{
  "hooks": {
    "PreToolUse": [{ "matcher": "", "hooks": [{ "type": "command", "command": "node ~/.claude/plugins/cache/context-mode/context-mode/1.0.59/hooks/pretooluse.mjs" }] }],
    "PostToolUse": [{ "matcher": "", "hooks": [{ "type": "command", "command": "node ~/.claude/plugins/cache/context-mode/context-mode/1.0.59/hooks/posttooluse.mjs" }] }],
[...]

PR #851, opened 2026-06-21
fix(codex): MCP sentinel namespace + PreToolUse rewrites + tool annotations (#844 #845 #846)

### LINKED ISSUE #844: MCP readiness sentinel can be deleted across shared /tmp and isolated PID namespaces
## Summary

The MCP readiness sentinel can be deleted by a hook/process running in a different PID namespace while sharing the same `/tmp`. When this happens, context-mode hooks fail open even though the MCP process is still alive.

This affects the readiness gate that decides whether direct fetch/shell operations should be redirected to context-mode MCP tools.

## Observed behavior

The normal lifecycle works:

- A real context-mode MCP child writes `/tmp/context-mode-mcp-ready-<pid>` after connect.
- `isMCPReady()` returns `true`.
- `SIGTERM` removes the sentinel owned by that MCP process.

The namespace case fails:

1. A host process is alive and writes a sentinel containing its host PID.
2. A sandboxed process shares `/tmp` but cannot see that host PID in its PID namespace.
3. `isMCPReady()` reads the sentinel and calls `process.kill(pid, 0)`.
4. The liveness probe fails because the PID is not visible from that namespace.
5. Current code treats the sentinel as stale and unlinks it.
6. The host process is still alive, but the readiness sentinel is gone.

This reproduces the observed state where context-mode MCP processes were still running but `/tmp/context-mode-mcp-ready-*` had disappeared.

## Git archaeology

Relevant history:

```text
04569b1 fix: graceful MCP redirect fallback via readiness sentinel (#230)
99cb1a6 / e56f17c / 35e264c fix(hooks): replace PPID sentinel with directory-scan for MCP readiness (#347)
b46a7af fix(hooks): scope MCP sentinel scan via env override for tests (#355)
```

The existing design protects real invariants:

- #230: do not redirect when MCP is unavailable.
- #347: do not depend on `process.ppid`, because shell wrappers/WSL spawn topology make PPID unreliable.
- #355: isolate tests from real `/tmp` sentinel pollution.

Current tests cover stale sentinel cleanup and child-PID/outside-process-tree cases, but not the combination of shared `/tmp` plus isolated PID namespace.

## Expected behavior

A hook should not delete a live MCP server's sentinel merely because the MCP PID is not visible from the hook process's PID namespace.

The readiness check should preserve both invariants:

- Avoid redirecting to a truly dead MCP server.
- Avoid deleting a live MCP server's sentinel from a PID namespace that cannot prove the process is dead.

## Actual behavior

`hooks/core/mcp-ready.mjs` treats any `process.kill(pid, 0)` failure as stale and unlinks the sentinel immediately.

That is correct for ordinary stale PIDs, but incorrect when `/tmp` is shared across PID namespaces.

## Suggested fix direction

Use a third-way readiness signal that preserves the old invariants:

- Have the MCP server refresh/touch its sentinel periodically while alive.
- Treat a recent sentinel as ready even if `kill(pid, 0)` fails from the current namespace.
- Only unlink/delete stale sentinels after a conservative age threshold.
- Keep explicit cleanup on normal server shutdown.

This would handle PID namespace isolation without regressing the original graceful-fallback behavior from #230.

### LINKED ISSUE #845: Codex PreToolUse formatter drops modify/context despite current updatedInput support
## Summary

The Codex PreToolUse formatter still drops `modify` and `context` decisions by returning `null`, based on an older assumption that Codex rejected `updatedInput` / `additionalContext`.

Current Codex source refs now show support for `permissionDecision: "allow"` with `updatedInput`, including tests and tool handlers that rewrite shell and MCP tool inputs. The context-mode formatter appears stale and can silently pass through commands that should be redirected to context-mode.

## Current context-mode behavior

In `hooks/core/formatters.mjs`, the Codex formatter currently does this:

```js
ask: () => null,
modify: () => null,
context: () => null,
```

That means a normalized routing decision such as a Bash/direct-fetch redirect can be dropped for Codex. The original command then continues instead of being rewritten or blocked with actionable guidance.

## Git archaeology

The current behavior traces back to:

```text
607dc70 fix(codex): add dedicated hook scripts and fix wire protocol (#225)
```

That commit explicitly says the formatter was verified against the then-current `codex-rs/hooks/src/schema.rs` and `output_parser.rs`, and it fixed real wire-protocol issues such as missing `hookEventName`.

Later history updated Codex hook documentation:

```text
b9acf5b fix(codex): update hooks.ts to reflect stable hook dispatch
```

However, `src/adapters/codex/hooks.ts` still documents `updatedInput not yet supported`, and `hooks/core/formatters.mjs` still drops Codex `modify` / `context` decisions.

## Current Codex refs

Current checked-in Codex refs show a different contract:

- `codex-rs/hooks/src/engine/output_parser.rs` parses `hookSpecificOutput.updatedInput`.
- `codex-rs/hooks/src/events/pre_tool_use.rs` has a test named `permission_decision_allow_can_update_input`.
- `codex-rs/core/src/tools/handlers/shell/shell_command.rs` rewrites the shell command from updated hook input.
- `codex-rs/core/src/tools/handlers/unified_exec/exec_command.rs` rewrites `exec_command.cmd` from updated hook input.
- `codex-rs/core/src/tools/handlers/mcp.rs` includes MCP updated-input rewrite coverage.

So the old context-mode assumption may have been correct for #225, but it appears stale for current Codex.

## Expected behavior

Codex formatting should match the current Codex hook contract:

- If Codex supports `allow + updatedInput`, context-mode should emit that shape for safe rewrites.
- If a specific Codex version/path does not support rewriting, direct-fetch/Bash routing should fail closed with a clear deny reason rather than silently passing through.
- `context` handling should also be revisited against the current Codex additional-context support.

## Actual behavior

Context-mode drops Codex `modify` and `context` decisions. For routing decisions whose purpose is to prevent raw direct-fetch output from entering model context, silent pass-through is not aligned with Context Mode's design.

## Validation needed before fixing

Do not rely on build/typecheck only. This needs an actual Codex hook-path validation:

- Confirm the installed Codex version and hook path being tested.
- Verify whether `PreToolUse permissionDecision: "allow" + updatedInput` rewrites a real shell/unified-exec command.
- Verify whether `additionalContext` is recorded or surfaced as expected.
- Preserve the #225 invariant: do not emit a wire shape that older Codex versions reject without a fallback.

### LINKED ISSUE #846: Codex cancels context-mode MCP tool calls because ctx_* descriptors lack annotations
## Summary

Context-mode MCP tools currently register without MCP tool `annotations`. In Codex, this causes read-only context-mode MCP calls such as `ctx_stats` and `ctx_doctor` to be cancelled before execution with `user cancelled MCP tool call`, even though the MCP server itself is reachable and the same tools work through direct MCP protocol probes.

This appears to be a descriptor/approval metadata gap, not a broken MCP server.

## Observed behavior

Validated on Linux with `codex-cli 0.139.0`:

- Direct MCP JSONL probes against context-mode returned the full tool list and `ctx_stats` executed successfully.
- Claude Code could invoke the context-mode MCP tool successfully.
- Codex could invoke other MCP servers in the same environment, including `openaiDeveloperDocs.list_api_endpoints` and `omg.notepad_stats`.
- Codex `exec` attempted `context-mode.ctx_stats` / `ctx_doctor`, but the MCP tool call completed with:

```text
user cancelled MCP tool call
```

A minimal MCP probe reproduced the same pattern: an unannotated test MCP tool was cancelled by Codex, while the same tool with annotations such as `readOnlyHint: true`, `destructiveHint: false`, `idempotentHint: true`, and `openWorldHint: false` was allowed.

## Git archaeology

History check did not find a prior annotations implementation:

```text
git log --all --oneline -S 'readOnlyHint' -- src tests hooks package.json
```

No matching commits were found. This looks like a validation gap rather than a regression that removed existing annotations.

Relevant current code:

- `src/server.ts` wraps `server.registerTool()` and records tool configs, but forwards the config as-is.
- Tool registrations such as `ctx_stats` and `ctx_doctor` provide `title`, `description`, and `inputSchema`, but no `annotations`.
- `tests/core/server.test.ts` has a tool-description style contract that recognizes `annotations` as a possible block delimiter, but does not require annotations or verify read-only/destructive metadata.

## Expected behavior

Context-mode should provide correct MCP tool annotations so MCP clients with approval/permission logic can distinguish read-only diagnostic/search tools from mutating or open-world tools.

At minimum, read-only tools such as `ctx_stats`, `ctx_doctor`, and other truly read-only query/diagnostic tools should carry accurate annotations. Mutating, destructive, process-spawning, or open-world tools should not be incorrectly marked read-only.

## Actual behavior

Context-mode exposes ctx_* tools without annotations. Codex treats unannotated context-mode tools conservatively and cancels calls that should be safe/read-only.

## Suggested fix direction

- Add explicit, accurate MCP annotations to each registered ctx_* tool.
- Add regression coverage that inspects the actual registered tool descriptors, not only descriptions.
- Avoid blanket `readOnlyHint: true`; classify each tool by real behavior.

ken-jo:
## What / Why / How

Fixes #844. Fixes #845. Fixes #846.

Three Codex/MCP bugs that share one theme — fail-open / approval semantics —
shipped as one PR with one commit per issue.

- **#844** — `isMCPReady()` deletes a live MCP server's readiness sentinel when
  the probing hook runs in a different PID namespace that shares `/tmp`
  (`process.kill(pid, 0)` throws `ESRCH` for a host PID it cannot see). Readiness
  then flips false and hooks fail open while the server is still running.
- **#845** — the Codex PreToolUse formatter returns `null` for `modify` and
  `context`, so curl/wget/HTTP/build redirects and guidance nudges are silently
  dropped; the raw command runs and floods the model context. Current Codex
  (codex-cli ≥ 0.141.0) supports `permissionDecision:"allow" + updatedInput`.
- **#846** — `ctx_*` MCP tools register without `annotations`, so Codex cancels
  read-only calls like `ctx_stats` / `ctx_doctor` before execution.

With this PR applied, the readiness sentinel survives cross-namespace probes,
Codex command redirects are rewritten on capable builds and fail closed on older
ones, and every `ctx_*` tool advertises accurate approval annotations.

This PR:

- #844: the reader treats `EPERM` as alive and, on `ESRCH`, trusts a
  recently-refreshed sentinel (90s window) instead of deleting it; only sentinels
  older than the window are cleaned. The server refreshes its sentinel mtime
  every 30s while alive and clears the timer on shutdown.
- #845: detects Codex capability at runtime (`hooks/core/codex-caps.mjs` parses
  `codex --version`, caches with a TTL, fails closed on any error — no opt-in env
  flag). Capable Codex → `allow + updatedInput` (command rewrite) and
  `additionalContext`; older Codex → command redirect becomes an enforceable
  `deny` carrying the same guidance; `ask` stays dropped.
- #846: adds explicit `annotations` to all 11 `ctx_*` tools, classified by real
  behavior (no blanket `readOnlyHint`).
- Tests are added to the existing owning files (no new test files).
- Preserves the #225 invariant (never emit a wire shape older Codex rejects
  without a fallback) and the #230/#347 invariants (no redirect when MCP is down;
  no `process.ppid` dependency).

## Affected platforms

Agent / CLI scope:

- [x] Codex CLI
- [x] All standalone MCP server launchers
- [ ] All platforms

Scope notes:

- #846 (annotations) applies to every MCP client that reads tool annotations
  (Codex is the one that cancels without them; others simply gain accurate hints).
- #844 (readiness sentinel) applies to every adapter whose hooks consult
  `isMCPReady()` — it hardens the shared gate, no per-adapter behavior change.
- #845 is Codex-only (the Codex PreToolUse formatter path). The Codex TS adapter
  `formatPreToolUseResponse` is a separate in-process layer not on the external
  hook path and is intentionally unchanged.

OS scope:

- [x] Linux
- [x] macOS
- [x] Windows
- [x] All operating systems

## Root Cause

Git archaeology:

- #844: `04569b1` (#230) added the graceful MCP-redirect fallback via the
  readiness sentinel; `#347` replaced the PPID lookup with a directory scan;
  `#355` added the `CONTEXT_MODE_MCP_SENTINEL_DIR` test override. The reader
  unlinked the sentinel on any `kill(pid, 0)` failure — correct for a dead PID,
  wrong when `/tmp` is shared across isolated PID namespaces (a live host PID is
  invisible → `ESRCH` → live sentinel deleted).
- #845: `607dc70` (#225) made the Codex formatter deny-only because Codex then
  rejected `updatedInput`. That assumption is now stale: codex-cli ≥ 0.141.0
  parses `permissionDecision:"allow" + updatedInput` and `additionalContext`
  (verified against the shipped 0.141.0 binary's `output_parser`).
- #846: `ctx_*` tools were registered without `annotations` since inception;
  Codex's approval logic cancels unannotated tool calls.

- #225: never emit a wire shape an older Codex rejects without a fallback — here,
  fail closed to `deny` when `updatedInput` is unsupported.
- #230 / #347: no redirect when MCP is unavailable; no `process.ppid` dependency.

## Validation

| --- | --- | --- |
| Linux local (codex 0.141.0) | `tsc --noEmit`; `npm run build`; `vitest` targeted + hooks/server sweep | passed (tsc 0, build green, 712 targeted, 1091 sweep / 1 skipped) |
| macOS via `ssh my-mac` (codex 0.141.0, Node 26) | `vitest run tests/hooks/formatters.test.ts tests/hooks/core-routing.test.ts` + live hook smoke | passed (143/143); hook → `allow + updatedInput` |
| Windows via `ssh my-window` (codex 0.139.0, Node 24) | same suites + live hook smoke | passed (143/143); hook → `deny` (fail closed) via `cmd.exe` caps path |

Behavioral evidence (#845):

- BEFORE: `Bash(curl -s https://example.com)` → formatter returns `null` → raw
  curl runs, HTTP body enters the model context.
- AFTER (codex 0.141.0): the hook emits `permissionDecision:"allow"` +
  `updatedInput` and Codex executes the rewritten command —
  `zsh -lc 'echo "context-mode: curl/wget redirected…"'` — not curl; the model
  receives the guidance, not the page body.
- AFTER (codex 0.139.0, below floor): the hook emits `permissionDecision:"deny"`
  with the guidance reason — no silent passthrough.

Agent-working smoke:

- codex-cli 0.141.0 (Linux), real `codex exec`, sandbox read-only: model called
  `Bash(curl …)`; PreToolUse hook fired (`tool_name` `Bash`, `isMCPReady` true);
  Codex executed the rewritten `echo` guidance. Confirms Codex honors
  `allow + updatedInput` end-to-end.
- codex-cli 0.141.0 (macOS) and 0.139.0 (Windows): live hook invocation against
  the installed binary → `allow + updatedInput` and `deny` respectively.

Claude Code regression (shared-core guard — #844 `isMCPReady`, #846 server
annotations, and the shared `formatDecision` signature all affect Claude Code):

- MCP surface (built server, stdio JSON-RPC as Claude Code speaks it): `initialize`
  OK; `tools/list` returns 11 ctx_* tools, all 11 carrying annotations
  (`ctx_stats` → `readOnlyHint:true`, `ctx_purge` → `destructiveHint:true`);
  `ctx_stats` call executes; readiness sentinel written on connect.
- Hook surface (`hooks/pretooluse.mjs`, platform `claude-code`): `Bash(curl …)`
  with MCP ready → `deny` + guidance (redirect fires); with no live sentinel →
  passthrough (no redirect) — the #230 readiness invariant is preserved, normal
  same-namespace path unchanged.

Cross-adapter hook smoke (shared core: `isMCPReady` #844 + `formatDecision`
#845 — a curl redirect run through every PreToolUse hook entrypoint, MCP ready):
claude-code `deny`; codex(0.141) `allow+updatedInput`; cursor `updated_input`;
gemini-cli `hookSpecificOutput.tool_input`; vscode/jetbrains `allow+updatedInput`;
copilot-cli `modifiedArgs`; kimi passthrough (deny-only host); kiro `exit 2`
block; antigravity-cli `decision:deny`. opencode/kilo (in-process plugin) covered
by the MCP `tools/list` registry probe. No adapter regressed.

- #844: fresh-but-unprobeable sentinel preserved; stale dead-PID sentinel cleaned;
  both in a single scan.
- #845: non-command `modify` (e.g. Agent prompt injection) dropped rather than
  blocking the tool; `ask` stays dropped; capability cache TTL re-probe.
- #846: regression test inspects the actual registered descriptors via
  `REGISTERED_CTX_TOOLS`, asserting read-only vs mutating classification.

Broad suite:

- `vitest run tests/hooks tests/core/server.test.ts` → 1091 passed, 1 skipped.

#845 touches only the Codex external-hook formatter path; other adapters keep
their own `modify`/`context` formatters and are unchanged (verified: other
formatter suites green). #844 and #846 are platform-agnostic — #844 hardens the
shared readiness gate (other adapters unaffected in behavior), #846 adds standard
MCP annotations all clients can read. Validated on Linux, macOS, and Windows
against real codex binaries (0.141.0 and 0.139.0).

## Checklist

- [x] Regression coverage integrated into existing relevant test files (no new
      test files); only new source module is `hooks/core/codex-caps.mjs`
- [x] `npm run build` passes
- [x] `npm run typecheck` passes (`tsc --noEmit`, exit 0)
- [x] Targeted tests pass (712) and broad sweep (1091)
- [x] Generated bundles not included — CI refreshes them
- [x] Docs/comments updated where behavior changed (`src/adapters/codex/hooks.ts`)
- [x] 3OS validation completed (Linux/macOS/Windows, real codex 0.141.0 + 0.139.0)
- [x] Agent-working smoke completed (`codex exec` honored the rewrite end-to-end)
- [x] Targets `next` branch


ken-jo:
## Cross-CLI / cross-adapter validation

Installed agent CLIs on the validation host: **codex 0.141.0, Claude Code, gemini, opencode, kilo/kilocode, copilot**. I verified the shared-core changes — #844 `isMCPReady`, #845 the `formatDecision` signature, #846 server tool annotations — do not regress any adapter.

### MCP server surface (built `build/server.js`, stdio JSON-RPC)

- `initialize` OK; `tools/list` returns all 11 `ctx_*` tools, each carrying `annotations` (`ctx_stats` → `readOnlyHint:true`, `ctx_purge` → `destructiveHint:true`); a `ctx_stats` call executes; the readiness sentinel is written on connect.
- This is the surface every MCP client consumes, including the OpenCode/Kilo in-process plugin registry — so #846 is covered for all of them.

### PreToolUse hook surface — a curl redirect run through every entrypoint (MCP ready)

| Adapter | Result |
| --- | --- |
| Claude Code | `deny` + guidance |
| Codex (0.141.0) | `allow + updatedInput` (rewrite) |
| Cursor | `updated_input` |
| Gemini CLI | `hookSpecificOutput.tool_input` |
| VS Code Copilot | `allow + updatedInput` |
| JetBrains Copilot | `allow + updatedInput` |
| Copilot CLI | `modifiedArgs` |
| Kimi | passthrough (deny-only host) |
| Kiro | `exit 2` (block — documented exit-code contract) |
| Antigravity CLI | `decision: "deny"` |

Each produced its correct platform-specific shape, exercising the shared `isMCPReady` and `formatDecision` paths. No adapter regressed.

### 3OS gate (real Codex binaries)

- **Linux** (codex 0.141.0): full suite + `tsc` + build green; a real `codex exec` session honored `allow + updatedInput` end-to-end — Codex executed the rewritten `echo` guidance, not the original curl.
- **macOS** via `ssh my-mac` (codex 0.141.0, Node 26): 143 targeted tests; live hook → `allow + updatedInput`.
- **Windows** via `ssh my-window` (codex 0.139.0, Node 24): 143 targeted tests; live hook → `deny` (fail closed) through the `cmd.exe` capability-detection path — exercises the older-build path on a real 0.139.0 binary.

### Conclusion

The shared-core changes are contained and verified across all installed CLIs and every hook adapter on Linux/macOS/Windows, with both #845 paths (rewrite on ≥0.141.0, fail-closed deny on 0.139.0) confirmed against real Codex binaries. Marking ready for review.

Issue #643, opened 2026-05-19
Make Pi MCP bridge `tools/call` timeout configurable via env var


jetnet:
## Problem
>
The Pi MCP bridge enforces a 120-second timeout on every `tools/call` round-trip:
>
```ts
// src/adapters/pi/mcp-bridge.ts:128
const DEFAULT_CALL_TIMEOUT_MS = 120_000;
```
>
The value is hardcoded — no env var, no config-file plumbing, no `bootstrapMCPTools` parameter. `callTool` always passes the constant through to `request()` (line 507). The executor itself has no default timeout when `timeout` is omitted (intentional), but the bridge's bound is independent and fires first.

In practice this means any `ctx_execute` / `ctx_execute_file` / `ctx_batch_execute` call that takes longer than 120 s fails at the bridge layer with `MCP request timeout after 120000ms: tools/call`, even though the subprocess in the child would have finished. Users hit this on test suites (2–5 min), `npm run build` on larger TS repos, `cargo test`, `go test ./...`, and single long-running fetches the bridge can't proactively chunk.

The existing escape hatch — `ctx_execute({ background: true, ... })` — works, but requires the model to know to set the flag, and the partial-output return changes the calling shape.

## Proposed solution

```ts
function resolveCallTimeoutMs(): number {
  const raw = process.env.CTX_BRIDGE_CALL_TIMEOUT_MS;
  if (!raw) return 120_000;
  const n = Number.parseInt(raw, 10);
  return Number.isFinite(n) && n > 0 ? n : 120_000;
}
const DEFAULT_CALL_TIMEOUT_MS = resolveCallTimeoutMs();
```

Naming follows the existing convention (`CTX_FETCH_STRICT`, `CONTEXT_MODE_*`). `CTX_BRIDGE_CALL_TIMEOUT_MS` is specific about which timeout — there's also `DEFAULT_REQUEST_TIMEOUT_MS = 60_000` for `initialize` / `tools/list`, and users probably shouldn't be tweaking that one.

## Alternatives considered

1. **Per-call timeout hint in `ctx_execute`.** More invasive; doesn't help when the model doesn't set it.
2. **Drop the bridge timeout entirely.** Comment at `mcp-bridge.ts:126–127` explicitly justifies keeping a bound. The bound is correct; only the value should be tunable.
3. **Status quo + `background: true`.** Works when partial output is acceptable; doesn't help users whose calling pattern needs the full result synchronously.

Happy to send a PR if the direction is acceptable.

[...]

PR #855, opened 2026-06-21
fix(lifecycle): reap idle MCP bridge children to stop pi/omp accumulation (#854)

### LINKED ISSUE #854: Pi host: MCP servers accumulate under a long-lived parent — self-shutdown gated entirely on parent-death (same class as #565, still present in 1.0.163)
# Pi host: MCP servers accumulate under a long-lived parent — self-shutdown is gated entirely on parent-process death (same class as #565, still present in 1.0.163)

## Summary

On hosts that spawn **one context-mode MCP server per sub-context/subagent under a single long-lived parent process**, the servers never self-terminate when their MCP client disconnects. They accumulate for the entire lifetime of the parent. With the `pi` coding agent as host, over a multi-day session I observed **188 `server.bundle.mjs` processes (~8.8 GB RSS)** under a handful of live parent PIDs.

This is the **same class as #565** ("OpenCode accumulates one context-mode MCP child per session — never reaped", closed COMPLETED 2026-05-14) — but it is **still reproducible in 1.0.163 with the pi host**, so whatever fix landed for #565 does not cover this path. It is also the "unresolved SDK stdin end-handler issue" referenced in #311 / #388.

It is **distinct from** the orphan-CPU bug fixed by #833 (those were *orphaned* servers; that fix makes a fatal-exception storm call `process.exit`). The servers here are **not orphaned** — their parent is alive — so #833 does not address them.

## Root cause

In the shutdown watchdog (`server.bundle.mjs`, deminified):

```js
function y1(t = {}) {                       // builds isParentAlive
  let e = t.getPpid ?? (() => process.ppid),
      n = t.readGrandparentPpid ?? g1,
      r = e(),                              // initial ppid
      o = n();                             // initial grandparent ppid
  return () => {
    let s = e();
    return !(s !== r || s === 0 || s === 1 || (!Number.isNaN(o) && o > 1 && n() === 1));
  };
}
var _1 = y1();                              // isParentAlive

function Rw(t) {                            // watchdog install
  let e = t.checkIntervalMs ?? S1(),
      n = t.isParentAlive ?? _1,
      r = false,
      o = () => { r || (r = true, t.onShutdown()); },   // shutdown
      s = setInterval(() => { n() || o(); }, e);         // (1) interval: parent-death only
  s.unref();
  let i = ["SIGTERM", "SIGINT"];
  process.platform !== "win32" && i.push("SIGHUP");
  for (let c of i) process.on(c, o);                     // (2) signals
  let a = () => { n() || o(); };                          // (3) stdin "end": ALSO gated on parent-death
  process.stdin.isTTY || process.stdin.on("end", a);
  ...
}
```

Every self-shutdown path is conditioned on the **parent process dying**:

1. The interval check calls `o()` only when `isParentAlive()` is false.
2. Signals (SIGTERM/SIGINT/SIGHUP) — requires the host to explicitly signal the child.
3. The **stdin `"end"` handler** — which is exactly the "my MCP client disconnected" signal — is `n() || o()`, so it shuts down **only if the parent is also dead**.

As a result there is **no path** by which a server exits while its spawning parent is still alive, even though its MCP client (the sub-context) is gone. If the host keeps one long-lived parent process and does not explicitly SIGTERM finished per-subagent servers, they leak for the parent's whole lifetime.

## Evidence

Installed version: **1.0.163** (latest; includes #833).

Leaked servers under a single live `pi` parent, idle, with stdin (fd 0) still an open unix socket:

```
$ for p in $(pgrep -f context-mode/server.bundle.mjs | head -3); do
    echo "pid $p age $(ps -o etime= -p $p) fd0=$(lsof -p $p -a -d 0 | tail -1 | awk '{print $5}')"
  done
pid 6517 age 41:51 fd0=unix
pid 7234 age 41:50 fd0=unix
pid 20567 age 41:03 fd0=unix
```

Grouped by parent PID, a single multi-day `pi` session had 13 servers parented to one PID; across the machine 188 total, ~8.8 GB, before manual cleanup.

## Suggested fix

Treat **MCP transport disconnect as an unconditional shutdown signal**, independent of parent liveness. The stdin-`"end"` (and transport `onclose`) handler should call `o()` directly:

```js
let a = () => { o(); };                 // was: () => { n() || o() }
process.stdin.isTTY || process.stdin.on("end", a);
```

Rationale: stdin EOF / transport close on a stdio MCP server unambiguously means the client is gone; the server has no reason to keep running regardless of whether the OS parent happens to still be alive. The `isParentAlive` interval can stay as a secondary safety net for true orphan cases.

(If the concern is hosts that reuse a transport, gating on `onclose` of the actual MCP transport rather than raw stdin would be even safer.)

## Workaround (for affected users today)

An external idle reaper that kills `server.bundle.mjs` processes older than N minutes with <2% CPU, keeping the newest few (they respawn on demand). This fully contains the leak but is obviously a band-aid.

## Related

- #565 — same accumulation class on OpenCode (closed COMPLETED, but this pi path still leaks in 1.0.163)
- #311 / #388 — the SDK stdin end-handler issue this pins down concretely
- #833 — orphan fatal-exception exit (merged, present in 1.0.163; does not cover live-parent accumulation)

## Environment

- context-mode 1.0.163 (latest)
- Host: `pi` coding agent (spawns one MCP server per sub-context under a long-lived parent)
- macOS, Node (host) / better-sqlite3 path

ken-jo:
## What / Why / How

Fixes #854.

On hosts that spawn **one MCP server per sub-context/subagent under a single long-lived parent** (pi / oh-my-pi `omp`), context-mode servers never self-terminate and accumulate for the parent's whole lifetime (reporter: 188 procs / ~8.8 GB over a multi-day session).

With this PR applied, an MCP **bridge child** that receives no MCP activity for an idle window self-terminates, so accumulation is bounded; the pi extension's single-flight path respawns one transparently on the next tool call.

This PR:

- **`src/lifecycle.ts`** — a request-idle self-shutdown timer, installed **only for MCP bridge children** (`CONTEXT_MODE_BRIDGE_DEPTH>0`). It shuts the child down after no MCP activity for `bridgeChildIdleTimeoutMs()` (default 3 min; override `CONTEXT_MODE_BRIDGE_IDLE_MS`, non-positive disables) **and only while no tool call is in flight** — a long single `ctx_execute`/`ctx_batch_execute` (one inbound frame, then unbounded #643) is never reaped mid-flight. New exports `bridgeChildIdleTimeoutMs()`, `noteMcpActivity()`, `noteRequestStart()/noteRequestEnd()`, `attachMcpActivityTap()`.
- **`src/server.ts`** — `attachMcpActivityTap(transport)` refreshes the idle clock on each inbound message; `wrapToolHandler` brackets every tool call with `noteRequestStart/End` (try/finally, success+error); `trackResponse` also bumps activity. No stdin touch.
- **tests** — `bridgeChildIdleTimeoutMs` table tests; idle-reaper behavior (reaps when idle, not while active, **not while a call is in flight**, depth-0 no-reap guard for #602); `attachMcpActivityTap` delegation.
- preserved invariants: depth-0/absent installs no reaper (#602); trigger is idle time via the MCP message layer, never stdin EOF (#236); in-flight guard (#643).

> Note: a complementary change to dedupe the per-sub-context spawn in the pi extension was prototyped but **reverted** — module-level bridge state can be stale (point at a dead child), so reusing it is unsafe. The depth-scoped idle reaper bounds the leak robustly on its own and is the focused fix here.

## Affected platforms

Agent / CLI scope:

- [x] OpenClaw / Pi  *(pi / oh-my-pi `omp` — the bridge-child path; this is where servers accumulate)*
- [ ] All other agents — `lifecycle.ts`/`server.ts` are shared, but the reaper is gated to `CONTEXT_MODE_BRIDGE_DEPTH>0` (only the pi MCP bridge sets that), so depth-0 servers on every other host are behavior-unchanged.

OS scope:

- [x] Linux *(build/typecheck/tests; mechanism verified)*
- [x] macOS *(leak reproduced on real omp+Claude; patched-server idle-reap verified end-to-end)*
- [ ] Windows *(OS-agnostic change — no platform-specific code path; not separately run)*

Scope notes:

- The fix only ever runs in a process launched as a context-mode MCP **bridge child** (`CONTEXT_MODE_BRIDGE_DEPTH>0`). Standalone MCP servers and all non-pi adapters run at depth 0 and never install the idle reaper.

## Root Cause

Git / changelog archaeology:

- The pi extension (`src/adapters/pi/extension.ts`) starts a bridge child (`node …/server.bundle.mjs`, `BRIDGE_DEPTH=1`) from `before_agent_start` — which fires per sub-context/subagent — but tears it down only at `session_shutdown`, which never fires for sub-contexts while the parent process lives.
- `src/lifecycle.ts` reaps only on **parent-process death** (ppid poll + signals + a stdin-EOF assist itself gated on parent-death). The stdin-EOF gating is deliberate — `ac11f10`/#236 removed unconditional stdin-EOF shutdown because transient pipe events caused spurious `-32000` shutdowns. So with the `omp` parent alive, none of the three triggers fire.
- Same class as #565 (OpenCode accumulation); its generic idle-shutdown (`ac11f10`) was narrowed (`5bcff7e`/#592) then fully reverted (`a888cac`/#602, "restore tools on 12 hosts"), so this path has had no defense.

Preserved invariants: #236 (no stdin-EOF shutdown), #311/#388 (orphan fast-detect), #602 (depth-0 keep-alive servers not reaped on idle), #534 (bridge-child 1 s poll).

## Validation

| --- | --- | --- |
| Linux local | `npm run build` | passed (assert-bundle + assert-asymmetric-drift OK) |
| Linux local | `npx tsc --noEmit` | passed |
| Linux local | `npx vitest run tests/lifecycle.test.ts` | 23 passed (incl. 9 new #854 + the #236/#311/#388 guards) |
| Linux local | `npx vitest run tests/adapters/pi-mcp-bridge.test.ts tests/adapters/pi-bridge-parent-death.test.ts tests/pi-extension.test.ts` | 99 passed |
| macOS (`ssh my-mac`) | real omp 16.1.11 + Claude, extension loaded, 6-subagent turn | **leak reproduced pre-fix**: 13 `server.bundle.mjs` (depth=1) accumulated under one `omp` parent, none reaped |
| macOS (`ssh my-mac`) | **patched** vs original `server.bundle.mjs` as a depth=1 bridge child under a live parent, `CONTEXT_MODE_BRIDGE_IDLE_MS=3000`, idle | original `ALIVE after 6000ms` (leak); **patched `EXITED at ~3297ms` (idle-reaped)** ✅ |

Behavioral evidence:

- BEFORE: depth-1 bridge children with a live parent persist indefinitely (timer/signals/stdin-EOF all gated on parent-death; parent alive ⇒ never reap). Confirmed on macOS: the original `server.bundle.mjs` stayed `ALIVE after 6000ms`.
- AFTER: a depth>0 child with no activity for the idle window self-exits (macOS: patched `server.bundle.mjs` `EXITED at ~3297ms` ≈ the 3000 ms window). With activity, or while a tool call is in flight, it does not; depth-0 never installs the reaper.

Agent-working smoke:

- omp 16.1.11 + Claude (macOS) reproduced the leak shape (13 depth=1 `server.bundle.mjs` under one omp parent); the patched binary self-reaps an abandoned/idle child while the parent stays alive, so accumulation is bounded. The pi single-flight path (`mcp-bridge.ts`) respawns one transparently on the next tool call.

- in-flight guard: a long single tool call (one inbound frame, then unbounded) is not reaped mid-flight (#643).
- depth-0 / absent: the reaper is never installed (#602 keep-alive hosts untouched).
- no stdin listener added; idle trigger is the MCP message layer, not stdin EOF (#236).

`lifecycle.ts` and `server.ts` are shared by every adapter's MCP server, but the new behavior is gated to `CONTEXT_MODE_BRIDGE_DEPTH>0` (only the pi bridge), and the `onmessage` wrap is a no-op beyond bumping a timestamp. Depth-0 servers on Claude Code, Codex, Cursor, Gemini, Copilot, Qwen/Kimi, Kiro, Antigravity, Zed, and standalone launchers are unchanged. OS-agnostic (no platform-specific code).

## Checklist

- [x] Tests added with a real behavioral regression guard (idle-reaper behavior + depth-0 #602 guard + #236 stdin invariant stays green)
- [x] Regression coverage integrated into the existing `tests/lifecycle.test.ts`
- [x] `npm run build` passes
- [x] Targeted tests pass (lifecycle 23, pi 99)
- [x] Generated bundles updated when source changes require it — **CI (`bundle.yml`) regenerates + commits**; local bundle diffs not included
- [x] Docs: behavior is internal (no user-facing doc change); env knob `CONTEXT_MODE_BRIDGE_IDLE_MS` documented in code
- [x] 3OS validation — Linux (build/typecheck/tests) + macOS (real-binary before/after) done; Windows is an OS-agnostic path (no platform-specific code), not separately run
- [x] Agent-working smoke — macOS patched-server before/after (original persists, patched idle-reaps ~3.3 s) + omp 16.1.11 leak-shape reproduction
- [x] Targets `next`

### DIFF
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -11,6 +11,12 @@
  * next poll tick), which closes the multi-day CPU-spin window seen in
  * #311/#388 without reintroducing the false-positive shutdowns of #236.
  *
+ * Additionally, for MCP BRIDGE CHILDREN only (CONTEXT_MODE_BRIDGE_DEPTH>0), a
+ * request-idle self-shutdown reaps a child that a pi/omp sub-context abandoned
+ * while its long-lived parent keeps running (#854) — gated so the depth-0
+ * keep-alive servers #602 restored are never reaped, never via stdin EOF, and
+ * never while a tool call is in flight (#643).
+ *
  * Cross-platform: macOS, Linux, Windows.
  */
 
@@ -23,6 +29,12 @@ export interface LifecycleGuardOptions {
   onShutdown: () => void;
   /** Injectable parent-alive check (for testing). Default: ppid-based check. */
   isParentAlive?: () => boolean;
+  /**
+   * #854: request-idle shutdown timeout (ms) for MCP bridge children. Default:
+   * {@link bridgeChildIdleTimeoutMs}() — 0 (disabled) unless CONTEXT_MODE_BRIDGE_DEPTH>0.
+   * Exposed for testing.
+   */
+  bridgeIdleMs?: number;
 }
 
 /** Read grandparent PID via `ps -o ppid= -p $PPID`. Returns NaN on failure or Windows. */
@@ -115,6 +127,81 @@ export function lifecycleGuardIntervalForEnv(
 }
 
 /**
+ * #854: idle-shutdown timeout (ms) for an MCP BRIDGE CHILD. Returns 0 (disabled)
+ * unless this process is a bridge child (CONTEXT_MODE_BRIDGE_DEPTH>0). depth-0 /
+ * absent always returns 0, so the long-lived keep-alive servers that #602
+ * restored are NEVER reaped on idle. Default for bridge children is 3 min;
+ * override with CONTEXT_MODE_BRIDGE_IDLE_MS (a non-positive value disables it).
+ * The reaper additionally never fires while a tool call is in flight (see
+ * {@link noteRequestStart}), so the window only bounds how fast *abandoned*
+ * children drain — it does not cap legitimate long-running calls.
+ *
+ * Exported for unit-testing.
+ */
+export function bridgeChildIdleTimeoutMs(env: NodeJS.ProcessEnv = process.env): number {
+  const depth = Number.parseInt(env.CONTEXT_MODE_BRIDGE_DEPTH ?? "", 10);
+  if (!Number.isFinite(depth) || depth <= 0) return 0;
+  const raw = env.CONTEXT_MODE_BRIDGE_IDLE_MS;
+  if (raw !== undefined) {
+    const v = Number.parseInt(raw, 10);
+    return Number.isFinite(v) && v > 0 ? v : 0;
+  }
+  return 180_000;
+}
+
+// #854 idle-reaper state, module-level by design: an MCP server is exactly one
+// process (one StdioServerTransport + one lifecycle guard), so these are never
+// shared across concurrent servers in production. Multiple startLifecycleGuard()
+// instances arise only in tests, which pair/reset these explicitly.
+/** Last MCP activity timestamp (inbound message, tool-call start/end, or response). */
+let _lastMcpActivity = Date.now();
+/** In-flight tool-call count — the reaper never fires while this is > 0. */
+let _inFlight = 0;
+
+/**
+ * #854: record MCP activity (inbound message or response). The server calls this
+ * so the bridge-child idle reaper in {@link startLifecycleGuard} can distinguish
+ * an actively-used child from an abandoned one. Cheap; safe on the hot path.
+ */
+export function noteMcpActivity(): void {
+  _lastMcpActivity = Date.now();
+}
+
+/**
+ * #854: mark a tool call as started. Suppresses the bridge-child idle reaper so a
+ * single long-running ctx_execute / ctx_batch_execute (which sends one inbound
+ * frame then runs unbounded, #643) is never reaped mid-execution.
+ */
+export function noteRequestStart(): void {
+  _inFlight++;
+  _lastMcpActivity = Date.now();
+}
+
+/** #854: mark a tool call as finished (success or error). */
+export function noteRequestEnd(): void {
+  if (_inFlight > 0) _inFlight--;
+  _lastMcpActivity = Date.now();
+}
+
+/**
+ * #854: wrap an MCP stdio transport's `onmessage` so each inbound message
+ * refreshes the idle clock. Best-effort: call after `connect()` (onmessage set);
+ * a no-op if it isn't a function, and a throw in noteMcpActivity never breaks
+ * dispatch. No stdin touch (preserves the #236 contract). Exported for testing.
+ */
+export function attachMcpActivityTap(
+  transport: { onmessage?: (message: unknown, extra?: unknown) => unknown } | null | undefined,
+): void {
+  if (!transport) return;
+  const prev = typeof transport.onmessage === "function" ? transport.onmessage.bind(transport) : null;
+  if (!prev) return;
+  transport.onmessage = (message: unknown, extra?: unknown) => {
+    try { noteMcpActivity(); } catch { /* never break message dispatch */ }
+    return prev(message, extra);
+  };
+}
+
+/**
  * Start the lifecycle guard. Returns a cleanup function.
  * Skipped automatically when stdin is a TTY (e.g. OpenCode ts-plugin).
  */
@@ -162,9 +249,42 @@ export function startLifecycleGuard(opts: LifecycleGuardOptions): () => void {
     process.stdin.on("end", onStdinEnd);
   }
 
+  // #854: request-idle self-shutdown for MCP BRIDGE CHILDREN only
+  // (CONTEXT_MODE_BRIDGE_DEPTH>0). Pi/omp loads the extension once per
+  // sub-context and spawns one bridge child each, tearing them down only at
+  // session_shutdown — which never fires for sub-contexts while the long-lived
+  // parent stays alive, so idle children accumulate (#854, same class as #565).
+  // A bridge child that receives no inbound MCP message for `idleMs` exits
+  // itself; the extension's single-flight path respawns one on the next call.
+  //
+  // Scoped strictly to depth>0 so the depth-0 keep-alive servers that #602
+  // restored are never reaped on idle. The trigger is idle TIME via
+  // noteMcpActivity() (NOT stdin EOF), so the #236 contract — and lifecycle's
+  // hands-off-stdin invariant — are untouched.
+  const idleMs = opts.bridgeIdleMs ?? bridgeChildIdleTimeoutMs();
+  let idleTimer: ReturnType<typeof setInterval> | undefined;
+  if (idleMs > 0) {
+    _lastMcpActivity = Date.now();
+    idleTimer = setInterval(() => {
+      // Reap only when truly quiescent: NO tool call in flight AND no MCP
+      // activity for `idleMs`. The in-flight guard prevents reaping a child
+      // mid-execution during a long single ctx_execute/batch that sends no
+      // further messages (#643 unbounded calls) — the false-reap regression the
+      // adversarial review flagged.
+      if (_inFlight === 0 && Date.now() - _lastMcpActivity >= idleMs) {
+        process.stderr.write(
+          `[context-mode] idle MCP bridge child self-shutdown after ${idleMs}ms with no activity (#854)\n`,
+        );
+        shutdown();
+      }
+    }, Math.max(1000, Math.min(Math.floor(idleMs / 4), 30_000)));
+    idleTimer.unref();
+  }
+
   return () => {
     stopped = true;
     clearInterval(timer);
+    if (idleTimer) clearInterval(idleTimer);
     for (const sig of signals) process.removeListener(sig, shutdown);
     process.stdin.removeListener("end", onStdinEnd);
   };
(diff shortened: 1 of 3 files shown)


ken-jo:
## End-to-end validation (macOS, real binary)

Ran the exact leak shape — an MCP **bridge child** (`CONTEXT_MODE_BRIDGE_DEPTH=1`) under a **live parent**, idle — against the original vs the patched `server.bundle.mjs`, with `CONTEXT_MODE_BRIDGE_IDLE_MS=3000`. The probe parent stays alive the whole time, so this isolates the new idle-reaper from the pre-existing parent-death path.

| Bundle | Result |
| --- | --- |
| context-mode 1.0.163 (no reaper) | `ALIVE after 6000ms` — NOT reaped (the #854 leak) |
| this PR (`IDLE_MS=3000`) | `EXITED at ~3297ms` — idle-reaped ✅ |

Combined with the earlier reproduction (omp 16.1.11 + Claude → 13 depth-1 `server.bundle.mjs` accumulating under one omp parent), this confirms: omp creates the accumulating bridge-child shape, and the patched binary self-reaps an abandoned/idle child while the parent stays alive — so accumulation is now bounded. The respawn-on-next-call path (`mcp-bridge.ts` single-flight) makes the reap transparent.

Env: macOS 26.2 arm64, Node 24, context-mode @ this branch. The global install was restored to 1.0.163 after the test (verified by `cmp`).

ken-jo:
## Cross-agent validation (core `lifecycle.ts`/`server.ts` is shared)

Since this touches code every adapter's MCP server runs, I verified the **depth-0** path (Claude Code, Codex, Cursor, Gemini, and every non-pi host) is unaffected on the **patched** `server.bundle.mjs`:

| Check | Result |
| --- | --- |
| Lifecycle gating — depth-0, idle, **`CONTEXT_MODE_BRIDGE_IDLE_MS=3000` set** | `ALIVE after 6s` — reaper not installed ✓ (#602 preserved) |
| Lifecycle gating — depth-1 (pi bridge), same env | `EXITED ~3s` — reaped ✓ |
| Patched depth-0 MCP path | `initialize` OK · `tools/list` = 11 tools · `tools/call ctx_stats` OK ✓ |
| **Claude Code** (real agent) | live `ctx_stats` MCP call returned ✓ |
| **Codex** (real `codex exec`, patched server wired as an MCP server) | `mcp: cmpatch/ctx_stats started → completed` → tool returned ✓ |

Codex/Claude Code spawn context-mode at **depth-0** (no `CONTEXT_MODE_BRIDGE_DEPTH`), so `bridgeChildIdleTimeoutMs()` returns 0 and the idle timer is never installed — their long-lived servers are never idle-reaped, and tool calls work unchanged through the modified `wrapToolHandler` / `trackResponse` / `attachMcpActivityTap` path. Only pi bridge children (depth>0) get the reaper.

riddlex-beep:
Confirmed fixed on my pi/omp setup — thanks for the quick turnaround.

I built `next`@`77e22a4` from source and validated it before and after installing the patched bundle into a live install (`context-mode` 1.0.163 base, macOS, Node 26).

**Build + unit**
- `tsc --noEmit` clean; `npm run bundle` produced a `server.bundle.mjs` carrying the new `CONTEXT_MODE_BRIDGE_IDLE_MS` path (absent in 1.0.163).
- `tests/lifecycle.test.ts`: 23/23 pass, including the idle bridge-child self-shutdown case.

**Repro + fix** (isolated harness: 6 concurrent depth-1 children, `CONTEXT_MODE_BRIDGE_IDLE_MS=3000`, stdin kept open)

| Bundle | Behavior |
|---|---|
| 1.0.163 | spawns 6, stays at 6 for the whole window — never reaped (accumulation) |
| patched | spawns 6, peak 6, all reaped ~3s after going idle → 0 |

**Over-reap safety** (patched)
- depth-0 keep-alive (#602): never reaped — stayed alive well past the idle window.
- busy depth-1 child: not reaped mid-work — steady inbound traffic kept it alive; it reaped only ~3s after traffic stopped.

**Live pi/omp**
Before the patch I had 9 orphaned `server.bundle.mjs` processes aged 3.5–5.5h that survived a pi restart (the accumulation this issue describes). After installing the patched bundle and restarting: idle bridge children self-reap, the live count returns to 0 when no `ctx_*` tool is active, and the next call transparently single-flight respawns one.

No accumulation over a long session, and no regressions on the keep-alive or in-flight paths. Behaves exactly as described — gated to `BRIDGE_DEPTH>0`, idle-time trigger, never via stdin EOF, never mid-call.

Issue #861, opened 2026-06-22
Windows: boot-time turndown/better-sqlite3 npm install fails EPERM (C:\Windows\node_modules) every session — shell:true drops cwd


lravizzoni:
## Summary

On Windows, the boot-time background install of the three optional fetch dependencies (`turndown`, `turndown-plugin-gfm`, `@mixmark-io/domino`) in `start.mjs` (~L444-465) fails on **every** MCP boot with `EPERM: mkdir 'C:\Windows\node_modules'`, and flashes a `cmd.exe` console window each time. Because the install never persists, the `existsSync(node_modules/<pkg>)` guard never short-circuits, so it re-fires indefinitely — once per Claude Code session.

Observed on `context-mode@1.0.162`, Windows 11, Node v25.8.1, npm 11.11.1 (Claude Code plugin cache install).

## Evidence

npm debug logs (`%LOCALAPPDATA%\npm-cache\_logs`) for the boot-time installs, all clustered within seconds of each other at session start:

```
argv "install" "turndown" "--no-package-lock" "--no-save" "--loglevel" "silent" "--no-audit" "--no-fund"
verbose stack Error: EPERM: operation not permitted, mkdir 'C:\Windows\node_modules'
error code EPERM
error syscall mkdir
```

Identical EPERM for `turndown-plugin-gfm`, `@mixmark-io/domino`, and `better-sqlite3`.

## Root cause

The spawn at ~L451 uses `{ cwd: __dirname, shell: IS_WIN32, detached: true }`. **On Windows + Node, `shell: true` drops the `cwd` option** — the spawned `cmd.exe` does not adopt `__dirname` and instead inherits the parent MCP-server process's working directory, which under Claude Code on Windows is `C:\Windows`. `npm install` then tries to create `C:\Windows\node_modules`, which an unprivileged process cannot write → `EPERM`, instantly, every time.

Minimal reproduction (filesystem evidence, isolates the option):

```js
const { spawn } = require("node:child_process");
const fs = require("fs"), os = require("os"), path = require("path");
for (const opts of [{ shell: true }, { shell: false }]) {
  const d = fs.mkdtempSync(path.join(os.tmpdir(), "cwd-"));
  // child writes process.cwd() to a RELATIVE file -> lands in its real cwd
  spawn(process.execPath, ["-e", "require('fs').writeFileSync('MARKER.txt', process.cwd())"],
        { cwd: d, stdio: "ignore", ...opts });
[...]

PR #863, opened 2026-06-22
fix(windows): run the better-sqlite3 boot-install via node, not npm.cmd+shell (#861 follow-up)

ken-jo:
## What / Why / How

Follow-up to #861 (`e918eac`). That fix stopped the Windows boot-install EPERM for the optional fetch deps in `start.mjs` by running npm's own CLI through node (`shell:false` honors `cwd`) instead of `npm.cmd` + `shell:true` (which DROPS `cwd` on Windows → npm resolves `node_modules` from the inherited cwd `C:\Windows` → `mkdir C:\Windows\node_modules` → EPERM). The **identical `npm.cmd` cwd-drop** still affected the **native-dep install** in `hooks/ensure-deps.mjs` — the reporter explicitly listed `better-sqlite3` among the EPERMs, and that path was left unfixed.

With this PR applied, the `better-sqlite3` boot-install uses the same cwd-honoring path, so it can no longer EPERM at `C:\Windows\node_modules`.

This PR absorbs `e918eac`'s approach for the native-dep install:

- **`hooks/ensure-deps.mjs`** — resolve `npm-cli.js` beside `process.execPath` and run it via `execFileSync(process.execPath, [npmCliJs, "install", …], { shell:false })` (honors `cwd`) + `windowsHide:true`; fall back to the `npm.cmd` shim only when `npm-cli.js` isn't beside node (so no host regresses); **surface failures to stderr** (was silently swallowed behind `stdio:"pipe"` + an empty `catch`).
- **`tests/hooks/ensure-deps.test.ts`** — pin the install branch to the node-CLI / `shell:false` path + the failure surfacing (source guard, matching this file's existing #408/#634 patterns; the bug is Windows-only and doesn't reproduce on Linux CI).
- preserved: the `ensureNativeCompat` ABI-rebuild paths are untouched (they use `npm.cmd` too, but don't `mkdir node_modules` → no EPERM; left for a separate change).

## Affected platforms

Agent / CLI scope:

- [x] All platforms — `hooks/ensure-deps.mjs` runs on every adapter that boots the MCP server via `start.mjs`. No MCP/adapter behavior change; only the native-dep bootstrap install.

OS scope:

- [x] Windows — the affected OS (`npm.cmd` drops `cwd`). Reachable on Windows + Node when `better-sqlite3` is missing (codex/marketplace clones, broken installs) — the same condition that triggers the `start.mjs` turndown installs `e918eac` already fixed.
- [ ] Linux / macOS — unaffected: `ensureDeps()` returns early on Bun, and POSIX already runs npm with `shell:false` (cwd honored). The node-CLI path is equivalent there.

Scope notes:

- Only the "package not installed at all" install branch is changed. The probe and ABI-rebuild logic are untouched.

## Root Cause

- `e918eac` (#861) fixed `start.mjs` but not `hooks/ensure-deps.mjs`, which still ran `execSync("npm.cmd install better-sqlite3 …", { shell: true })`. On Windows, `shell:true`/`npm.cmd` does not honor the `cwd` option (verified), so the install targets the inherited cwd → `C:\Windows\node_modules` → EPERM.
- `ensureDeps()` only early-returns on Bun (no modern-Node gate), so on Windows + Node it does run this install when `better-sqlite3` is absent.

## Validation

| Check | Result |
| --- | --- |
| `npm run build` + `npm run typecheck` | passed (0 errors) |
| `vitest run tests/hooks/ensure-deps.test.ts` | 26 passed (existing #206/#148/#408/#331 + 2 new #861-follow-up guards) |
| **Windows** (`ssh my-window`, Node v24.13) | `node <npm-cli.js> install is-number` launched from `cwd=C:\Windows` → installed into `<target>\node_modules` (`installedInTarget:true`, `eperm:false`); `npm-cli.js` found beside node |

Behavioral evidence:

- BEFORE: `execSync("npm.cmd install better-sqlite3", { shell:true })` — same cwd-drop class as the (now-fixed) turndown installs → EPERM at `C:\Windows\node_modules` when `better-sqlite3` is missing.
- AFTER: `node <npm-cli.js> install … (shell:false)` installs into the plugin dir regardless of cwd — verified on real Windows from `cwd=C:\Windows`.

Minimal. One install call site in `hooks/ensure-deps.mjs`; no new dependency; no MCP/adapter/hook-routing change. The node-CLI path is the same mechanism `e918eac` already shipped and validated for `start.mjs`. POSIX behavior is unchanged (already `shell:false`).

## Checklist

- [x] Tests added (source guard for the node-CLI/`shell:false` install + failure surfacing), integrated into the existing `tests/hooks/ensure-deps.test.ts`
- [x] `npm run build` + `npm run typecheck` pass
- [x] Generated bundles regenerate via CI (not committed)
- [x] No new runtime dependency
- [x] One PR, single logical fix (absorbs `e918eac` for the native-dep install)
- [ ] 3OS validation — Windows verified (mechanism); Linux/macOS unaffected (Bun early-return / POSIX shell:false)
- [x] Targets `next`

### DIFF
--- a/hooks/ensure-deps.mjs
+++ b/hooks/ensure-deps.mjs
@@ -20,7 +20,7 @@
  */
 
 import { existsSync, copyFileSync, renameSync, unlinkSync } from "node:fs";
-import { execSync } from "node:child_process";
+import { execSync, execFileSync } from "node:child_process";
 import { resolve, dirname } from "node:path";
 import { fileURLToPath, pathToFileURL } from "node:url";
 import { createRequire } from "node:module";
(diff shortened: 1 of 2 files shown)


NgoQuocViet2001:
@ken-jo reviewed the Windows install path here.

The change matches the #861 invariant: the primary native-dep install now runs `node <npm-cli.js>` with `shell:false`, so `cwd` is honored instead of falling through the `npm.cmd` + shell path.

I also checked it locally on Windows:
- `npx vitest run tests/hooks/ensure-deps.test.ts` -> 26 passed
- smoke from parent cwd `C:\Windows` with `cwd` set to a temp target -> package installed into the temp target `node_modules`; nothing was written under `C:\Windows\node_modules`

Looks good to me.

PR #864, opened 2026-06-22
Hide Pi context injection from user entry

RamJam824:
## Summary

Fix Pi adapter context injection so context-mode's internal routing/memory context is appended as a hidden custom message instead of a normal user-role message.

## Why

In Pi, user-role messages can be treated/rendered as interactive conversation entries. Injecting the context-mode guidance as `role: "user"` can leak internal text such as `context-mode active...` into the CLI entry/editor surface.

Pi's extension/session model already supports hidden custom messages via `role: "custom"`, `customType`, and `display: false`; this preserves LLM context injection without presenting the text as user input.

## Changes

- Change Pi context hook injection to:
  - `role: "custom"`
  - `customType: "context-mode-context"`
  - `display: false`
- Update comments to document why user-role injection is avoided.
- Add a regression test proving Pi context-mode guidance is injected as hidden custom context, not a user entry.

## Verification

- `pnpm vitest run tests/adapters/pi-help-skip-bootstrap.test.ts` — 19 tests passed

Note: a broader ad-hoc targeted `tsc` command currently surfaces existing repo/typeconfig issues unrelated to this patch (`codex` adapter narrowing and better-sqlite3 import flags), so I used the existing focused Vitest coverage for this Pi adapter path.

### DIFF
--- a/src/adapters/pi/extension.ts
+++ b/src/adapters/pi/extension.ts
@@ -696,10 +696,13 @@ export default function piExtension(pi: any): void {
       }
 
       // Store extra context (routing anchor, active_memory, resume, behavioralDirective)
-      // for injection via the 'context' hook as a message, NOT as a systemPrompt
-      // modification. Mutating systemPrompt breaks prefix prompt caching on
-      // DeepSeek/Anthropic/OpenAI because the system message sits at messages[0]
-      // and any change invalidates the entire cache chain.
+      // for injection via the 'context' hook as a hidden custom message, NOT as a
+      // systemPrompt modification. Mutating systemPrompt breaks prefix prompt
+      // caching on DeepSeek/Anthropic/OpenAI because the system message sits at
+      // messages[0] and any change invalidates the entire cache chain. A plain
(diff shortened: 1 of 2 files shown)


NgoQuocViet2001:
@RamJam824 I checked the Pi path locally. The new focused test passes, but the broader Pi suite still has stale context-hook expectations from the earlier `role: "user"` behavior.

Ran:
- `npx vitest run tests/pi-extension.test.ts tests/adapters/pi-help-skip-bootstrap.test.ts`

Result:
- `tests/adapters/pi-help-skip-bootstrap.test.ts` passed
- `tests/pi-extension.test.ts` failed 4 assertions because the injected context message is now `role: "custom"`, while the tests still expect `role: "user"` at `tests/pi-extension.test.ts:681`, `:725`, `:921`, and `:1009`

I think the patch is directionally consistent, but those existing Pi resume/routing/active-memory tests should be updated to assert the hidden custom shape (`role`, `customType`, `display: false`) while keeping the current content and no-systemPrompt checks.

PR #876, opened 2026-06-24
fix(omp): seed APPEND_SYSTEM.md routing instructions

NgoQuocViet2001:
## Summary
- seed the packaged OMP routing instructions into `~/.omp/agent/APPEND_SYSTEM.md` when the plugin loads
- preserve OMP's built-in system prompt and native tool definitions
- preserve an existing user `APPEND_SYSTEM.md` instead of overwriting it
- add regression coverage for missing and existing append prompts

## Why `APPEND_SYSTEM.md`
OMP treats `SYSTEM.md` as a replacement for its default system prompt. Using `APPEND_SYSTEM.md` appends context-mode's routing rules while retaining OMP's built-in tool guidance.

Closes https://github.com/mksglu/context-mode/issues/873

## Validation
- `npx vitest run tests/adapters/omp-plugin.test.ts` — 24/24 passed
- `npm run typecheck`
- `npm run build`
- `git diff --check`

### DIFF
--- a/src/adapters/omp/plugin.ts
+++ b/src/adapters/omp/plugin.ts
@@ -141,18 +141,18 @@ function ensureMcpServerRegistered(): void {
 }
 
 /**
- * Seed OMP's native system-prompt file with context-mode routing rules.
+ * Append context-mode routing rules to OMP's native system prompt.
  *
- * OMP discovers `~/.omp/agent/SYSTEM.md` on startup, but `omp plugin install`
- * only installs the npm package. Copy the packaged instructions lazily and
- * never overwrite a user's existing system prompt.
+ * OMP discovers `~/.omp/agent/APPEND_SYSTEM.md` on startup and keeps its
+ * built-in tool guidance intact. Copy the packaged instructions lazily and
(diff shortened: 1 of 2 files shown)


sjclayton:
@NgoQuocViet2001 Should be `APPEND_SYSTEM.md` as per the docs, and not `SYSTEM.md` as seeding to `SYSTEM.md` will clobber all built-in tool definitions in the OMP default system prompt. 

I have verified this… as OMP becomes completely unaware of any of the native tools in a given session when copying the `SYSTEM.md` supplied by the plugin into `~/.omp/agent/` directly. I figured it out because I noticed the memory recall tool no longer working, model couldn't find it at all, and stated it didn't exist.

When using `APPEND_SYSTEM.md` everything works as it should again.

NgoQuocViet2001:
@sjclayton Thanks for flagging this and for verifying the OMP behavior. I switched the target to `~/.omp/agent/APPEND_SYSTEM.md`, kept existing append prompts untouched, and added coverage that ensures `SYSTEM.md` is not created. The focused suite passes 24/24, plus typecheck and build.

PR #898, opened 2026-06-30
fix(store): cap oversized markdown chunks

### LINKED ISSUE #878: #chunkMarkdown allows oversized single paragraphs, bypassing MAX_CHUNK_BYTES
# `#chunkMarkdown` allows oversized single paragraphs, bypassing `MAX_CHUNK_BYTES`

## Summary

`context-mode@1.0.166` can persist Markdown chunks far above the 4 KiB
`MAX_CHUNK_BYTES` limit when a Markdown section contains one oversized
paragraph or line.

This remains reachable through `ctx_batch_execute`: command output is wrapped
under a `# <label>` heading and passed to `store.index(...)`, selecting the
Markdown chunker. The plain-text hardening from #781 therefore does not cover
this path.

## Environment

- context-mode: `1.0.166`
- platform: Codex CLI `0.141.0`
- OS: Windows
- Node: bundled/global Node.js used by context-mode
- storage: isolated `CONTEXT_MODE_DATA_DIR`

## Minimal reproduction

Call `ctx_batch_execute` with one long output line:

```json
{
  "commands": [
    {
      "label": "Synthetic oversized line",
      "command": "$s='a'*1200000; Write-Output ($s+' needleXYZ')"
    }
  ],
  "queries": ["needleXYZ"],
  "query_scope": "batch",
  "concurrency": 2,
  "timeout": 30000
}
```

The call succeeds quickly:

```text
Executed 1 commands (7 lines, 1172.0KB). Indexed 2 sections.
elapsed: 524ms
```

Inspecting the resulting SQLite FTS tables:

```sql
SELECT count(*), max(length(content)), sum(length(content)) FROM chunks;
```

Actual result:

```text
2 | 1200010 | 1200089
```

The same oversized content is also present in `chunks_trigram`.

## Expected

No persisted chunk should exceed `MAX_CHUNK_BYTES` except for the documented
maximum 1–4 byte Unicode-code-point overshoot.

For this input, the 1.2 MB paragraph should be split into byte-capped chunks,
or the payload should be rejected/truncated with a diagnostic.

## Actual

A single `1,200,010`-character chunk is persisted despite
`MAX_CHUNK_BYTES = 4096`.

## Root cause

In `src/store.ts`, `#chunkMarkdown` only flushes an oversized candidate when:

```ts
Buffer.byteLength(candidate) > maxChunkBytes && accumulator.length > 1
```

If one paragraph itself exceeds the cap, `accumulator.length` remains `1`.
`flushAccumulator()` then persists that paragraph unchanged.

Relevant locations in v1.0.166:

- `src/store.ts:153`: `MAX_CHUNK_BYTES = 4096`
- `src/store.ts:1646`: `#chunkMarkdown`
- `src/store.ts:1667`: paragraph-boundary splitting
- `src/store.ts:1689`: split condition requiring `accumulator.length > 1`
- `src/store.ts:1793`: existing byte-aware `#splitOversizedPlainChunk`

## Suggested fix

When a Markdown paragraph/code block exceeds `maxChunkBytes`, sub-split it by
UTF-8 bytes before persistence. The existing byte-aware helper could be
generalized and reused.

Also consider a final invariant at the store insertion choke point:

```ts
Buffer.byteLength(chunk.content) <= maxChunkBytes + 4
```

That would prevent future format-specific chunkers from silently bypassing the
storage cap.

## Regression tests

Add cases asserting the maximum persisted byte length for:

1. one oversized Markdown paragraph;
2. one oversized Markdown line;
3. one oversized fenced code block;
4. multibyte CJK/emoji content;
5. `ctx_batch_execute` output wrapped under a Markdown heading.

## Relationship to #781

#781 fixed the plain-text chunker. This is a separate Markdown-path gap exposed
by `ctx_batch_execute`'s heading wrapper, so reopening #781 may be less precise
than filing a follow-up issue linked to it.

ken-jo:
## What / Why / How

Fixes #878.

`ContentStore.index()` routes Markdown-shaped content through `#chunkMarkdown`. A `ctx_batch_execute` response is stored as `# <label>\n\n<output>`, so one oversized command-output line becomes one oversized Markdown paragraph. On `origin/next` before this patch, a 1.2 MB payload persisted one `1,200,010` byte row in both `chunks` and `chunks_trigram`, bypassing the `MAX_CHUNK_BYTES = 4096` storage cap.

With this PR applied, oversized Markdown paragraphs, lines, and fenced code blocks are split with the existing UTF-8 byte-aware splitter before insertion. The same 1.2 MB repro now stores 294 chunks, with `maxBytes = 4096` in both FTS tables.

This PR:

- routes oversized single Markdown paragraphs through the existing byte-capped splitter
- preserves paragraph-boundary chunking for ordinary oversized Markdown sections
- preserves the existing CJK/emoji-safe byte splitting invariant from the plain-text hardening
- adds regression coverage to the existing byte-cap test file for Markdown paragraph, line, fenced code, multibyte text, and the `ctx_batch_execute` heading-wrapper shape
- updates generated `server.bundle.mjs` and `cli.bundle.mjs`

## Affected platforms

Agent / CLI scope:

- [ ] Claude Code
- [ ] Cursor
- [ ] VS Code Copilot
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw / Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [x] All standalone MCP server launchers
- [x] All platforms

OS scope:

- [x] Linux
- [ ] macOS
- [ ] Windows
- [x] All operating systems

Scope notes:

- This is a core `ContentStore` Markdown chunking fix. Any adapter that indexes Markdown content through `ctx_index`, `ctx_fetch_and_index`, or `ctx_batch_execute` benefits from the same storage cap.
- The code path uses `Buffer.byteLength` and existing Unicode-safe splitting, so the behavior is not OS-specific. macOS and Windows are covered by PR CI.

## Root Cause

Git archaeology:

- `609a3a5` introduced `MAX_CHUNK_BYTES` and Markdown paragraph-boundary splitting for oversized sections.
- `6f699fa` / #789 hardened `#chunkPlainText` with `#splitOversizedPlainChunk` and `#byteCappedPrefix`, preserving UTF-8 code points while keeping stored chunks at or below 4096 bytes.
- The Markdown path still kept the `accumulator.length > 1` guard, so a single paragraph or code block wider than the cap stayed in the accumulator and was emitted unchanged by `flushAccumulator()`.

The failure happens because paragraph-boundary splitting only flushed after at least two paragraphs were accumulated. A single oversized paragraph never reached the split helper.

- ordinary Markdown sections still split at paragraph boundaries first
- oversized single lines use the same UTF-8 byte-aware splitting as #789
- generated bundles stay in sync with source

## Validation

| --- | --- | --- |
| Linux local, Node v22.22.3 | `npx tsx -e '<ContentStore #878 repro harness>'` before patch | Confirmed: `chunks.maxBytes = 1200010`, `chunks_trigram.maxBytes = 1200010` |
| Linux local, Node v22.22.3 | `npx tsx -e '<ContentStore #878 repro harness>'` with this PR applied | Passed: `chunks.count = 294`, `chunks.maxBytes = 4096`, `chunks_trigram.maxBytes = 4096` |
| Linux local, Node v22.22.3 | `npm run build` | Passed |
| Linux local, Node v22.22.3 | `npm run typecheck` | Passed |
| Linux local, Node v22.22.3 | `npx vitest run tests/store-bytecap.test.ts tests/store.test.ts tests/core/search.test.ts --reporter=dot` | Passed: 3 files, 286 tests |
| macOS via PR CI | `test (macos-latest)` | Pending |
| Windows via PR CI | `test (windows-latest)` | Pending |

Behavioral evidence:

- BEFORE: the 1.2 MB Markdown heading-wrapper repro stored 2 rows and one row was `1,200,010` bytes in both `chunks` and `chunks_trigram`.
- AFTER: the same repro stores 294 rows and both FTS tables report `maxBytes = 4096`.

Agent-working smoke:

- Not adapter-specific. This patch changes the shared content-store insertion path used by MCP tools across adapters. PR CI covers cross-OS execution.

- oversized Markdown paragraph under heading
- oversized single Markdown line
- oversized fenced code block
- multibyte CJK paragraph split by UTF-8 bytes
- `ctx_batch_execute`-style `# <label>\n\n<output>` wrapper with a 1.2 MB line

The affected path is the shared `ContentStore.index()` Markdown chunker. It is not tied to a specific agent host, hook format, path separator, shell, or OS. The patch reuses existing `Buffer.byteLength` and code-point-preserving splitting already used by the plain-text path.

## Checklist

- [x] Regression coverage integrated into an existing relevant test file, or new test file justified
- [x] `npm run build` passes
- [x] Targeted tests pass
- [x] Generated bundles updated when source changes require it
- [x] Docs updated if user-facing behavior changed
- [ ] 3OS validation completed or each unverified OS has a concrete reason
- [x] Agent-working smoke completed for agent-facing changes
- [x] Targets `next` branch unless this is an explicit hotfix


Tomodad (commented):
The core fix is directionally correct and addresses #878: oversized Markdown paragraphs/lines are routed through the existing UTF-8 byte-safe splitter, and the 1.2 MiB batch-style regression demonstrates the row-shape fix.

A few follow-ups are worth making explicit before merge (or tracking separately if you want to keep this PR narrow):

1. Please add a direct oversized Markdown emoji regression alongside the CJK case. The helper should already be surrogate-safe, but the Markdown path is not directly covered.
2. This is still a format-specific guard, not a final persistence invariant before FTS insertion. A later index path could bypass the cap again; a lossless final assert/normalizer would be stronger.
3. Oversized fenced code blocks are split as plain lines, so child chunks do not necessarily contain balanced opening/closing fences.
4. `hasCode` is section-wide, so prose children from a mixed code/prose section can be classified as code.

I did not find a silent hard clip: the splitter consumes the full remaining string rather than truncating it. I would treat the fence/classification points as either targeted fixes here or clearly documented follow-ups, not as reasons to duplicate this PR.

PR #906, opened 2026-07-02
docs(codex): pin platform env in manual config

NgoQuocViet2001:
Follow-up to #722.

## Summary
- add `CONTEXT_MODE_PLATFORM=codex` to `configs/codex/config.toml`
- pin that sample config with a small Codex config integrity test

## Validation
- `node_modules/.bin/vitest.cmd run tests/adapters/codex.test.ts -t "Codex matcher parity" --reporter=dot`
- `node_modules/.bin/tsc.cmd -b --noEmit`
- `git diff --check`

### DIFF
--- a/configs/codex/config.toml
+++ b/configs/codex/config.toml
@@ -6,3 +6,6 @@ hooks = true
 
 [mcp_servers.context-mode]
 command = "context-mode"
+
+[mcp_servers.context-mode.env]
+CONTEXT_MODE_PLATFORM = "codex"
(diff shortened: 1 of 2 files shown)


NgoQuocViet2001:
@mksglu @ken-jo small follow-up while catching up on the install docs: README/manual fallback and the Codex plugin config already set `CONTEXT_MODE_PLATFORM=codex`, but the shipped `configs/codex/config.toml` sample was still missing it.

Kept this separate from #884 since it touches the config sample + a tiny guard test, not the larger docs sync.

Issue #659, opened 2026-05-21
Preview truncation slices UTF-16 surrogate pairs, producing orphan `\uD83D` in tool_result and breaking the host API request


ccheng555:
# Preview truncation slices UTF-16 surrogate pairs, producing orphan `\uD83D` in tool_result and breaking the host API request

## Summary

`ctx_batch_execute` (and any path that emits a markdown preview via function `QF`) truncates the preview with `String.prototype.slice(0, cT)` where `cT = 3072`. `slice` counts UTF-16 code units. When position 3072 falls between the two halves of an emoji surrogate pair, the high half stays and the low half is dropped. `JSON.stringify` later emits the orphan as a literal `\uD83D` escape in the tool_result payload. The host LLM API then rejects the next request with `400 The request body is not valid JSON: no low surrogate in string` per RFC 8259.

The session cannot recover until the bad message is removed from the transcript.

## Minimal repro

```js
// Node, no deps. Reproduces the QF slice behavior exactly.
const cT = 3072;
const filler = "a".repeat(cT - 1);
const md = filler + "🟡 status indicator";   // 🟡 = U+1F7E1 = \uD83D \uDFE1
const sliced = md.length > cT
  ? md.slice(0, cT) + "\n…\n[truncated, use ctx_search() for full content]"
  : md;

// Verify the slice ends with an unpaired high surrogate.
const lastUnit = sliced.charCodeAt(cT - 1);
console.assert(lastUnit >= 0xD800 && lastUnit <= 0xDBFF, "should be high surrogate");
console.assert(sliced.charCodeAt(cT) < 0xDC00 || sliced.charCodeAt(cT) > 0xDFFF, "should NOT be a low surrogate after");

// JSON.stringify emits the orphan as a literal \u escape.
const body = JSON.stringify({content: sliced});
console.assert(/\\uD83D[^\\]/i.test(body), "body should contain orphan \\uD83D");

console.log("Repro confirmed: orphan high surrogate present at offset", body.indexOf("\\uD83D"));
```

Send `body` to any RFC-8259-strict JSON consumer (e.g. POST it as user content to a chat completions endpoint). Returns `400 ... no low surrogate in string`.

## Root cause

`cli.bundle.mjs` v1.0.146, function `QF` (also present byte-for-byte in v1.0.131; `cmp -s` returns 0 between the two bundles):

```js
xr(Buffer.byteLength(t.markdown));
let s = t.markdown.length > cT
[...]

PR #907, opened 2026-07-02
fix(session): keep SessionStart truncation surrogate-safe

### LINKED ISSUE #903: SessionStart injection truncates UTF-16 surrogate pairs, producing orphan high surrogates that break the host API request
# SessionStart injection truncates UTF-16 surrogate pairs, producing orphan high surrogates that break the host API request

## Summary

Same failure class as #659, but on a different path: the **SessionStart context injection** hooks truncate event data with `String.prototype.substring()` / `.slice()`, which count UTF-16 code units. When the cut lands between the two halves of an astral-plane character (emoji), an orphan high surrogate is left at the end of the truncated string. It flows into `additionalContext` emitted by `hooks/sessionstart.mjs`, the host (Claude Code) embeds it into the **first API request of the session**, and the API rejects it:

```
API Error: 400 The request body is not valid JSON: no low surrogate in string: line 1 column 125449 (char 125448)
```

Because this happens in SessionStart injection, the session is broken from the very first prompt — there is no bad tool_result to remove; the user cannot recover without disabling the plugin or purging the offending event from the session DB.

The fix for #659 covered the MCP preview path (`slice(0, 3072)`), but the hook-side truncations were not covered.

## Affected sites (current `main`, v1.0.169)

- `hooks/session-directive.mjs:43` — `raw.substring(0, DATA_REF_ENTRY_MAX - 3)`
- `hooks/session-directive.mjs:288` — `lastPrompt.substring(0, 297)`
- `hooks/session-directive.mjs:340,357` — `ev.data.substring(0, 147)`
- `hooks/session-directive.mjs:403` — `ev.data.substring(0, 117)`
- `hooks/auto-injection.mjs:37` — `roleEvent.data.slice(0, 400)`
- `hooks/auto-injection.mjs:45` — `d.data.slice(0, 100)`
- `hooks/session-snapshot.bundle.mjs` — `o.slice(0,80)` (bundled)

Any event stored in the session DB whose text contains emoji (commit messages, i18n strings, user prompts, tool output) can trigger it when its length exceeds the truncation limit and the boundary splits a surrogate pair.

## Minimal repro

```js
// Node, no deps. Mirrors session-directive.mjs:340.
const data = "x".repeat(146) + "🟡 status";      // 🟡 = U+1F7E1 = \uD83D \uDFE1
const text = data.length > 150 ? data.substring(0, 147) + "..." : data;

console.log(text.charCodeAt(146).toString(16));   // "d83d" — orphan high surrogate
JSON.parse(JSON.stringify(text));                  // JS tolerates it (well-formed stringify
                                                   // emits literal \ud83d escape)…
// …but a strict RFC 8259 parser (the API backend) rejects the request body:
// python3 -c 'import json; json.loads(open("payload").read())'
// → "no low surrogate in string"
```

## Suggested fix

Either fix each truncation with a well-formed helper:

```js
const truncate = (s, n) => {
  let out = s.slice(0, n);
  // drop a split high surrogate at the cut point
  if (/[\uD800-\uDBFF]$/.test(out)) out = out.slice(0, -1);
  return out;
};
```

…or (belt-and-suspenders, catches all current and future truncation sites in one place) sanitize `additionalContext` once at the emit point in `hooks/sessionstart.mjs` before `JSON.stringify`:

```js
additionalContext = additionalContext.replace(
  /[\uD800-\uDBFF](?![\uDC00-\uDFFF])|(?<![\uD800-\uDBFF])[\uDC00-\uDFFF]/g,
  "",
);
```

`String.prototype.toWellFormed()` would also work if the minimum supported Node is ≥ 20.

I've verified the emit-point sanitizer locally (v1.0.135): paired emoji survive, orphan surrogates are stripped, and the hook output parses under a strict JSON parser.

NgoQuocViet2001:
Fixes #903. Thanks @mykhailoPavliuk for the clean repro.

## Summary
- add a hook-local `charSafePrefix` helper so direct hook runtime paths do not split surrogate pairs
- use it in SessionStart directive and compact auto-injection truncation paths
- use the existing source helper for snapshot query prefixes and regenerate the session snapshot bundle
- add regression coverage for the #903 boundary cases

## Validation
- `npm run typecheck`
- `npm run build`
- `node_modules/.bin/vitest.cmd run tests/session/resume-leg-boundary.test.ts tests/session/session-snapshot.test.ts --reporter=dot`
- `node_modules/.bin/vitest.cmd run tests/hooks/integration.test.ts -t 'buildAutoInjection' --reporter=dot`
- `node_modules/.bin/vitest.cmd run tests/hooks/integration.test.ts -t 'empty stdin resilience' --reporter=verbose`
- `git diff --check`


NgoQuocViet2001:
@mksglu @ken-jo Consolidating the current status of my three pending PRs: #876, #906, and this one are rebased on `next`, have no outstanding review threads, and their CI is green. I kept them intentionally separate by scope; please let me know if you prefer a particular landing order or any follow-up changes.

valkyriweb:
Independent reproduction of the `buildQueries` site this PR fixes, in case it helps prioritise the review.

I hit the same defect from the other direction — auditing a downstream install for the #659/#903 class rather than reading the tracker — and arrived at the identical one-line change before finding this PR. Reporting it here rather than opening a duplicate.

## Site

`src/session/snapshot.ts`, `buildQueries`:

```js
const trimmed = s.length > 80 ? s.slice(0, 80) : s;
```

Compiled to `build/session/snapshot.js` and inlined into `hooks/session-snapshot.bundle.mjs` as `o.length>80?o.slice(0,80):o`, which is the `o.slice(0,80)` entry already listed in #903.

## Reproduction

Against an unmodified 1.0.169 install. 79 filler units place the high surrogate at index 79 and the low surrogate at index 80, so an 80-unit cut splits the pair:

```js
const term = "x".repeat(79) + "😀 tail";

const bare = term.length > 80 ? term.slice(0, 80) : term;   // buildQueries, verbatim
bare.charCodeAt(79).toString(16);      // "d83d" — orphan high surrogate
JSON.stringify(bare).slice(-14);       // 'xxxxxxx\ud83d"'

const safe = charSafePrefix(term, 80); // 79 units, pair dropped intact
```

Boundary sweep — only the splitting offset is affected, which is why this stays latent until an emoji lands on the cut:

| filler | `slice(0, 80)` | `charSafePrefix(s, 80)` |
|---|---|---|
| 77 | clean | clean |
| 78 | clean | clean |
| **79** | **orphan `\uD83D`** | clean, backs off to 79 |
| 80 | clean | clean |

## Reachability beyond SessionStart

Worth noting for scope, since #903 frames this as SessionStart-specific: on the Pi adapter the same `buildQueries` output reaches the host through a second path that does not involve `hooks/sessionstart.mjs` at all.

`buildResumeSnapshot` is called from the Pi extension's `session_before_compact` handler using `db.getEvents(sessionId)` with no `opts` — so no `minPriority` filter, up to 1000 events of any category. The snapshot is stored via `upsertResume` and injected verbatim on the next `before_agent_start`, with no re-truncation between storage and delivery.

That means the payload is poisoned for the turn immediately after every compaction, and the offending text can come from any stored event — `error_tool` data is `safeString(tool_response)` (raw tool output, uncapped), `rule_content` events carry verbatim `CLAUDE.md`/`AGENTS.md` contents, and decision events carry user-typed answers. Emoji in any of those is enough.

This PR's change to `src/session/snapshot.ts` covers that path too, since both routes go through `buildQueries`. Flagging it only because the fix is broader than the issue title suggests, which may matter for how it gets prioritised.

## On the fix itself

`charSafePrefix` is the right call over a local helper — it already handles the `<= maxChars` passthrough, so the swap is behaviour-preserving below the cut, and it keeps the surrogate logic in `truncate.ts` where #273 and #286 already established it.

I verified the change in a scratch copy of the installed package: snapshot output contains no lone surrogate, survives strict UTF-8 round-tripping, `JSON.stringify` emits no orphan escape, and the truncated term is still present. Removing the change reintroduces the orphan, so the check isn't passing vacuously.

## Status

At time of writing, neither `next` nor `main` contains the fix, and this PR reports `mergeable_state: clean` with no outstanding review threads.

Happy to port the reproduction into `tests/session/session-snapshot.test.ts` in the existing style if a regression test covering the split boundary specifically would help — but this PR already adds tests there, so it may well be redundant. Not looking to duplicate anyone's work; say the word if it's useful.

Issue #856, opened 2026-06-22
Pi host: a casual user phrase gets frozen as behavioral_directive and re-injected every turn → do-nothing loop (1.0.163 & next)


riddlex-beep:
## Summary

On the pi/omp adapter, a casual past user utterance can get frozen as a `<behavioral_directive>` and **re-injected on every turn** as a standing order. In long sessions this produces a do-nothing loop: the agent keeps replying *"since you said 'that's fine for now,' I'll leave it and not touch anything else"* and ignores new instructions, regardless of what the user types.

It lives entirely in the prompt-injection path (`adapters/pi/extension.js` + `hooks/`), and the responsible code is present on stock `1.0.163` and on current `next`.

## Root cause

1. **Over-eager classification.** `src/session/extract.ts` tags a short imperative with no `?` and no comma as a `role` event (priority 3). Casual, throwaway phrases (e.g. `that's fine for now`, `go with the second option`) qualify and get stored as a `role`.

2. **Per-turn re-injection.** In `build/adapters/pi/extension.js` the `before_agent_start` handler builds `behavioralDirective` from `db.getEvents(_sessionId, { minPriority: 3 })` (role = priority 3, so it is always included) and pushes it into the prompt **every turn** — not just on compaction.

3. **Standing-order framing.** `hooks/routing-block.mjs:60-61` reinforces it:
   > Skills, roles, and decisions set during this session remain active until the user revokes them.
   > Do not drop behavioral directives as context grows.

Combined effect: a one-off phrase becomes a frozen, high-priority directive the model is told never to drop and to treat as active "until revoked." Normal follow-ups (`continue`, questions) don't re-classify as a new `role` (questions become `intent`), so the stale directive **never clears** and the agent loops.

## Reproduction

1. In a pi/omp session, say a short imperative with no `?`/comma that implies stopping, e.g. `that's fine for now`.
2. Continue the session past a compaction.
3. Observe `<behavioral_directive>that's fine for now</behavioral_directive>` injected each turn.
4. Ask the agent to do anything new — it restates the directive and declines, looping.

The frozen text is visible in the injected block every turn, e.g.:
```
<behavioral_directive>
that's fine for now
</behavioral_directive>
```

## Impact

- Long sessions stall: the agent refuses to act on live instructions.
- The phrase is often something the user considers resolved, so the behavior is confusing and hard to attribute.
- Only an explicit "revoke all prior directives" message reliably breaks it (matching the routing-block wording), which users won't guess.

[...]

PR #918, opened 2026-07-04
fix(routing-block): remove injection-shaped framing (#911)

### LINKED ISSUE #911: Injected session_continuity/priority_instructions framing trips Claude Code's auto-mode classifier (Auto-Mode Bypass)
## Summary

Claude Code's auto-mode permission classifier flagged an Agent-tool dispatch as a policy violation (`Auto-Mode Bypass` category) because the dispatch prompt contained context-mode's injected `<context_window_protection>` routing block — specifically the `<session_continuity>` clause ("a past phrase does not bind you"). The denial text read:

> [Auto-Mode Bypass] The Agent dispatch prompt contains an injected "context_window_protection" block instructing the agent to load and route through new MCP tools and to disregard prior session directives — a prompt-injection payload attempting to manipulate agent behavior and evade oversight, not something the user authorized.

This also blocked `ctx_upgrade`'s self-update mechanism on retry, since the classifier treated pulling new code from a source already flagged in the same session as a "Code from External" risk requiring direct human review outside auto mode.

## Root cause (as far as I can tell)

Not a data-exfiltration issue — I checked the bundled `cli.bundle.mjs`/`server.bundle.mjs` directly for outbound network calls and found nothing beyond a public npm-registry version check and a manually-invoked `ctx_insight` browser-open. The "nothing leaves your machine" claim in the README held up under inspection.

The actual trigger is purely the **framing** of `hooks/routing-block.mjs`'s `createRoutingBlock()`: content wrapped in a block literally named `priority_instructions`, delivered automatically via a SessionStart/PreToolUse hook (not user-initiated), telling the agent "a past phrase does not bind you." Structurally that's indistinguishable from a classic prompt-injection technique (claim elevated authority, tell the agent to discount prior instructions), regardless of intent.

I found commit f262852 (2026-06-22, #856) already softened this once — from an "inescapable active-until-revoked standing order" to the current "memory aid, not a standing order" phrasing, to fix a do-nothing-loop bug where casual phrases froze as re-injected `behavioral_directive`s. That's a real improvement, but the residual framing still reads as injection-shaped to Claude Code's classifier even after the softening.

## Reproduction

1. Install context-mode as a Claude Code plugin, in a session with auto mode enabled.
2. Dispatch any subagent via the Agent tool (the injected `<context_window_protection>` block rides along automatically).
3. The permission classifier may flag the dispatch under `Auto-Mode Bypass`, citing the injected block's "disregard prior session directives" framing.

Not deterministic — happened in one session/window but not consistently across repeated attempts in another, consistent with an LLM-judged classifier rather than a static rule match.

## Suggested remediation directions

- Drop or further soften the `<session_continuity>` "a past phrase does not bind you" line — the intent (avoid stale captured context freezing into a standing directive) can likely be conveyed without language that reads as "discount prior instructions."
- Consider renaming `<priority_instructions>` to something that doesn't claim elevated authority over the rest of the prompt (e.g. `<usage_hint>` or `<tool_guidance>`).
- An opt-out/quiet-mode flag for security-conscious users who want the sandboxing/search tools without the injected routing framing would also address this without changing default behavior.

Happy to share the full transcript/denial text if useful. Thanks for the tool — the sandboxing and FTS5 search are genuinely useful, this is specifically about the injected framing tripping a security classifier.

YuryKoltashevModdelix:
## Summary

Fixes #911. The injected `<context_window_protection>` routing block (spliced into every Agent-tool dispatch prompt, see #233) used two pieces of framing that read as classic prompt-injection to Claude Code's auto-mode permission classifier:

- A `<priority_instructions>` tag name that claims elevated authority over the rest of the prompt.
- A `session_continuity` clause stating "the user's most recent message always takes precedence... a past phrase does not bind you" — instructing the agent to discount prior captured directives.

Both are accurate/well-intentioned in isolation (avoid a stale captured role/skill freezing into a standing directive, per the #856 fix), but structurally they're indistinguishable from a prompt trying to claim priority and tell the model to ignore prior instructions — which is exactly what a security classifier is supposed to catch.

## Changes

- `hooks/routing-block.mjs`: rename `<priority_instructions>` → `<usage_hint>`; drop the "past phrase does not bind you" sentence from `<session_continuity>`, keeping the already-softened "memory aid, not a standing order" framing from #856 intact.
- `tests/opencode-plugin.test.ts`: update assertion for the renamed tag.
- `tests/core/routing.test.ts`: add a regression test (`Issue #911`) asserting the block never contains the old tag name or the discounting-prior-instructions phrasing.
- `cli.bundle.mjs` regenerated via `npm run build` (only bundle that embeds the routing block; `server.bundle.mjs` and the other hook bundles are unaffected).

Intentionally did not add an opt-out/quiet-mode flag (also suggested in the issue) — that's a larger feature decision I'll leave to a maintainer, this PR is scoped to the two concrete wording triggers.

## Test plan

- [x] `npm run build` — tsc + bundle + assert-bundle + assert-asymmetric-drift all pass
- [x] `npx vitest run` — full suite: 4716 passed, 0 failed, 28 skipped
- [x] `npx vitest run tests/core/routing.test.ts tests/opencode-plugin.test.ts` — 96 passed including new #911 regression tests


caioribeiroclw-pixel:
This looks like the right minimal fix for the two obvious classifier triggers.

One thing I would still consider before merging: the issue's second denial quoted the “raw output kept out of the conversation” shape as an oversight-bypass signal, not just the `priority_instructions` / stale-directive wording.

The remaining `usage_hint` still says:

> surface only the derived answer — the raw bytes stay out

That is true for context economy, but to a permission/security classifier it can still read as “hide tool output from the review surface.” A small auditability clause may make the routing block less ambiguous without weakening the product:

```text
Use context-mode to reduce context volume, not to hide actions or evidence. If the user, host, or permission review needs the raw result, provide or retrieve the relevant raw/indexed source reference instead of only a derived summary.
```

Even better if the regression test asserts the block contains this kind of “not for concealment / raw source remains retrievable for review” invariant. That gives the classifier a positive safety statement instead of only removing suspicious phrasing.

PR #924, opened 2026-07-06
start: restore original CWD before importing server bundle

### LINKED ISSUE #923: start.mjs: process.chdir(__dirname) changes MCP subprocess CWD, breaking gnome-terminal new-tab directory
## Bug

`start.mjs` calls `process.chdir(__dirname)` early in startup and never restores the original working directory. This causes gnome-terminal's \"open new tab in same directory\" feature to land in the plugin cache directory instead of the user's project directory.

## Root cause

```js
const originalCwd = process.cwd();
process.chdir(__dirname);   // CWD is now e.g. ~/.claude/plugins/cache/context-mode/context-mode/1.0.162
// ... never restored
await import("./server.bundle.mjs");
```

gnome-terminal (and other VTE-based terminals) determine the \"current directory\" for new tabs by reading `/proc/<pid>/cwd` of the foreground process group. The MCP server subprocess inherits the terminal's process group, so terminals see its CWD = plugin directory.

## Why the chdir is unnecessary

ES module `import("./server.bundle.mjs")` resolves relative to `import.meta.url`, **not** `process.cwd()`. The original CWD is already captured and written to `CLAUDE_PROJECT_DIR` / `CONTEXT_MODE_PROJECT_DIR` env vars before the chdir. The server uses those env vars, not `process.cwd()`. The self-heal code all uses `__dirname` and `resolve()` with absolute paths.

## Fix

Restore the original CWD after env vars are set and before importing the server bundle:

```js
// After: process.env.CONTEXT_MODE_PROJECT_DIR = safeOriginalCwd;
process.chdir(originalCwd);   // restore — import() uses import.meta.url, not cwd
await import("./server.bundle.mjs");
```

## Environment

- context-mode v1.0.162
- gnome-terminal / VTE on Linux
- Claude Code plugin system

mattst88:
Fixes #923.

`start.mjs` calls `process.chdir(__dirname)` early in startup for self-heal path operations, but never restores the original working directory. This causes VTE-based terminals (gnome-terminal, etc.) to open new tabs in the plugin cache directory instead of the user's project directory.

VTE determines the "open new tab here" directory by reading `/proc/<pid>/cwd` of the foreground process group. The MCP server subprocess inherits the terminal's process group, so its CWD is what the terminal sees.

ES module `import()` resolves relative to `import.meta.url`, not `process.cwd()`, so restoring the original CWD before the server bundle import is safe. The project directory env vars (`CLAUDE_PROJECT_DIR`, `CONTEXT_MODE_PROJECT_DIR`) are already set from `originalCwd` before this point, so server behavior is unaffected.

### DIFF
--- a/start.mjs
+++ b/start.mjs
@@ -597,6 +597,16 @@ if (!process.env.VITEST) {
   }
 }
 
+// Restore original CWD before importing server — ES module imports resolve
+// relative to import.meta.url, not process.cwd(), so this is safe. Without
+// this, any tool that resolves a directory from the foreground process cwd
+// (VTE terminals reading /proc/<pid>/cwd, pane/workspace managers) sees the
+// plugin cache dir instead of the project.
+// Restore safeOriginalCwd, not originalCwd: on respawn after /ctx-upgrade the
+// launch cwd is itself the plugin install dir, which would reintroduce the bug.
+// See: https://github.com/mksglu/context-mode/issues/923
+try { process.chdir(safeOriginalCwd ?? homedir()); } catch { /* best effort — restore cwd for foreground-cwd inheritance */ }
+
 // Bundle exists (CI-built) — start instantly
 if (existsSync(resolve(__dirname, "server.bundle.mjs"))) {
   await import("./server.bundle.mjs");
--- a/tests/util/start-mjs-cwd-restore.test.ts
+++ b/tests/util/start-mjs-cwd-restore.test.ts
@@ -1,0 +1,70 @@
+/**
+ * Issue #923 — start.mjs must restore the original CWD before importing the
+ * server bundle so tools that resolve a directory from the foreground process
+ * cwd (VTE terminals, pane/workspace managers) see the project directory rather
+ * than the plugin cache directory.
+ *
+ * Uses static analysis of start.mjs source — the same pattern as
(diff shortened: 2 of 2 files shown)


bTimor:
Hey @mattst88! 👋 Thanks for tracking this down — the diagnosis is correct, the fix placement is right, and the ESM-resolution reasoning is sound.

Two things needed before we can merge:

**Required: wrap the restore in try/catch**

```js
// Before:
process.chdir(originalCwd);

// After:
try { process.chdir(originalCwd); } catch { /* best effort — restore cwd for VTE tab inheritance */ }
```

Without this, if `originalCwd` no longer exists at startup time (network FS, container, temp dir cleaned up), the server crashes on launch — which is strictly worse than the original wrong-tab issue. Every other potentially-failing op in `start.mjs` follows the `try { } catch {}` defensive pattern; this should too.

**Recommended: add a Vitest test**

The `VITEST` guard pattern in `start.mjs` (`if (!process.env.VITEST)`) already exists to make the startup sequence testable. A minimal test verifying that `process.cwd()` after the setup phase equals the pre-launch directory would prevent this from silently regressing. See `tests/server-stdin-eof-exit.test.ts` for the pattern.

The fix logic is solid — just these two changes and this is good to go. Thanks again!

mattst88:
Done, thanks!

Issue #901, opened 2026-07-01
[Bug]: \upgrade\ only updates one of two parallel install trees on Windows; \doctor\ reads the other


RGalli:
[Bug]: `upgrade` only updates one of two parallel install trees on Windows; `doctor` reads the other

## Platform
OpenCode (Windows, standalone MCP mode, `CONTEXT_MODE_PLATFORM=opencode`)

## context-mode version
1.0.169 (current latest at the time of filing). Bug present in all versions I tested during the upgrade path 1.0.166 → 1.0.169.

## Debug script output (REQUIRED)

```json
{
  "version": "2.0.0",
  "generated": "2026-07-01T15:09:39Z",
  "sections": {
    "1. System Info": {
      "info": {
        "OS type": "windows",
        "Architecture": "x86_64",
        "Windows ver": "Microsoft Windows [versão 10.0.26200.8737]",
        "Shell": "/usr/bin/bash",
        "Bash version": "5.3.15(1)-release"
      }
    },
    "2. Runtime Versions": {
      "info": {
        "Node.js": "v24.18.0",
        "Bun": "1.3.14",
        "npm": "11.16.0",
        "npm global root": "C:\\Users\\ruan_\\AppData\\Roaming\\npm\\node_modules"
      },
      "warnings": [
        "Codex ≥0.118.0: exec-mode MCP broken (openai/codex#16685). Pin to ≤0.116.0 for exec-mode."
      ]
    },
    "3. context-mode Installation": {
      "checks": [
        { "pass": true, "label": "build/ directory exists" },
        { "pass": true, "label": "hooks/pretooluse.mjs exists" },
        { "pass": true, "label": "hooks/sessionstart.mjs exists" },
        { "pass": true, "label": "server.bundle.mjs exists" },
        { "pass": true, "label": "cli.bundle.mjs exists" }
      ],
      "info": {
        "Installed version": "1.0.169",
        "Plugin root": "C:/Users/ruan_/AppData/Roaming/npm/node_modules/context-mode",
        "Install method": "npm global/local",
        "npm latest": "1.0.169"
      }
    },
    "4. better-sqlite3 Native Module": {
      "checks": [
        { "pass": true, "label": "better-sqlite3 .node binary exists" },
        { "pass": false, "label": "require('better-sqlite3') succeeds" },
        { "pass": true, "label": "npm ignore-scripts is false" }
      ]
    },
    "5. Adapter Detection": {
      "info": {
        "Active adapter (env)": "none",
[...]

PR #930, opened 2026-07-08
fix(docs): avoid GraphQL issue listing

ousamabenyounes:
## What
Update the marketing ops issue query guidance to use REST search through `gh api` instead of `gh issue list`.

## Why
The `gh issue list` path can route through GitHub GraphQL fields affected by Projects classic deprecations. The REST search endpoint avoids that path for the open-issue marketing workflow.

## Coverage added
- Adds a regression test for the marketing ops guide.
- Locks the expected REST `search/issues` query shape.

## Test plan
- [x] `rtk test npx vitest run tests/ops/marketing-github-api.test.ts`
- [x] RED proof: reverted `.claude/skills/context-mode-ops/marketing.md`, reran the same focused test, and observed failure.
- [x] `rtk test npm run typecheck`

### DIFF
--- a/.claude/skills/context-mode-ops/marketing.md
+++ b/.claude/skills/context-mode-ops/marketing.md
@@ -53,7 +53,7 @@ Before writing, read these files and use ONLY verified numbers:
 | Adapter count | `tests/adapters/` (count test files) |
 | GitHub stars | `gh api repos/mksglu/context-mode --jq '.stargazers_count'` |
 | GitHub forks | `gh api repos/mksglu/context-mode --jq '.forks_count'` |
-| Open issues | `gh issue list --state open --json number --jq 'length'` |
+| Open issues | `gh api 'search/issues?q=repo:mksglu/context-mode+is:issue+is:open' --jq '.total_count'` |
 | Recent release | `gh release list --limit 1` |
 
 If you cannot verify a number, do not use it.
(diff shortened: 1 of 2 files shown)


bTimor:
Hey @ousamabenyounes! 👋

The REST-vs-GraphQL change in the marketing skill looks fine as a standalone improvement. However, the `Fixes #901` claim needs to be removed — this PR does not address issue #901.

Issue #901 is a Windows filesystem problem: `cachePluginRoot()` in `src/cli.ts` resolves to `%LOCALAPPDATA%/<platform>/packages/...` on Windows, while `OpenCodeAdapter.getInstalledVersion()` in `src/adapters/opencode/index.ts` uses `homedir()/.cache/<platform>/node_modules/...` (old pre-2024 layout, wrong path on Windows). After a `ctx-upgrade`, the upgrade writes to one tree and `ctx-doctor` reads from the other — permanently diverged.

Could you either:
1. Remove the `Fixes #901` reference and let this merge as a standalone marketing improvement, or
2. Add the actual path-unification fix in `src/adapters/opencode/index.ts` to close #901 properly

Thanks!

ousamabenyounes:
Removed the incorrect `Fixes #901` claim; this PR is now scoped as a standalone marketing workflow improvement.

Issue #925, opened 2026-07-06
Shell execution wrapper breaks commands that start with compound shell syntax


aaamosh:
### Summary

When `context-mode` executes shell commands through the batch execution path, the generated wrapper appears to prepend `NODE_OPTIONS=...` directly before the user command. That works for simple commands, but it is invalid shell syntax when the command starts with a compound construct such as `if`, `for`, or a subshell group.

The examples below are fully anonymized and use placeholders only.

### Environment

- context-mode version: 1.0.169
- client/plugin surface: Codex plugin
- OS observed on: Linux Mint 22.3

### Minimal examples

These commands are valid when run by a normal shell:

```sh
if command -v example-tool >/dev/null; then example-tool --version; fi

for x in /tmp/example/*; do [ -x "$x" ] && "$x" --version; done

(printf "one\n"; printf "two\n") | sed -n "1,10p"
```

When routed through the context-mode shell execution path, they fail with syntax errors similar to:

```text
/tmp/.ctx-mode-XXXX/script.sh: line 2: syntax error near unexpected token `then`
/tmp/.ctx-mode-XXXX/script.sh: line 2: `NODE_OPTIONS='--require /tmp/cm-fs-preload-XXXX.js' if command -v example-tool >/dev/null; then example-tool --version; fi`

/tmp/.ctx-mode-XXXX/script.sh: line 2: syntax error near unexpected token `do`
/tmp/.ctx-mode-XXXX/script.sh: line 2: `NODE_OPTIONS='--require /tmp/cm-fs-preload-XXXX.js' for x in /tmp/example/*; do [ -x "$x" ] && "$x" --version; done`

/tmp/.ctx-mode-XXXX/script.sh: line 2: syntax error near unexpected token `(`
/tmp/.ctx-mode-XXXX/script.sh: line 2: `NODE_OPTIONS='--require /tmp/cm-fs-preload-XXXX.js' (printf "one\n"; printf "two\n") | sed -n "1,10p"`
```

### Expected behavior

Shell commands that begin with valid shell compound constructs should either execute successfully or be wrapped in a way that preserves shell grammar.

For example, instead of generating something equivalent to:

```sh
NODE_OPTIONS=... if ...; then ...; fi
```

the wrapper could export the environment variable separately before the command, or execute the user command via an inner shell with the environment already set.

### Actual behavior

[...]

PR #933, opened 2026-07-08
fix(batch): export NODE_OPTIONS as a statement so compound shell comm…

hakanbaysal:
…ands work (#925)

The POSIX branch of buildBatchNodeOptionsPrefix emitted an inline `NODE_OPTIONS='...' <command>` assignment. That form is only valid before a *simple* command, so any batch command starting with a compound construct (`if`, `for`, `while`, a `(...)` subshell, or a `{ ...; }` group) was rewritten into invalid shell and failed with `syntax error near unexpected token`. The failed run was then indexed, surfacing the syntax error instead of the intended output.

Emit a standalone `export NODE_OPTIONS='...'; ` statement instead. This mirrors the PowerShell (`$env:...; `) and cmd (`set "..." && `) branches, which already use a separate statement, and composes with any following command. As a bonus the exported value now also reaches `node` invocations *inside* compound constructs (e.g. a `for` loop body), which the inline form could not do.

Updates the POSIX prefix test and adds real-compound-command coverage.

## Affected platforms

- [ ] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Checklist

- [ ] `npm test` passes

<details>

</details>

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -1466,7 +1466,12 @@ export function buildBatchNodeOptionsPrefix(shellPath: string, preloadPath: stri
     return `set "NODE_OPTIONS=${option.replace(/"/g, '""')}" && `;
   }
 
-  return `NODE_OPTIONS=${quotePosixSingle(option)} `;
+  // Emit a standalone `export` statement rather than an inline `VAR=val cmd`
+  // assignment: the inline form is only valid before a *simple* command, so it
+  // breaks when the command starts with a compound construct (`if`, `for`,
+  // `(...)`, `{ ...; }`). The `; ` separator mirrors the PowerShell/cmd
+  // branches above and keeps the assignment a separate statement (#925).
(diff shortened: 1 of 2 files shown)


bTimor:
Hey @hakanbaysal! 👋

This is a duplicate of #934 — same head branch, but targeting `main` directly. All fixes go through `next` first per our branching convention.

Closing in favour of #934. Thanks for the fix!

hakanbaysal:
Thanks — yes, this one was opened against `main` by mistake. #934 (targeting `next`) is the canonical PR. Closing.

Issue #817, opened 2026-06-11
PreToolUse/PostToolUse hooks intercept small Bash/WebFetch calls without a size threshold (Claude Code) — unlike Gemini CLI's size-aware BeforeTool matcher


monossido:
**Environment:** context-mode v1.0.162, Claude Code (plugin install), Linux

## Description

On Claude Code, the `PreToolUse`/`PostToolUse` hooks fire on every `Bash` invocation and every `WebFetch`/`curl`/`wget` call regardless of expected output size, redirecting the model toward `ctx_batch_execute` / `ctx_execute_file` / `ctx_fetch_and_index` even for outputs of a few hundred bytes to a few KB.

This contrasts with the Gemini CLI integration, whose README explicitly restricts the `BeforeTool` matcher to "tools that produce large output... avoids unnecessary hook overhead on lightweight tools."

## Evidence (3 A/B tests on a ~137K LOC backend repo)

1. `git log --stat -30` (~6-7KB raw output) via plain Bash vs `ctx_batch_execute` (same command + 2 queries): the batch call returned **18.1KB** to context — *more* than the direct command — while padding the "without context-mode" baseline by +36KB (largely duplicated content across the two queries). The reported "51% kept out" ratio is an artifact of that inflated baseline.

2. Direct `Read` (offset/limit) of a 41-line function (~870B): **zero** context-mode overhead, zero captures — correct, matches the tool's own "small file → Read directly" guidance.

3. Same 41-line function via `ctx_execute_file`: added **+1KB** to context (echoes the executed snippet plus the result), added a "sandbox" capture, and the "kept out" ratio actually *dropped* 51%→50%.

4. `curl`/`wget` to fetch a small README is hard-blocked with no size check, forcing `ctx_fetch_and_index` even when a direct fetch would return a small, immediately-useful payload.

## Suggested fix

Add a size-aware threshold to the Claude Code hook config (e.g., only intercept `Bash`/`WebFetch` when expected/observed output exceeds ~X KB), mirroring the Gemini CLI `BeforeTool` matcher design. This would let small commands and small file reads pass through untouched, reserving the `ctx_*` sandbox tools for genuinely large outputs where they deliver the documented savings.

🤖 Filed with [Claude Code](https://claude.com/claude-code)

mksglu:
[...]

Issue #843, opened 2026-06-18
[Feature]: Env-var opt-out for the main-session WebFetch deny (e.g. CONTEXT_MODE_ALLOW_WEBFETCH=1)


rrva:
## Summary

The `PreToolUse` `WebFetch` branch hard-denies WebFetch in the **main session** and redirects to `ctx_fetch_and_index` (`hooks/core/routing.mjs` ~L800, `action: "deny"`). This is intentional and valuable as a default, but there is currently **no first-class way to opt out** of just this interception while keeping the rest of context-mode enabled. I'd like to request an env-var opt-out, e.g. `CONTEXT_MODE_ALLOW_WEBFETCH=1`, checked in the `canonical === "WebFetch"` branch and returning pass-through (`null`) when set.

## Environment

- context-mode **1.0.162** (plugin via marketplace), Claude Code CLI, macOS
- MCP server running normally in the main session (`ctx_doctor` all green)

## Why this is needed

Confirmed (source + DeepWiki) that no env var, config file, or per-tool allowlist/denylist disables the WebFetch deny — the routing is hardcoded. The only way to restore WebFetch today is to edit the plugin-cache `hooks/hooks.json` (drop the `WebFetch` matcher) or patch the `routing.mjs` branch to `return null`. Both work, but live under the install-cache path and are **silently overwritten on every plugin update**, so they aren't durable.

There are legitimate reasons to want WebFetch directly in the main session:
- WebFetch's built-in prompt-based extraction (fetch + summarize in one model pass) is a different workflow from `ctx_fetch_and_index` + `ctx_search`; for small/one-off pages the index round-trip is more overhead than the bytes saved (overlaps with #817).
- Users who have their own context discipline and just want the native tool back, without losing the Bash/Read/Grep nudges, the `ctx_*` sandbox tools, or session capture.

## Precedent

This is the same shape already requested for a different hook branch in **#832** (`CONTEXT_MODE_DISABLE_AGENT_INJECTION=1` for the `Agent`-injection branch). A symmetric `WebFetch` flag would be consistent with that, and with the existing env-var surface (`CONTEXT_MODE_REQUIRE_SECURITY`, `CONTEXT_MODE_PLATFORM`, `CONTEXT_MODE_DEBUG`, …).

## Relationship to existing issues

- **#817** (open) — proposes a *size threshold* for Bash/WebFetch interception. Related, but a binary opt-out is a simpler, orthogonal escape hatch (and useful even if a threshold lands).
[...]

PR #945, opened 2026-07-10
fix(routing): let claude.ai Artifact URLs pass through WebFetch (#938)

### LINKED ISSUE #938: WebFetch PreToolUse redirect has no exception for claude.ai/code/artifact URLs — ctx_fetch_and_index only ever gets the empty unauthenticated SPA shell
## Summary

The `WebFetch` PreToolUse hook (`hooks/core/routing.mjs`, `canonical === "WebFetch"` branch, ~line 874) unconditionally denies and redirects **every** `WebFetch` call to `ctx_fetch_and_index`, with no exception for `claude.ai/code/artifact/*` URLs.

Claude Code's native `WebFetch` tool has a documented, built-in exception for these URLs: they are fetched using the caller's authenticated claude.ai session (the tool description explicitly says so — "claude.ai/code/artifact/{uuid} URLs ... ARE fetchable — WebFetch uses your claude.ai login"). This is necessary because an Artifact page is a client-rendered SPA shell gated behind auth; the actual artifact HTML is injected client-side after an authenticated handshake.

`ctx_fetch_and_index` does a plain, unauthenticated HTTP fetch. For these URLs it can only ever retrieve the empty SPA shell (~0.1KB: `<!doctype html>...Claude Artifact...Content is user-generated and unverified.`), never the real content. So the blanket redirect doesn't just add overhead here (as in #817) — it makes the tool call **strictly worse than useless**: the model gets back a plausible-looking but empty page and has no way to know it's empty-by-design vs. a real 0-byte artifact.

## Environment

- context-mode 1.0.169 (plugin via marketplace)
- Claude Code CLI, macOS
- Reproduced directly against `hooks/pretooluse.mjs` (not just observed via the model)

## Reproduction

```bash
cd <plugin-root>/hooks
echo '{"tool_name":"WebFetch","tool_input":{"url":"https://claude.ai/code/artifact/<any-uuid-you-own>","prompt":"describe this page"}}' | node pretooluse.mjs
```

Result: `permissionDecision: "deny"`, redirecting to `ctx_fetch_and_index`. Following that advice:

```
ctx_fetch_and_index(url: "https://claude.ai/code/artifact/<uuid>", source: "...")
```

returns only:

```
Claude Artifact
[](/code/artifacts) Content is user-generated and unverified.
```

— the SPA shell, not the artifact's actual HTML — because the fetch has no claude.ai auth cookie/session. Meanwhile the *native* `WebFetch` tool, unblocked, correctly returns the full rendered artifact content (verified by patching around the hook locally).

## Root cause

`hooks/core/routing.mjs`, WebFetch branch:

```js
if (canonical === "WebFetch") {
  const url = getWebFetchUrl(toolInput);
  return mcpRedirect({
    action: "deny",
    reason: `context-mode: WebFetch redirected. Call ${t("ctx_fetch_and_index")}(...)`,
    ...
  }, mcpToolsAvailable);
}
```

No URL-pattern check exists before the deny — every `WebFetch` call is treated identically, regardless of whether the target requires the caller's authenticated session (which only native `WebFetch` has access to; `ctx_fetch_and_index` is a bare `fetch()` in a sandboxed process with no session cookies).

Related but distinct existing reports on the same root issue (blanket, unconditional WebFetch interception with no escape hatch):
- #794 — blocked subagents can't reach `ctx_*` tools, so the redirect advice is unusable in that context
- #817 — no size threshold, so even tiny outputs get redirected
- #165 — closed feature request for a JS-rendering fallback in `fetch_and_index` generally

This report is narrower: it's not about output size or subagent tool access, it's about a URL class where the redirect target is *categorically incapable* of succeeding (no auth), while the tool being bypassed has a documented working path for exactly this case.

## Suggested fix

Add a passthrough exception before the deny, mirroring native WebFetch's own documented exception:

```js
if (canonical === "WebFetch") {
  const url = getWebFetchUrl(toolInput);
  if (/^https:\/\/(preview\.)?claude\.ai\/code\/artifact\//.test(url)) {
    return null; // let native WebFetch handle it — needs the caller's authenticated session
  }
  const url = getWebFetchUrl(toolInput);
  return mcpRedirect({ ... });
}
```

I patched this locally in `hooks/core/routing.mjs` and confirmed via the same `pretooluse.mjs` stdin repro above: the artifact URL now passes through untouched (no hook output / no deny), while a normal URL (`https://example.com`) is still correctly redirected. Happy to open a PR with this exact change if useful — it's a 4-line, narrowly-scoped addition with no effect on any other URL.

tongliuTL:
## What / Why / How

Fixes #938.

**What:** The `WebFetch` PreToolUse hook (`hooks/core/routing.mjs`) unconditionally denied and redirected every `WebFetch` call to `ctx_fetch_and_index`, with no exception for `claude.ai/code/artifact/{uuid}` URLs.

**Why:** Artifact pages are a client-rendered SPA shell gated behind the caller's authenticated claude.ai session. Native `WebFetch` has a documented exception for these URLs — it fetches them using that session — while `ctx_fetch_and_index` does a bare unauthenticated fetch and can only ever retrieve the empty shell (~0.1KB, no real content, confirmed via the repro in #938). So the redirect here isn't just overhead like #817 — it makes the tool call strictly worse than useless: the agent gets back a plausible-looking but permanently empty page with no signal that it's empty-by-design vs. a genuinely empty artifact.

**How:** Added `isClaudeArtifactUrl(url)`, parsed via the `URL` constructor rather than a substring/regex check on the raw string (so a lookalike host like `claude.ai.evil.com` can't slip through), and let `WebFetch` pass through untouched (`return null`) for `claude.ai` / `preview.claude.ai` `/code/artifact/` URLs specifically, before the existing deny+redirect logic. Every other URL — including other `claude.ai` paths — keeps redirecting exactly as before; verified with `git diff` scoped to just these two lines of new branching logic.

## Affected platforms

- [x] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

(`WebFetch` is Claude Code's native tool name; this hook path is specific to that adapter's tool surface.)

## Test plan

Added 5 new cases to the existing `describe("WebFetch tool", ...)` block in `tests/hooks/core-routing.test.ts`:
- `claude.ai/code/artifact/...` passes through untouched (`result === null`)
- `preview.claude.ai/code/artifact/...` passes through untouched
- a non-artifact `claude.ai` URL (e.g. `/settings`) still redirects
- a lookalike host (`claude.ai.evil.com/code/artifact/...`) still redirects — regression guard on the URL-parsing approach
- a plain-`http://` artifact URL still redirects (the exception is HTTPS-only)

Ran locally: full suite `npx vitest run` → 4718 passed, 0 failed, 29 skipped (no regressions). `npx tsc --noEmit` → no errors.

## Checklist

- [x] `npm test` passes
- [ ] Docs updated if needed (README, platform-support.md) — n/a, this is an internal routing exception, not user-facing config
- [x] No Windows path regressions (forward slashes only) — change is pure URL parsing, no filesystem paths touched

### DIFF
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -641,6 +641,26 @@ function getWebFetchUrl(toolInput) {
   return "";
 }
 
+// claude.ai Artifact pages (#938) are a client-rendered SPA shell gated
+// behind the caller's authenticated claude.ai session. Native WebFetch has
+// a documented exception for these URLs — it fetches them using that
+// session — while ctx_fetch_and_index does a bare unauthenticated fetch and
+// can only ever retrieve the empty shell (~0.1KB, no real content). Parsed
+// via the URL constructor (not a substring/regex check on the raw string)
+// so a lookalike host like claude.ai.evil.com can't slip through.
+function isClaudeArtifactUrl(url) {
+  if (typeof url !== "string" || url === "") return false;
+  let parsed;
+  try {
+    parsed = new URL(url);
+  } catch {
+    return false;
+  }
+  if (parsed.protocol !== "https:") return false;
+  if (parsed.hostname !== "claude.ai" && parsed.hostname !== "preview.claude.ai") return false;
+  return parsed.pathname.startsWith("/code/artifact/");
+}
+
 function getCodexConfigDir(env = process.env) {
   const codexHome = env.CODEX_HOME;
   if (codexHome && codexHome.trim() !== "") return resolve(codexHome);
@@ -874,6 +894,10 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
   // ─── WebFetch: deny + redirect to sandbox ───
   if (canonical === "WebFetch") {
     const url = getWebFetchUrl(toolInput);
+    // Let claude.ai Artifact URLs through untouched — see isClaudeArtifactUrl (#938).
+    if (isClaudeArtifactUrl(url)) {
+      return null;
+    }
     return mcpRedirect({
       action: "deny",
       reason: `context-mode: WebFetch redirected. Call ${t("ctx_fetch_and_index")}(url: "${url}", source: "...") to fetch + index the page, then ${t("ctx_search")}(queries: [...]) to query the indexed content — the raw page bytes stay in storage instead of entering your conversation. Or call ${t("ctx_execute")}(language, code) when you want to derive your answer in one round trip (parse, extract, count) without persisting the response. Both have full network access. Retry the same call on a transient DNS error (EAI_AGAIN, ETIMEDOUT, ENETUNREACH).`,
--- a/tests/hooks/core-routing.test.ts
+++ b/tests/hooks/core-routing.test.ts
@@ -395,6 +395,43 @@ describe("routePreToolUse", () => {
       expect(result!.reason).toContain(url);
     });
 
+    it("passes claude.ai Artifact URLs through untouched (#938)", () => {
+      const url = "https://claude.ai/code/artifact/51de1b8d-61cb-488e-a964-547ef217f5e9";
+      const result = routePreToolUse("WebFetch", { url });
+      // Native WebFetch fetches these using the caller's authenticated
+      // claude.ai session; ctx_fetch_and_index cannot (no session cookies)
+      // and would only ever get back the empty SPA shell.
+      expect(result).toBeNull();
+    });
+
+    it("passes preview.claude.ai Artifact URLs through untouched (#938)", () => {
+      const url = "https://preview.claude.ai/code/artifact/51de1b8d-61cb-488e-a964-547ef217f5e9";
+      const result = routePreToolUse("WebFetch", { url });
+      expect(result).toBeNull();
+    });
+
+    it("still redirects non-artifact claude.ai URLs (#938)", () => {
+      const result = routePreToolUse("WebFetch", { url: "https://claude.ai/settings" });
+      expect(result).not.toBeNull();
+      expect(result!.action).toBe("deny");
+    });
+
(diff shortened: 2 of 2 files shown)


tongliuTL:
@mksglu please kindly review my commits and let me know what you think.

roeex5:
Independent confirmation on context-mode 1.0.151: the WebFetch deny at `hooks/core/routing.mjs:800` is unconditional, and `ctx_fetch_and_index` returns only a ~0.1 KB SPA shell for `claude.ai/code/artifact/*` since it has no authenticated session.

This isn't only a read problem — it breaks **writes** too. Updating a shared artifact requires reading the current version to merge onto, so without a passthrough the only options are force-publishing a stale copy or not editing at all. We hit exactly that: a stale local copy would have clobbered a teammate's changes to a shared status board, and only the server-side version-conflict check caught it.

Applied this PR's logic locally as a patch and verified both directions:

- `https://claude.ai/code/artifact/<uuid>` → hook emits nothing (passthrough), WebFetch runs and returns real content
- `https://example.com` → still returns the `deny` redirect

Two housekeeping notes for whoever triages: #984 and #1006 look like duplicates of #938, and #843 was closed as *completed* proposing `CONTEXT_MODE_ALLOW_WEBFETCH=1`, but no such env var exists anywhere in the shipped 1.0.151 tree — so there is currently no supported escape hatch and users are patching the plugin cache by hand (which any version upgrade silently discards).

Issue #962, opened 2026-07-16
RFC: bound trigram and vocabulary growth for logs, JSONL, and high-entropy sources


Tomodad:
## Problem

Large logs and JSONL are duplicated into porter and trigram FTS, while insert-only vocabulary accumulates paths, UUIDs, hashes, escaped payloads, base64-like strings, and other high-entropy tokens. A 4096-byte row invariant such as #878 does not control this storage growth, and #947's timeout boundary does not define indexing policy.

## Proposed design

- source-shape classifier for log/JSONL/dense machine output;
- porter-only default above a configurable size/entropy threshold;
- trigram indexing only by explicit policy/opt-in for large sources;
- vocabulary filters for UUID/hash/base64/very long path/high-entropy tokens;
- per-source and global unique-token budgets;
- metrics for porter bytes, trigram bytes, vocabulary rows, and dropped-token reasons;
- source deletion GC or an explicit vocabulary rebuild command.

## Open questions

- Should the policy be based on source type, byte size, entropy, or a combination?
- Should full raw content be stored outside FTS and only bounded windows/structured fields be indexed?
- Should deleting a source synchronously garbage-collect vocabulary, or should vocabulary be rebuilt periodically?

## Acceptance tests

- large JSONL stays searchable by porter while skipping trigram;
- ordinary documentation retains trigram substring search;
- hashes, UUIDs, and base64-like tokens do not consume vocabulary budget;
- deleting/rebuilding a source removes obsolete vocabulary according to the selected GC model;
- every policy decision is visible rather than silent.

mradwankhalil:
Field data for this RFC, from a production store that hit exactly the failure mode described. Posting because the proposal is sound but has no measured numbers attached yet, and one consequence of unbounded vocabulary isn't captured here.

**Environment:** context-mode 1.0.169, Windows 11, OpenCode + OpenChamber. Measurements taken read-only via `node:sqlite` (Node v26.7.0) with all harness processes stopped, so nothing was competing for the file.

## The store

| metric | value |
|---|---|
| `.db` | 2,825,302,016 B (2.83 GB) |
| `sources` | 550 |
| `chunks` / `chunks_trigram` rows | 14,726 each |
| actual indexed content | 509,631,192 B (486 MB) |
| `vocabulary` rows | **3,631,386** |
| **avg vocabulary token length** | **28.7 chars** |
| **max vocabulary token length** | **12,483 chars** |

`dbstat` page attribution:

| object | size |
|---|---|
[...]

Issue #961, opened 2026-07-16
RFC: add indexed byte/chunk budgets and separate batch display from indexed input


Tomodad:
## Problem

`ctx_batch_execute` can capture tens to hundreds of MiB, assemble display Markdown, and persist the full payload into porter, trigram, and vocabulary. A per-row cap such as #878 does not bound total indexed data and can turn one 100 MiB payload into tens of thousands of valid-sized chunks.

This is separate from #947: a whole-handler deadline limits wall-clock exposure, but it does not define how much command output should be indexed by default.

## Proposed policy surface

- `max_bytes_per_command`
- `max_total_indexed_bytes`
- `max_generated_chunks`
- explicit opt-in for larger ingestion
- streaming byte/chunk counters instead of first building one giant JS string
- no silent truncation: return a clear partial/not-indexed diagnostic

Separate output presentation from indexing:

- Markdown remains the tool display format;
- command label, exit status, and duration become structured metadata;
- each stdout/stderr body is indexed as plain/structured text rather than concatenated display Markdown.

## Initial defaults for benchmark discussion, not final policy

- 8 MiB per command
- 16 MiB per batch

These values should not be fixed without representative benchmarks.

## Acceptance tests

- one command crosses the per-command byte budget;
- many individually small commands cross the total byte budget;
- generated chunks cross the chunk budget;
- no partial payload is presented as complete;
- explicit opt-in succeeds;
- display Markdown and indexed plain input remain semantically linked;
- budget decisions and dropped/indexed byte counts are visible in the tool response and metrics.

Issue #947, opened 2026-07-10
[Bug]: ctx_batch_execute timeout does not bound indexing/search, so Claude Code agents can hang for hours


stgmt:
## Summary

`ctx_batch_execute` accepts a `timeout` parameter, but in practice it does not bound the whole MCP handler. In Claude Code, I observed `ctx_batch_execute` calls with `timeout: 120000` remain in-flight for more than 4 hours after the command phase, apparently during context-mode indexing/search/formatting work against a bloated SQLite/FTS content DB.

## Environment

- context-mode: **1.0.169**
- Claude Code: **2.1.202**
- OS: Windows 11
- storage: default `~/.claude/context-mode/content`

## Evidence from real Claude Code sessions

Three background agents were stuck around context-mode calls for hours:

```text
agent-aee8e6d09a69e41c7 | ctx_search        | 2026-07-10T15:01:51Z -> 18:50:37Z | 228.8 min
agent-a5ba93371b699a949 | ctx_batch_execute | 2026-07-10T14:33:39Z -> 18:50:42Z | 257.1 min | call had timeout: 120000
agent-ae18a471c25b61a57 | ctx_batch_execute | 2026-07-10T15:02:27Z -> 18:51:51Z | 229.4 min
```

The project content DB at the time was heavily bloated:

```text
content DB: 645,128,192 bytes
WAL:        349,454,312 bytes
```

Largest chunks persisted in `chunks`:

```text
36,754,737 bytes | tracing and runtime observability evidence (2)
36,163,374 bytes | Visual and post-render artifact terminology (2)
35,988,924 bytes | Full current unstaged diff indexed for targeted queries
```

These huge chunks are related to #878, but this issue is specifically about timeout semantics: even if #878 is fixed, the tool should not be able to keep a Claude Code agent blocked forever.

## Why `timeout` is misleading today

From source inspection, `resolveExecTimeout` may omit a timeout for Claude Code unless explicitly provided. When `ctx_batch_execute` does get a timeout, it appears to apply to command execution (`runBatchCommands`) but not necessarily to the later phases:

- store/indexing
- FTS/trigram search
- fallback search
- result formatting

Those later phases can be synchronous SQLite work. A JavaScript timer around only the command runner cannot interrupt a blocked synchronous DB/index/search path.

## Expected

A `ctx_batch_execute` call with `timeout: 120000` should either finish or return a timeout error in roughly 120 seconds. It should not remain in-flight for hours.

## Actual

[...]

PR #963, opened 2026-07-16
fix(store): bound FTS search result hydration for oversized rows

Tomodad:
## Summary

Refs #947.

FTS ranking currently selects full `chunks.content` and materializes `highlight()` before the server clips output. A legacy 53 MiB row can therefore turn a millisecond MATCH into a multi-minute synchronous highlight operation.

This PR keeps the scope to the search path:

- rank porter/trigram candidates using bounded metadata only (`rowid`, title, type, timestamp, source, rank, session id);
- preserve the historical full content + FTS5 highlight result for normal rows at or below `MAX_CHUNK_BYTES`;
- preserve ranked-row identity on Node 22/better-sqlite3 using an FTS5-compatible rowid constraint;
- hydrate legacy oversized rows with a match-position lookup and a bounded `substr()` window;
- preserve source/content-type filters, session attribution, BM25, RRF, and result markers;
- rebuild the tracked server/CLI bundles.

It deliberately does **not** implement Markdown chunking (#878/#898), ingestion budgets (#961), trigram/vocabulary policy (#962), or a whole-handler deadline/worker model.

## TDD

RED on `next` after adding the regression tests:

```text
PASS (0) FAIL (3) skipped (153)
porter content:   1,200,022 chars
trigram content:  1,200,031 chars
filtered content: 1,200,025 chars
```

Node 22.5 CI exposed a second RED after the first push:

```text
existing intent-search scenarios: 3 failed
ranked metadata rowids: distinct
hydrated content: first MATCH row repeated for porter and trigram
```

A focused multi-row regression now asserts that each ranked porter/trigram candidate hydrates its own row.

GREEN:

```text
tests/core/search.test.ts: 157 passed
tests/core/search.test.ts + tests/store.test.ts: 279 passed, 1 skipped
search/highlight/snippet server subset: 26 passed
npm run typecheck: passed
npm run build: passed (bundle and asymmetric-drift assertions included)
git diff --check: passed
```

A complete `tests/core/server.test.ts` run in the managed Windows sandbox also exposed three pre-existing process/path tests that fail with `STACK_TRACE_ERROR`/`EPERM`; the directly affected search/formatting subset is green.

## Before/after: isolated 1.2 MiB legacy row

| path | before | after | returned content/highlight |
|---|---:|---:|---|
| porter | 20.445 ms | 11.078 ms | 1,200,022/1,200,024 -> 315/317 chars |
| trigram | 30.406 ms | 9.248 ms | 1,200,022/1,200,024 -> 315/317 chars |
| fallback | 49.629 ms | 17.163 ms | 1,200,022/1,200,024 -> 315/317 chars |

Node 22.5 after the rowid-compatibility follow-up: porter 4.977 ms, trigram 4.337 ms, fallback 8.046 ms; all still returned 315/317 chars with the match preserved.

## Compatibility

Normal capped chunks retain full-content/highlight behavior. Only legacy rows that already violate `MAX_CHUNK_BYTES` return a bounded window. Proximity reranking for those legacy rows therefore uses the window rather than the full document, which can change ordering among multiple giant rows. The metadata-first + hydration path adds a small DB round-trip cost for normal rows (~0.045 ms/search in a local 500-iteration microbenchmark).


Tomodad:
Fresh Windows/Codex recurrence on the official marketplace build `v1.0.169` (2026-07-26):

- `ctx_batch_execute` hit the host error `timed out awaiting tools/call after 300s` during bounded local inspection.
- A subsequent `ctx_execute`, limited to walking the installed plugin's `src` tree, also hit the same 300s host timeout.
- After switching away from CTX, bounded native RTK/PowerShell probes completed in seconds.
- The active content store was large (`a20afb028fd78de8.db`: 712,667,136 bytes; WAL: 700,503,032 bytes), but as noted in #947, database size alone is not causal. Without phase watchdogs, I cannot attribute this fresh incident specifically to the full-row `highlight()` path.

This is fresh evidence that `v1.0.169` can still expose users to a 300-second blocking failure. PR #963 remains the focused fix for the confirmed oversized-row hydration path; broader whole-handler deadline/worker/backpressure work still belongs in #947.

All five checks on this PR are green and GitHub currently reports it clean/mergeable. @mksglu, would you prefer any further split/rebase or an additional regression fixture before review?

PR #966, opened 2026-07-17
fix: make concurrent SQLite startup resilient

SSS135:
## Summary

- Retries WAL, schema, and migration startup work after transient SQLite busy or locked errors.
- Makes destructive legacy migrations atomic under concurrent initializers.
- Distinguishes benign migration races from real failures.
- Closes database handles on every constructor failure path to prevent Windows lock leaks.

## Design

This preserves ADR-0001. WAL mode, the busy timeout, and bounded retries remain the concurrency mechanism. The change does not add lockfiles, exclusive mode, or a global writer guard.

## Validation

- `corepack pnpm exec vitest run tests/util/startup-locking.test.ts`: 11 passed.
- `corepack pnpm typecheck`: passed.
- `corepack pnpm build`: passed, including bundle and asymmetric-drift assertions.
- Standalone `corepack pnpm assert-bundle` and `corepack pnpm assert-asymmetric-drift`: passed.
- `git diff --check`: passed.


SSS135:
Pausing this for now. The branch, commits, and PR description are being kept so the work can be published again later.

Issue #985, opened 2026-07-22
Multi-session contention on shared per-project content DB → busy-wait CPU spin & event-loop hang (Windows)


alove20:
## Summary

On Windows, with multiple concurrent Claude Code sessions open in the **same project**, context-mode MCP server instances intermittently (a) pin a CPU core at ~100% accumulating hundreds of CPU-seconds, or (b) go completely unresponsive to `ctx_*` calls while sitting at low CPU. Both trace to **write contention on the shared per-project content DB**, amplified by a **synchronous busy-wait retry backoff**.

## Root cause

**1. The content DB is shared per-project, not per-session.**
Captures are stored in `~/.claude/context-mode/content/<projectHash>.db`, keyed by project. Every concurrent session working in the same project therefore opens and writes the *same* SQLite file. (Per-session DBs under `sessions/` are not contended — their WALs stay empty.) On one real machine this DB had grown to **386 MB with a 30 MB un-checkpointed WAL** (checkpoint starvation) with **5 live server instances** writing it.

**2. The `SQLITE_BUSY` backoff is a synchronous busy-wait.**
The retry wrapper (minified `Yr` in `server.bundle.mjs`) uses backoff steps `[100, 500, 2000]` ms, but implements the delay as a spin loop:

```js
function Yr(t, e=[100,500,2000]) {
  for (let r=0; r<=e.length; r++)
    try { return t(); }
    catch(o) {
      const s = o?.message ?? String(o);
      if (!s.includes("SQLITE_BUSY") && !s.includes("database is locked")) throw o;
      if (r < e.length) {
        const i = e[r], a = Date.now();
        for (; Date.now() - a < i; );   // <-- synchronous busy-wait: pins a core
      }
    }
  throw new Error("SQLITE_BUSY: database is locked after " + e.length + " retries...");
}
```

Each fully-contended operation spins a core for up to 100+500+2000 = ~2.6 s. Under a burst of contended writes this accumulates unbounded CPU. Observed: an instance at **99% of a core, 665 lifetime CPU-seconds**.

**3. better-sqlite3 is synchronous, so a blocked query freezes the event loop.**
[...]

Issue #560, opened 2026-05-13
No exclusive lock on SQLite database — multiple server instances cause unbounded WAL growth and query hangs


ishabana:
## Bug

The server does not guard against multiple instances opening the same SQLite database simultaneously. When more than one server process is running (e.g. after an upgrade that left the old process alive, or after a user registers the server both as a plugin and as a separate MCP server via npx), all instances write concurrently to the WAL file.

SQLite WAL auto-checkpoint requires no active readers to truncate. With multiple servers constantly reading and writing, the checkpoint window never opens. Every write from every instance just appends to the WAL indefinitely.

## Reproduction

1. Start two instances of the context-mode server pointing at the same data directory
2. Call any write tool (e.g. `ctx_index`) from both instances repeatedly
3. Observe the WAL file size growing without bound
4. Observe all queries progressively slowing until they hang

## Real-world result

On a system with 5 concurrent instances (2 from upgrades not cleaning up old processes, 3 from a user registering the server as both plugin and npx MCP server), the WAL grew to **238MB**. Every `ctx_search` call hung indefinitely — the server appeared completely unresponsive, with no error returned.

## Fix options

Either:

**Option A** — Exclusive SQLite lock on startup:
```js
db.pragma('locking_mode = EXCLUSIVE');
```
This causes the second instance to fail immediately with a clear error rather than silently competing.

**Option B** — PID lockfile:
Write a `.lock` file in the data directory on startup containing the current PID. On startup, check if the lockfile exists and if that PID is still alive. If so, refuse to start with a clear error message.

Option B is more user-friendly as it surfaces a descriptive error.

## Expected behaviour

Starting a second server instance against an already-open database should fail immediately with a clear error: `"Another context-mode server is already running (PID: XXXX). Stop it before starting a new instance."`

## Environment

- Platform: Windows 11
- Claude Code plugin
- context-mode v1.0.127
- SQLite WAL mode

mksglu:
Fixed in v1.0.128, just published. Thank you @ishabana — both fix proposals (Option A `locking_mode=EXCLUSIVE` + Option B PID lockfile) shipped as belt-and-braces. The 238MB WAL + 5-instance cause-effect chain made the failure mode obvious; without your data this would have stayed silent until the next user hit it harder.

## Fix architecture (v1.0.128)

**Two layers of defense**:

**Layer 1 — PID lockfile (your Option B)**
[...]

PR #987, opened 2026-07-22
fix(lifecycle): detect parent death on Windows via PID existence probe (#982)

### LINKED ISSUE #982: Windows: MCP server child orphans & CPU-spins on session end (parent-death guard is Linux/Bun-only)
## Summary

On Windows, the context-mode MCP server child process is **not reaped when its client (Claude Code / `claude.exe`) session ends, crashes, or the machine reboots**. The orphan survives and **busy-loops, accumulating unbounded CPU** (hundreds to thousands of CPU-seconds) while holding 170–490 MB RSS. Healthy instances sit at ~1–3 CPU-seconds, so a hung one is unmistakable. Over a multi-session workday the strays pile up, hang `ctx_*` calls, and contribute to system-wide memory/CPU pressure.

## Root cause

The parent-death guard already exists — but it never runs on Windows.

`start.mjs` (v1.0.169, lines ~62–135, added for issue #862) forwards stdin-EOF → SIGTERM → SIGKILL to the server child, which is exactly what prevents the orphan. But the entire block is gated behind:

```js
if (typeof globalThis.Bun === "undefined" && process.platform === "linux") { ... }
```

plus a discovered Bun binary. On Windows the MCP server runs as a bare `node …\cli.bundle.mjs` child with **no `process.stdin.on('end'|'close')` exit handler and no parent-liveness (ppid) poll**. When the parent goes away ungracefully, nothing tells the child to exit, so it orphans and pins a core.

(Windows also does not cascade-kill children on parent death unless the launcher uses a Job Object with `JOB_OBJECT_LIMIT_KILL_ON_JOB_CLOSE`, so the server needs its own guard for hard exits.)

## Environment

- Windows 11
- Claude Code desktop host `2.1.209`
- context-mode plugin **v1.0.169**

## Evidence (observed orphans, all `node …\context-mode\cli.bundle.mjs`)

| Started      | CPU-seconds | RSS    |
|--------------|-------------|--------|
| prior day    | ~4,323      | —      |
| —            | ~789        | 485 MB |
| —            | ~778        | 299 MB |
| —            | ~7,177      | 173 MB |
| —            | ~628        | 184 MB |

Healthy instances at the same times: ~1–3 CPU-seconds. Killing a stray immediately restored `ctx_*` responsiveness each time. Every live child had a `claude.exe` parent (one per active session); the leak is that *ended* sessions' children persist.

The sibling `token-savior` indexer shows the same "per-session helper not reaped" shape (seen at 145 / 116 CPU-s) — likely the same class of bug.

## Suggested fix

Ungate the #862 guard so it also runs on the Windows/node boot path. At minimum, at server startup:

```js
process.stdin.on('end',   () => process.exit(0));
process.stdin.on('close', () => process.exit(0));
const ppid = process.ppid;
setInterval(() => { try { process.kill(ppid, 0); } catch { process.exit(0); } }, 15000).unref();
```

and ensure the MCP transport's `onclose`/`close` tears the process down, with any keep-alive `setInterval`/reconnect loop `.unref()`'d and backed off. The stdin handlers are passive (they don't consume transport data) and safe to add alongside `StdioServerTransport`; the `.unref()`'d ppid poll is the belt-and-suspenders for hard exit / reboot where stdin EOF may not arrive.

alove20:
Fixes #982.

`startLifecycleGuard` runs on every platform (server.ts), but `makeDefaultIsParentAlive` (src/lifecycle.ts) only recognizes parent death two ways: the direct ppid *changing* (POSIX reparent-to-init), or the grandparent becoming PID 1 (a `ps` probe that returns `NaN` on Windows).

**On Windows neither fires.** A child is not reparented when its parent dies — `process.ppid` keeps returning the original (now-dead) value — so the equality check stays green and the `ps` grandparent probe is unavailable. Result: `isParentAlive()` returns `true` forever, the guard never calls `onShutdown`, and the MCP server orphans and lingers after the Claude Code session ends/crashes/reboots (pre-existing busy-wait then pinned a core — see #985).

**Change:** add a positive PID existence probe via `process.kill(ppid, 0)` (`ESRCH` ⇒ gone, `EPERM` ⇒ alive). This is the one death signal that works cross-platform, and it's strictly additive — POSIX keeps its existing faster paths. The probe is injectable (`isPidAlive`) for deterministic tests.

Adds a `#982` test proving a stable-ppid + NaN-grandparent instance is reaped once the parent PID disappears, and threads the injected probe through the existing #311 grandparent tests. Full build + lifecycle/stdin/pi-bridge suites pass.

### DIFF
--- a/src/lifecycle.ts
+++ b/src/lifecycle.ts
@@ -61,6 +61,32 @@ export interface IsParentAliveDeps {
   getPpid?: () => number;
   /** Read the grandparent ppid. Default: ps-based POSIX probe, NaN on Windows. */
   readGrandparentPpid?: () => number;
+  /**
+   * Positive existence probe for a PID. Default: {@link defaultIsPidAlive}
+   * (signal-0 via `process.kill`). This is the only parent-death signal that
+   * works on Windows, where an orphaned child keeps its original ppid (#982).
+   */
+  isPidAlive?: (pid: number) => boolean;
+}
+
+/**
+ * Cross-platform "does this PID still exist?" probe. Node maps signal 0 to an
+ * existence/permission check without delivering a signal: it throws `ESRCH`
+ * when the process is gone and `EPERM` when it exists but can't be signalled
+ * (treated as alive). Any other error is treated as alive to avoid a spurious
+ * reap. Unlike the ppid-equality and grandparent checks, this fires on Windows,
+ * where a parent's death does NOT reparent the child or change its ppid — so
+ * without this probe the guard's `isParentAlive()` returned `true` forever and
+ * the orphaned server never shut down (#982).
+ */
+export function defaultIsPidAlive(pid: number): boolean {
+  if (!pid || pid <= 1) return false;
+  try {
+    process.kill(pid, 0);
+    return true;
+  } catch (err) {
+    return (err as NodeJS.ErrnoException).code === "EPERM";
+  }
 }
 
 /**
@@ -81,6 +107,7 @@ export interface IsParentAliveDeps {
 export function makeDefaultIsParentAlive(deps: IsParentAliveDeps = {}): () => boolean {
   const getPpid = deps.getPpid ?? (() => process.ppid);
   const readGp = deps.readGrandparentPpid ?? readGrandparentPpidImpl;
+  const isPidAlive = deps.isPidAlive ?? defaultIsPidAlive;
   const originalPpid = getPpid();
   const originalGrandparentPpid = readGp();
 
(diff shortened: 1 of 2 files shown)


alove20:
Framing note for reviewers: this extends the zombie/orphan-elimination strategy that **ADR 0001** already relies on, to Windows.

ADR 0001 keeps the content store intentionally multi-writer and identifies the real cause of WAL bloat + `ctx_search` hangs (#560) as processes that never exit gracefully — so `closeDB()`'s `wal_checkpoint(TRUNCATE)` never fires and a lingering reader lock starves other checkpoints. On POSIX that zombie is eliminated by reparent-to-init detection (ppid change / grandparent→PID 1). On Windows neither signal exists: an orphaned child keeps its original ppid and `readGrandparentPpid` returns NaN, so `makeDefaultIsParentAlive()` returns true forever and the guard never reaps. This PR adds the one cross-platform death signal (`process.kill(ppid, 0)`), so the ADR's graceful-exit assumption finally holds on Windows too. No locking is introduced.

PR #990, opened 2026-07-23
fix: auto-index mid-size exec output for ctx_search without intent

niemst:
`ctx_execute` / `ctx_execute_file` output between 5KB and 100KB was returned raw inline but never indexed, so a later `ctx_search` could not find it — the mid-size gap between the 5KB intent threshold and the 100KB pointer threshold. Fix: index that range into the FTS5 content KB as a side effect at the 5KB threshold while still returning stdout inline (intent-search and large-output pointer paths unchanged), using distinct per-call source labels (`execute:<lang>:<hash>`) so successive execs don't clobber each other's index.

Test: added a regression test in `tests/core/server.test.ts` (~6KB no-intent exec returned raw inline AND retrievable via `ctx_search`); full suite green except one pre-existing macOS-locale test unrelated to this change. Build: `npm run build` passes (tsc + bundle + assert-bundle + assert-asymmetric-drift); bundles regenerated.


niemst:
Superseded by #991 — retargeted to next per CONTRIBUTING.

PR #1002, opened 2026-07-27
fix(pi): keep runtime context at system boundary

Tarun-joy:
## Summary

- inject Pi runtime routing, active memory, and resume context through `before_agent_start.systemPrompt`
- keep automation at Pi's host-owned model-only boundary
- remove the custom/user message injection path so runtime instructions cannot appear as user-authored chat

## Why

Pi's `display: false` only controls the interactive transcript. API consumers and adapters may still reinterpret custom messages as user messages. Runtime automation is system-owned context, so it belongs in the system prompt rather than in conversation history.

This intentionally favors transcript correctness and stable role semantics over the previous prompt-cache optimization.

## Validation

- `npx vitest run tests/pi-extension.test.ts` — 68 passed
- focused Resume/Routing tests — 7 passed
- tests assert no `context` handler is registered
- clean scoped diff

### DIFF
--- a/src/adapters/pi/extension.ts
+++ b/src/adapters/pi/extension.ts
@@ -743,8 +743,11 @@ export default function piExtension(pi: any): void {
       const ctx = _pendingContext;
       _pendingContext = "";
       event.messages.push({
-        role: "user",
+        role: "custom",
+        customType: "context-mode",
         content: ctx,
+        display: false,
+        details: { kind: "runtime_context" },
       });
       return { messages: event.messages };
     } catch {
--- a/src/adapters/pi/extension.ts
+++ b/src/adapters/pi/extension.ts
@@ -160,10 +160,6 @@ let _buildAutoInjection:
   | null
   | undefined = undefined;
 
-// Pending context to inject via the 'context' hook (avoiding systemPrompt mutation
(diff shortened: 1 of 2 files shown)


Tarun-joy:
Review verdict: **APPROVE**

The revised implementation uses Pi's canonical `before_agent_start` system-prompt result instead of relying on `display: false` for role correctness. This keeps context model-visible, prevents API/adapters from reclassifying automation as user chat, preserves the existing prompt prefix, and removes pending-message state plus the one-shot `context` hook.

Validated with the complete Pi adapter suite (68/68), focused Resume/Routing coverage (7/7), typecheck/build and bundle drift checks. The branch is clean, scoped to the Pi adapter and tests, current with upstream `main`, and mergeable. No blockers found.

Tarun-joy:
Revalidated after rebasing onto current upstream `main` (`da14dc1`).

- Pi adapter suite: **68/68 passed**
- Typecheck: passed
- Build + bundle integrity + asymmetric drift: passed
- Branch: clean and `MERGEABLE` / `CLEAN`
- Overlap rechecked: #864 still uses hidden custom messages, which Pi/API adapters can reclassify as user-role content; this PR keeps runtime automation at the canonical system boundary instead.

PR #1082, opened 2026-08-24
mcp: name execution timeout in milliseconds

### LINKED ISSUE #1081: Execution timeout units are absent from the public tool schema
ctx_execute, ctx_execute_file, and ctx_batch_execute expose a numeric timeout field whose runtime unit is milliseconds. The field name and public schema do not state that unit. A caller using timeout=120 for a two-minute operation receives a 120-millisecond deadline. The surrounding OMP MCP transport has a separate millisecond timeout, which makes the effective deadline harder to identify.

aktanazat:
Closes #1081

Renames the public execution deadline to timeoutMs for ctx_execute, ctx_execute_file, and ctx_batch_execute. The retired timeout field returns a bounded migration error instead of silently changing or dropping the deadline. Skill references are regenerated. The generated bundles are left out, since the bundle workflow rebuilds and commits them on push.

Tests: 6 live-server timeout tests, 646 focused tests after cleanup, 4727 full-suite passes with 21 skips, typecheck, build assertions, changed-line anti-slop scan.

### DIFF
--- a/skills/context-mode/references/anti-patterns.md
+++ b/skills/context-mode/references/anti-patterns.md
@@ -198,7 +198,7 @@ BAD — will timeout on API calls:
     const resp = await fetch('https://api.slow-service.com/data');
     console.log(await resp.json());
   language: javascript
-  timeout_ms: 5000  ← API may take 10+ seconds
+  timeoutMs: 5000  ← API may take 10+ seconds
 
 GOOD — generous timeout for network:
   Tool: execute
@@ -206,11 +206,11 @@ GOOD — generous timeout for network:
     const resp = await fetch('https://api.slow-service.com/data');
     console.log(JSON.stringify(await resp.json(), null, 2));
   language: javascript
-  timeout_ms: 30000  ← 30 seconds for network calls
+  timeoutMs: 30000  ← 30 seconds for network calls
 ```
 
-**Recommended timeouts:**
-| Operation | timeout_ms |
+**Recommended timeouts (`timeoutMs` is milliseconds):**
+| Operation | timeoutMs |
 |-----------|-----------|
 | File reading/parsing | 5000 - 10000 |
 | Local computation | 10000 |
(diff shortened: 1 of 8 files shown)


aktanazat:
Rebased onto current main (clean, no conflicts) and dropped the generated bundles from the diff, since the bundle workflow rebuilds and commits them on push to main. Locally on this head: tsc --noEmit clean, npm run bundle + assert-bundle OK, and tests/core/timeout-ms-1081.test.ts plus tests/adapters/zod3tov4.test.ts pass (25/25); the CI and OpenClaw E2E runs are still in action_required because fork PRs need maintainer approval. Closes #1081.

aktanazat:
Follow-up: retargeted to next per the PR template checklist (7 of 8 other open PRs target next) and rebased onto current origin/next; mergeable is now clean. The 414-line tests/core/timeout-ms-1081.test.ts is gone: the cases moved into tests/core/server.test.ts, the file CONTRIBUTING.md assigns to the Server & tools domain, trimmed to 162 lines covering the schema contract, the 120ms deadline, and rejection of the legacy timeout key; tests/core/server.test.ts and tests/adapters/zod3tov4.test.ts pass 521/521 and tsc --noEmit is clean at 07311ac.

Bortlesboat (commented):
Checked `07311aca52c80fc8dcbb6b190de0067811d18459` on native Windows with Node 24.14.0. Typecheck and the complete build/bundle assertions pass. The three new stdio regression tests and all 19 Zod adapter tests pass (22 total).

I also exercised the rebuilt server over stdio beyond the added cases: `ctx_execute_file` and serial/parallel `ctx_batch_execute` calls with `timeoutMs: 120` stop a four-second workload before it prints its completion marker. A serial batch control with `timeoutMs: 6000` completes and returns the marker, confirming that the short-deadline result isn't just a command that could never run. File/parallel responses explicitly report 120ms; measured request times were about 0.9–1.0s including process startup/cleanup.

One non-blocking observation: the serial batch response reports `(no output)` without a timeout annotation after capturing the command section. The serial formatting path already has this behavior on main at `33f7eb9`, so it isn't caused by this rename, but clients cannot infer that timeout from the successful response envelope alone. The parallel path does include `(timed out after 120ms)`.

The public schema and legacy-key rejection behave as intended in the checked MCP paths. I did not run the full repository suite or host UI integrations.

PR #1089, opened 2026-08-27
fix: derive context-mode registry key

### LINKED ISSUE #1078: [Bug]: cache-heal gates on "context-mode@context-mode" but the registry key is "context-mode@claude-context-mode" — the self-heal never runs
## Summary

The plugin-cache self-heal never executes. It gates on a registry key that does not exist, so its loop `continue`s on every entry and the heal body is unreachable.

Version installed: **1.0.169**. Runtime: Claude Code on macOS, Node v24.19.0.

## The mismatch

`installed_plugins.json` keys entries as `<plugin>@<marketplace>`. This plugin's marketplace is `claude-context-mode`, so its real key is:

```
context-mode@claude-context-mode
```

But the deployed heal script and `start.mjs` both compare against `context-mode@context-mode`.

Verified on a live install:

```console
$ node -e 'const p=JSON.parse(require("fs").readFileSync(process.env.HOME+"/.claude/plugins/installed_plugins.json","utf8")).plugins;
  console.log("context-mode@context-mode        ->", Object.keys(p).includes("context-mode@context-mode"));
  console.log("context-mode@claude-context-mode ->", Object.keys(p).includes("context-mode@claude-context-mode"));'
context-mode@context-mode        -> false
context-mode@claude-context-mode -> true
```

The full key list from that machine contains `'context-mode@claude-context-mode'` and no `context-mode@context-mode`.

## Where

In `1.0.169`, the wrong key appears at three sites in `start.mjs`:

```
:190   if (key !== "context-mode@context-mode") continue;
:206   if (key !== "context-mode@context-mode") continue;
:243   const pluginKey = "context-mode@context-mode";
```

and `start.mjs:378` writes that same template out to the user's hooks dir:

```js
writeFileSync(healHookPath, healScript, { mode: 0o755 });
```

so the deployed `~/.claude/hooks/context-mode-cache-heal.mjs` carries the same dead gate. The hook is registered in `settings.json` and runs — it just always falls through.

## Impact

1. The heal this hook exists to perform (repairing `installPath` after auto-update breaks `CLAUDE_PLUGIN_ROOT`, and normalizing stale version paths) has **never run** on affected installs.
2. Because `start.mjs:378` rewrites the file on every start, a user who corrects the deployed hook by hand loses the correction at the next start. Observed directly: a hand-applied fix was reverted within a day, at `2026-08-24 11:03:53`.

Point 2 is what makes this worth reporting rather than patching locally — there is no user-side fix that survives.

## Expected

The gate matches the actual `<plugin>@<marketplace>` key so the heal body runs.

## Suggested direction

Derive the key rather than hardcoding it — resolve the marketplace name at runtime, or match on the entry whose `installPath` lives under this plugin's own cache root, so a future marketplace rename cannot silently disable the heal again. If it stays a literal, all three sites plus the emitted template need to change together.

---
🤖 Investigated and drafted with Claude Code on the reporter's behalf. Happy to open a PR if you'd like the derived-key version.

7487:
## What / Why / How

Fixes #1078.

Claude Code keys installed plugins as `<plugin>@<marketplace>`, but cache healing assumed the marketplace was also named `context-mode`. Installs registered as `context-mode@claude-context-mode` therefore skipped every startup, postinstall, mid-session, and upgrade heal.

This change resolves the live registry key by the `context-mode@` plugin segment, uses it at each registry caller, and corrects stale `.mcp.json` cache traversal to map `<plugin>@<marketplace>` onto `<marketplace>/<plugin>/<version>`. The generated SessionStart hook uses the same marketplace-independent gate.

## Affected platforms

- [x] Claude Code
- [ ] Cursor
- [ ] JetBrains Copilot
- [ ] Gemini CLI
- [ ] Qwen Code
- [ ] OpenCode
- [ ] KiloCode
- [ ] Codex CLI
- [ ] OpenClaw (Pi Agent)
- [ ] Pi
- [ ] Kiro
- [ ] Antigravity
- [ ] Zed
- [ ] All platforms

## Test plan

- Red: new custom-marketplace cases failed in startup gating, cache-path mapping, key resolution, and global postinstall recovery.
- Green: focused suite passes (4 files, 97 tests).
- Full suite passes (213 files, 4,780 passed, 25 skipped).
- `npm run typecheck` passes.
- `npm run build` passes, including bundle and asymmetric-drift assertions.
- `npm run lint --if-present` passes (no lint script is defined).

## Checklist

- [x] `npm test` passes
- [x] Docs updated if needed (no user-facing docs change needed)
- [x] No Windows path regressions (uses existing `node:path` joins/resolves)

### DIFF
--- a/scripts/heal-installed-plugins.mjs
+++ b/scripts/heal-installed-plugins.mjs
@@ -20,6 +20,14 @@
 import { existsSync, readFileSync, writeFileSync, readdirSync, unlinkSync, statSync } from "node:fs";
 import { resolve, sep } from "node:path";
 
+export function resolveContextModePluginKey(plugins) {
+  if (!plugins || typeof plugins !== "object" || Array.isArray(plugins)) {
+    return "context-mode@context-mode";
+  }
+  return Object.keys(plugins).find((key) => key.startsWith("context-mode@"))
+    ?? "context-mode@context-mode";
+}
+
 /**
  * @typedef {Object} HealResult
  * @property {string[]} healed - one of: "entry-version", "enabled-plugins"
@@ -525,10 +533,10 @@ export function healClaudeJsonMcpArgs({ dotClaudeJsonPath, pluginCacheParent, ne
 
 /**
  * Remove every `.mcp.json` from per-version directories under
- * `<pluginCacheRoot>/<owner>/<plugin>/<X.Y.Z>/`.
+ * `<pluginCacheRoot>/<marketplace>/<plugin>/<X.Y.Z>/`.
  *
  * @param {{ pluginCacheRoot: string, pluginKey: string }} opts
- *   pluginKey is the "<owner>@<plugin>" form (e.g. "context-mode@context-mode").
+ *   pluginKey is the "<plugin>@<marketplace>" registry form.
  * @returns {SweepResult}
  */
 export function sweepStaleMcpJson({ pluginCacheRoot, pluginKey }) {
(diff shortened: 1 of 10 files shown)


7487:
Hi! Just a gentle follow-up on this one — the branch is conflict-free against main and the fix is small and self-contained. Happy to rebase or adjust anything if needed. Thanks!

PR #1111, opened 2026-09-01
fix(stats): stop heartbeat lifetime scans

koenvg:
## What / Why / How

`persistStats()` currently refreshes lifetime totals every 30 seconds by calling `getLifetimeStats()`. Each MCP bridge therefore opens and queries every SessionDB independently. With thousands of project databases and multiple concurrent bridges, that periodic heartbeat amplifies CPU usage, directory I/O, and SQLite sidecar churn.

This change moves the expensive aggregation onto the explicitly requested `ctx_stats` path. The result is published atomically to one shared, validated lifetime-token snapshot per adapter stats directory. Periodic bridge heartbeats only read that O(1) snapshot and never discover or open SessionDBs.

The snapshot uses unique temporary files before rename, so concurrent `ctx_stats` publishers cannot collide. Missing or corrupt snapshots retain the existing best-effort behavior.

## Affected platforms

- [x] All platforms

## Test plan

- Added snapshot round-trip coverage to `tests/session/lifetime-stats.test.ts`.
- Added a regression/performance test with 2,443 DB files and 12 concurrent reader processes, asserting bounded completion and no WAL/SHM creation.
- Added server wiring assertions proving `persistStats()` does not call `getLifetimeStats()` and `ctx_stats` publishes the shared snapshot.
- `bun run typecheck`
- `bunx vitest run tests/session/lifetime-stats.test.ts tests/core/server.test.ts` — 507 passed.
- `bun run test` — 4,776 passed, 27 skipped.

## Checklist

- [x] `npm test` passes
- [x] Docs updated if needed (no user-facing documentation change required)
- [x] No Windows path regressions (uses `node:path` joins)

<details>

The shared snapshot path is built with `node:path.join()`. Concurrent writes use per-process UUID temporary names and atomic rename; no hardcoded separators or `/tmp` paths were added.

</details>


jrobber:
This is biting me often, this would be great to get merged in!

PR #1128, opened 2026-09-08
Fix curl/wget pipe and multiline routing bounds

### LINKED ISSUE #1125: curl/wget rule ignores pipe targets: `curl … | head -c 200` is redirected, and the rest of a multi-line block is dropped with it
### Summary

The Bash `curl`/`wget` rule in `hooks/core/routing.mjs` classifies a command as a context flood from redirection/`-o` flags **on the same chain segment only**. It never inspects pipe targets, so `curl -s URL | head -c 200` — output already bounded — is redirected. And because a redirect rewrites the **entire** command with an `echo`, every unrelated command in a multi-line block is silently dropped with it.

Version: 1.0.169. Platform: Claude Code 2.1.263, Linux.

### Reproduction

```js
import { routePreToolUse } from "./hooks/core/routing.mjs";
const t = c => console.log(
  routePreToolUse("Bash", { command: c }, { mcpToolsAvailable: true, projectDir: process.cwd() })
    ?.action === "modify" ? "BLOCK" : "pass", "|", c);

t(`curl -s http://x/ | head -c 200`);   // BLOCK  <- output is bounded
t(`curl -s http://x/ | jq -r .name`);   // BLOCK  <- output is one line
t(`curl -s http://x/ > /tmp/out.json`); // pass
t(`curl -s -o /tmp/out.json http://x/`);// pass
```

### Two distinct defects

**1. Pipe-blindness.** `hasFileOutput` only looks for `-o`/`--output`/`>`/`>>`. A pipeline is a deliberate reduction by the agent — `| head`, `| jq -r`, `| grep -c`, `| wc -l`, `| python3 -c` all bound the output — but the pipe stage is never examined, so all of them are treated as unbounded floods.

**2. Newline is not a segment separator.** The splitter is `stripped.split(/\s*(?:&&|\|\||;)\s*/)`. A multi-line script is therefore one segment, so a single bare `curl` anywhere in it marks the whole block dangerous. Since the redirect replaces the whole command, unrelated `ssh` / `docker ps` / `grep` calls in that block never run and the agent gets no indication they were dropped.

### Impact measured

Replayed 13 days of real transcripts through the unmodified hook: **25 redirects across 20 sessions in 9 projects**, and roughly 22 of them had provably bounded output. Health checks (`curl -s -o /dev/null -w '%{http_code}'` inside a multi-line block), API probes piped to `jq -r`, and header inspections piped to `grep` were all hit.

### Suggested fix

Two small changes to that block:

```js
// 1. newline separates commands; join backslash continuations first
const segments = stripped.replace(/\\\n/g, " ").split(/\s*(?:&&|\|\||;|\n)\s*/);

// 2. a pipeline is a deliberate reduction — only pass-through sinks still flood
const pipeStages = s.split("|").slice(1).map(x => x.trim()).filter(Boolean);
const boundedByPipe = pipeStages.length > 0 &&
  !pipeStages.every(st => /^(cat|tee|less|more)\b/.test(st));

if (/\s(-v|--verbose|--trace|-D\s+-)\b/.test(s)) return true; // stderr, no pipe bounds it
if (boundedByPipe) return false;
if (!hasFileOutput) return true;
```

Result on the same corpus: **25 → 5 blocked**, and all 5 remaining are genuine bare `curl -s URL`. `tests/hooks/core-routing.test.ts` still passes 95/95 unchanged.

Happy to open a PR if the approach looks right to you.

Tiancheng-Xu:
## Summary
- split curl/wget commands on newlines and escaped newlines as well as existing chain operators
- allow bounded pipeline sinks such as head, jq, grep, and wc while preserving unbounded/pass-through handling
- add regression coverage for curl pipelines, multiline commands, and backslash continuation

Fixes #1125

## Validation
- `npm run test -- tests/hooks/core-routing.test.ts`
- `npm run typecheck`
- `npm run build`
- `git diff --check`
- targeted secret regex scan (clean)

Prettier is not configured in this repository, and `gitleaks` is unavailable in the local environment.

### DIFF
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -731,8 +731,10 @@ export function routePreToolUse(toolName, toolInput, projectDir, platform, sessi
     // curl/wget — allow silent file-output downloads, block stdout floods (#166).
     // Algorithm: split chained commands, evaluate each segment independently.
     if (/(^|\s|&&|\||\;)(curl|wget)\s/i.test(stripped)) {
-      // Split on chain operators (&&, ||, ;) to evaluate each segment
-      const segments = stripped.split(/\s*(?:&&|\|\||;)\s*/);
+      // Split on chain operators (&&, ||, ;, newline) and normalize escaped newlines first.
+      const segments = stripped
+        .replace(/\\\r?\n/g, " ")
+        .split(/\s*(?:&&|\|\||;|\r?\n)\s*/);
       const hasDangerousSegment = segments.some(seg => {
         const s = seg.trim();
         // Only evaluate segments that contain curl or wget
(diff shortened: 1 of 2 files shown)


Tiancheng-Xu:
Addressed both review points in commit `7f3a978`: pass-through sink detection now handles `|&` pipelines plus `command`/`builtin` and absolute-path forms such as `/bin/cat`; added regression tests for each equivalent form. Targeted `vitest run tests/hooks/core-routing.test.ts` passes (106 tests).

PR #1133, opened 2026-09-08
fix: only route curl and wget in command position

### LINKED ISSUE #1106: PreToolUse curl/wget regex still matches unquoted argument words, not the command (residual of #63)
## Bug Description

The Bash PreToolUse hook's curl/wget detection matches the literal word "curl" or "wget" anywhere in the command as a whitespace-delimited token, not only when it is the command being run. #63 fixed the quoted-argument case with `stripQuotedContent()`, but that function only strips quoted and heredoc text. An unquoted mention of the word still matches.

## Reproduction

```bash
which curl
type curl
which -a curl
```

None of these run curl. All three get silently rewritten to the canned "context-mode: curl/wget redirected" echo message instead of executing.

## Root Cause

In `hooks/core/routing.mjs`:

```js
if (/(^|\s|&&|\||\;)(curl|wget)\s/i.test(stripped)) {
```

This tests for the token anywhere in the command, not specifically in command position. `stripQuotedContent()` only removes quoted/heredoc content before this test runs, so an unquoted bare word like "curl" as an argument to `which`, `type`, `command -v`, etc. still matches.

## Expected Behavior

Same as #63's original ask: only match when curl/wget is the actual command token of a segment (the first word after a chain operator), not whenever the word appears anywhere in the command text.

## Environment

context-mode v1.0.169, Claude Code on macOS.

## Impact

Same as #63: the substitution looks like success (exit 0) but the original command never runs. Checking whether curl is shadowed, or mentioning it in a comment or echo, silently produces a fabricated tool-redirect message instead of the real output.

Rithb898:
## What / Why / How

Fixes #1106. The curl/wget PreToolUse guard now matches only executable command positions (segment or pipeline starts), so commands such as `which curl` are not rewritten. The existing per-segment safety checks reuse the same matcher, preventing later arguments from turning a safe chain into a redirect.

## Affected platforms

- [x] All hook-capable platforms

## Test plan

- [x] `npx vitest run tests/hooks/core-routing.test.ts` (97 passed)
- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npm test` (4758 passed, 43 skipped)
- [x] `git diff --check`

### DIFF
--- a/hooks/core/routing.mjs
+++ b/hooks/core/routing.mjs
@@ -240,6 +240,10 @@ function stripQuotedContent(cmd) {
     .replace(/"[^"]*"/g, '""');                   // double-quoted strings
 }
 
+// Match curl/wget only when they are the command at the start of a shell
+// segment or pipeline stage, not an unquoted argument to another command.
+const CURL_WGET_COMMAND = /(?:^|(?:&&|\|\||;|\|&?|[\r\n])\s*)(curl|wget)(?=\s|$)/i;
+
 /**
  * Built-in allowlist of structurally-bounded Bash commands (#463).
  *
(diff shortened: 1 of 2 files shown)


Rithb898:
Closing as a duplicate of the already-open PR #1128, which covers the same command-position routing behavior and broader pipeline/multiline cases.

PR #1148, opened 2026-09-11
fix: bundle bin/statusline.mjs's analytics import (marketplace installs never get build/)

tmurphy24-lab:
## Summary

- `bin/statusline.mjs`'s KPI numbers never render on a marketplace/git-clone install — it dynamically imports `build/session/analytics.js`, but `build/` is gitignored and only produced by `tsc`, which only runs on `npm install` (`prepublishOnly`). Every other `session/*.ts` consumer (`session-db`, `session-extract`, `session-snapshot`) has a co-located CI bundle with `build/` as documented fallback (see `hooks/session-loaders.mjs`, `hooks/session-helpers.mjs`); `analytics.ts` was the one left off that list.
- Effect: the statusline permanently shows the `saves ~98% of context window` marketing headline instead of real numbers, for every Claude Code plugin-marketplace install (and any other adapter that installs by cloning rather than `npm install`) — even though the underlying session tracking (`ctx_stats`) works correctly the whole time. `loadAnalytics()`'s `try/catch` swallows the failure silently, so nothing surfaces it.
- Fix: bundle `src/session/analytics.ts` the same way as the other session modules (`bin/analytics.bundle.mjs`, `git add -f`'d by `bundle.yml` same as the rest), and give `bin/statusline.mjs` the same bundle-first/`build/`-fallback resolution `hooks/session-helpers.mjs`'s `loadSessionDbModule()` already uses.

## Repro (before/after, per CONTRIBUTING.md)

Simulated a git-clone-only install (no `build/`, no bundle):
```
before: context-mode  ●  saves ~98% of context window
after:  context-mode  ●  84.9 KB kept out  ·  84.9 KB/day  ·  preserved across compact, restart & upgrade
```

## Test plan

- [x] Added tests to `tests/core/cli.test.ts` (existing file, per "do NOT create new test files") — new `describe("bin/analytics.bundle.mjs — statusline marketplace install support")` block mirroring the existing `cli.bundle.mjs` block's conventions: package.json bundle/assert-bundle coverage, bundle artifact exists + readable, `bin/statusline.mjs` bundle-first/fallback source contract.
- [x] `npm run build` (tsc + bundle + assert-bundle + assert-asymmetric-drift) — clean
- [x] `npx vitest run tests/core/cli.test.ts` — 224 passed, 1 skipped (pre-existing skip)
- [x] `npm test` (full suite) — 4754 passed, 6 failed. The 6 failures (`tests/executor.test.ts` background-detach PID capture, `tests/core/server.test.ts` RPC-init timeout + doctor-renderer) reproduce identically on unmodified `next` with this change stashed — pre-existing environment flakiness (Windows process-detach/PID timing, MCP init timeout), not touched by this diff.
- [x] `.gitignore` and `.github/workflows/bundle.yml` updated to cover the new `bin/analytics.bundle.mjs` the same way the existing bundles are (gitignored, `git add -f`'d by CI).

Did not touch `cli.bundle.mjs` / `server.bundle.mjs` even though `npm run build` regenerates them — that diff is pure minifier-output churn (different esbuild run, no source change to `server.ts`/`cli.ts`), so I left them out of the commit to keep the diff scoped to the actual fix.


tmurphy24-lab:
@zen fix merge @claude same

zen:
What?

Issue #868, opened 2026-06-22
[Bug]: Very intrusive message do not allow to type


carolitascl:
### Platform

Pi

### context-mode version

1.0.165

### Debug script output (REQUIRED)

```json
none
```

### Exact prompt that triggered the bug (REQUIRED)

```text
Appears doing nothing
```

### Full error output (REQUIRED)

```text
No error
```

### Steps to reproduce (REQUIRED)

Suddendly this message appears in the editor box and do not allow to type

[mcp-bridge] [context-mode] idle MCP bridge child self-shutdown after 180000ms with no activity (#854)

### What have you tried to fix it?

No

### Pre-submission checklist

- [x] I have run the debug script and pasted the output above
- [x] I am using the latest version of context-mode
- [x] I have searched existing issues for duplicates
- [x] I have included steps to reproduce the issue

### Operating System

macOS (Apple Silicon)

frsswq:
Same thing happened on Pi coding agent (Windows 11, WSL 2)

mksglu:
Fixed in **v1.0.166**.

Two coupled fixes:

1. **The line no longer reaches your editor.** context-mode runs in-process under Pi, and its MCP bridge was writing diagnostics to `process.stderr` — which is Pi's raw-mode TUI terminal, so the text rendered straight into the prompt box. All Pi-adapter diagnostics now route to `pi.logger` (Pi's rotating log under `~/.omp/logs/`), the documented channel for in-process extensions — never the terminal. (Your keystrokes were actually still being read underneath; it was the *display* that got corrupted — but either way it's gone now.)

2. **The helper no longer shuts down on you.** The underlying event — `idle MCP bridge child self-shutdown` — was the idle reaper releasing the **active** session's helper after 3 minutes of inactivity. That reaper exists to stop *abandoned sub-agent* helpers from piling up, but it shouldn't touch the foreground session. It now keeps the interactive session's helper alive; only sub-agent / non-interactive helpers are idle-reaped. (Your helper is still released immediately when Pi itself exits.)

@carolitascl, @frsswq — please update and confirm the editor stays clean and your tools don't drop after a pause:

```
/context-mode:ctx-upgrade        # or: npm update -g context-mode
```

Thank you for the precise report — the exact line and where it appeared made the root cause unambiguous.

PR #1166, opened 2026-09-16
fix(pi): spawn a Windows-spawnable runtime and keep bridge diagnostics

zhenda-hub:
## What / Why / How

Fixes the Pi bridge registering **zero** `ctx_*` tools on Windows, and makes that
failure diagnosable.

**Root cause.** `resolveJsRuntimeForBridge()` returned the bare command name from
`detectRuntimes()` (`"bun"`). On Windows `child_process.spawn()` without
`shell: true` goes through CreateProcess, which executes real binaries only —
npm-style installs (nvm4w, scoop, `npm -g`) ship `bun` / `bun.cmd` / `bun.ps1`
and **no `bun.exe`**, so the spawn failed with `ENOENT`. That lands on the child's
`error` event, which the bridge reported as `"MCP server exited"` — a spawn that
never produced a process looked exactly like a crash.

**Why it was invisible.** `makeBridgeDiag()` writes to `pi.logger`, but Pi's
extension API has no `logger` (checked on 0.85.1). The line was dropped, and by
design it never goes to stderr/TUI (#868). The only symptom was
`ctx_* tools will not be callable`, with no cause anywhere on disk.

**Changes** — all confined to `src/adapters/pi/mcp-bridge.ts`:

1. `resolveJsRuntimeForBridge()` — on Windows a candidate is accepted only when it
   names an executable file (`.exe` / `.com`), and bare names are resolved through
   PATH first. POSIX behaviour is unchanged.
2. The child `error` / `exit` handlers no longer share a bare `onExit()`; the cause
   is propagated (`spawnExitError` carries the exit code / signal).
3. `makeBridgeDiag()` falls back to `<pi config dir>/context-mode/bridge-diag.log`
   when no logger exists. The terminal is still never touched.

### Evidence from the affected machine

```
$ node -e "require('child_process').spawn('bun',['-e',''])"
  spawn error: ENOENT - spawn bun ENOENT

$ node -e "require('child_process').spawn('node',['-e',''])"
  ok

$ where bun
C:\nvm4w\nodejs\bun        <- extensionless shim (no .exe)
C:\nvm4w\nodejs\bun.cmd
C:\nvm4w\nodejs\bun.ps1
```

Before: `WARNING: failed to bridge MCP tools to Pi (MCP server exited).`
After: `bootstrapMCPTools()` registers all 11 tools and the bridge child stays up.

## Affected platforms

- [x] Pi

## Test plan

6 new cases in `tests/adapters/pi-mcp-bridge.test.ts`:

| Case | Asserts |
|:--|:--|
| Windows: `bun` shim skipped | falls back to `node.exe` |
| Windows: real `bun.exe` present | still preferred (no speed regression) |
| Windows: only shims on PATH | returns `null` instead of an unspawnable name |
| POSIX | the `.exe` rule does not apply |
| `spawnExitError` | carries code / signal / `unknown` |
| `makeBridgeDiag` fallback | writes `<config>/context-mode/bridge-diag.log`, still never stderr |

Results:

- `tests/adapters/pi-mcp-bridge.test.ts` — **36/36 pass** (was 30)
- `npm run typecheck` — clean
- Full suite — 4701 passed. The 2 remaining failures are **pre-existing
  Windows-only** (`cache-heal-self-heal.test.ts` symlink creation,
  `executor.test.ts` detached-process kill); re-running both on a stashed,
  unmodified tree reproduces them identically.

## Checklist

- [ ] `npm test` passes — 2 pre-existing Windows-only failures, unrelated (see above)
- [ ] Docs updated if needed — no documented surface changed
- [x] Targets `next` branch


zhenda-hub:
CI is sitting at `action_required` (first-time contributor), so no workflow has
actually run yet — flagging in case that is not visible on your end.

While it waits, everything is verified locally:

- `npm run typecheck` — clean
- `npm run build` + all bundle asserts — pass
- `tests/adapters/pi-mcp-bridge.test.ts` — **36/36** (30 existing + 6 new)
- full suite — 4701 passed; the only 2 failures are pre-existing Windows-only ones
  (`cache-heal-self-heal.test.ts` symlink creation, `executor.test.ts` detached-process kill).
  I confirmed both by re-running them on a stashed, unmodified tree — identical failures,
  so they are unrelated to this change.

Happy to adjust scope or split anything out if you would prefer it smaller.

PR #1169, opened 2026-09-16
fix(opencode): support OpenCode 2 plugin API (V1/V2 dual export)

Scratchydisk:
## Summary

OpenCode 2's plugin loader ignores the V1 `{ id, server }` shape this package's `src/adapters/opencode/plugin.ts` exports — per OpenCode's own migration guide (opencode.ai/v2/docs/migrate-v1): *"V1 plugin implementations do not run in V2."* It only recognizes a default export produced by `Plugin.define({ id, setup })` from the new `@opencode/plugin` package.

### Repro against the currently-published package (verified against a real `opencode@2.0.3` install)

```jsonc
// opencode.jsonc
{ "plugins": ["context-mode"] }
```

```
level=WARN message="failed to load plugin" target=context-mode
cause="Cause([Fail(PluginModule.LoadError: Plugin must export a default
definition with an id and an effect or setup function. (cause:
SchemaError(Missing key at ["default"]["effect"] Missing key at
["default"]["setup"])))])"
```

`ctx_*` tools never register, no hooks fire, and `opencode plugin list` silently shows nothing for it — worse than a crash, since there's no indication anything is wrong until a user notices the tools are gone.

## Fix

Adds a V2 `setup()` entrypoint alongside the existing V1 `server()`, following OpenCode's own documented "support V1 and V2 from one package" pattern (`opencode.ai/v2/docs/build/plugins#support-v1`): spread `Plugin.define({ id, setup })` into the same default-export object that already carries `server`. OpenCode 1.x and KiloCode keep calling `server()` — completely unchanged — while OpenCode 2 reads `id`/`setup()` off the same object and ignores `server()`.

The V2 `setup()` re-implements all five hooks using OpenCode 2's documented mapping (`opencode.ai/v2/docs/build/plugins/migrate-v1`):

| V1 hook | V2 API |
|---|---|
| `tool.execute.before` | `ctx.tool.hook("execute.before", ...)` |
| `tool.execute.after` | `ctx.tool.hook("execute.after", ...)` |
| `event` (message.updated) | `ctx.event.subscribe()` |
| `chat.message` | `ctx.session.hook("prompt", ...)` |
| `experimental.session.compacting` | `ctx.session.hook("compaction", ...)` |
| `experimental.chat.system.transform` | `ctx.session.hook("context", ...)` |
| `tool` map | `ctx.tool.transform(editor => editor.add(...))` |

The custom `ctx_*` tools now declare JSON Schema (required by V2) instead of a Zod shape, reusing the existing `zod3ShapeToV4` conversion (added for #574) plus Zod v4's own `toJSONSchema()`.

`@opencode/plugin` is imported dynamically (`await import(...)` inside a try/catch), not as a static top-level import, so a stale/partial install missing that dependency degrades to V1-only instead of breaking the whole module for existing OpenCode 1.x/KiloCode users.

### Verification

- Empirically confirmed against a real `opencode@2.0.3` binary (not just docs): the unmodified published package fails with the exact error above; this branch's build loads with no error and reaches model dispatch.
- All 120 pre-existing `opencode-plugin`/`adapters/opencode`/`parse-opencode-usage` tests pass unchanged — zero behavior change to the V1/KiloCode path.
- 2 new tests lock in the dual-export shape (`id`+`server` always present; `setup` present when `@opencode/plugin` resolves).
- `npm run typecheck` and `npm run build` both clean.

### Known gap for maintainer review

The exact field names on OpenCode 2's `ToolExecuteBefore`/`ToolExecuteCompleted`/`ToolExecuteFailed` event types aren't spelled out in the public docs beyond short examples (`event.tool`, `event.input`, `event.status`, `event.result`, `event.error`). I inferred `event.sessionID` by analogy with every other hook interface OpenCode documents (`SessionRequestHook`, `SessionRetryHook`, `PermissionEvaluation`, etc. all expose it), and defensively no-op when it's absent — but I don't have a way to exercise a live tool call under OpenCode 2 in this environment (no provider credentials), so the `tool.execute.after` → session-event-capture path should be exercised against a real agent turn before release. Everything else (plugin discovery/loading, config parsing, the dual-export mechanics, and the hook-name mapping itself) was verified directly against the real binary.

## Test plan

- [x] `npm run typecheck`
- [x] `npm run build`
- [x] `npx vitest run tests/opencode-plugin.test.ts tests/adapters/opencode.test.ts tests/session/parse-opencode-usage.test.ts` (122 passed)
- [ ] Maintainer: exercise `tool.execute.after` against a live OpenCode 2 session with real provider credentials to confirm `event.sessionID` and the result-text extraction shape


Scratchydisk:
Closing for now — pulling this back to do more thorough local testing against a live OpenCode 2 session before resubmitting. Will reopen (or open a fresh PR) once that's done.

